PR #2, opened 2026-07-07
Fix PCM metadata: add missing author.contact field

gcolonese:
## Bug

Installing the prebuilt `konnect-pcm-v0.1.1.zip` release via KiCAD 10's Plugin and Content Manager fails with:

```
Unable to parse package metadata:
At /author, value: {"name": "mixelpixx"}
required property 'contact' not found in object
```

## Root cause

`packaging/metadata.json` (the source manifest copied into the PCM zip by `packaging/build-pcm.ps1`) declares `author` with only a `name` field. KiCAD's PCM package schema requires `author` to be an object with both `name` and `contact` — see the [KiCAD packages3d schema](https://gitlab.com/kicad/services/kicad-packages3d/-/blob/master/repository/schemas/pcm.v1.schema.json), `ContactInformation` definition (`required: ["name", "contact"]`).

## Fix

Add the missing `contact` field, pointing at the author's GitHub profile:

```diff
   "author": {
-    "name": "mixelpixx"
+    "name": "mixelpixx",
+    "contact": {
+      "web": "https://github.com/mixelpixx"
+    }
   },
```

Minimal, JSON-only change — no Rust code touched.

## Checklist

- [x] `cargo test --workspace --lib --tests` passes (51 tests, 0 failed)
- [x] `cargo clippy --workspace -- -D warnings` clean
- [x] `cargo fmt --all` applied (no changes — no Rust code touched)
- [x] N/A: no tool added/removed, so `tool_count`/`tool-directory.md` untouched

PR #3, opened 2026-07-07
Fix plugin Start Server on KiCAD 10 / Windows / OneDrive

gcolonese:
Note: this branch was originally cut off `fix-pcm-metadata-contact` (#2) since it wasn't merged yet. #2 has since merged, so this diff is now clean -- only `plugin/__init__.py`.

## Summary

After #2's fixes let us actually install the plugin via PCM and click Start Server on KiCAD 10 / Windows, we hit six more bugs in `plugin/__init__.py`, all blocking Start Server:

1. **`pcbnew.ShowInfoBarMsg` / `ShowInfoBarError` removed in KiCAD 10** -- every call raised `AttributeError` on button-click and killed the plugin before the server could start. The fatal "binary not found" path now uses `wx.MessageBox`; the redundant started/stopped/already-running info messages were dropped since the dialog already shows Running/Stopped state on next open.

2. **GUI-parent subprocess spawns attach a new console on Windows** -- pcbnew has no console of its own, so Windows attaches a new one to the child process. That console makes stdin look like a TTY, which trips konnect's `is_terminal()` double-click-install-wizard path instead of MCP server mode (visible as console-window flashes when clicking Start). `Popen` now passes `creationflags=CREATE_NO_WINDOW` on Windows.

3. **`stderr=None` gave the GUI child a broken handle** -- pcbnew has no valid stderr to inherit, so `tracing_subscriber` init in konnect crashed silently. `Popen` now uses `stderr=subprocess.DEVNULL`.

4. **Enterprise Windows with OneDrive-redirected Documents** -- PCM installs the plugin under `Documents/`, which OneDrive syncs. Both `bin/konnect.exe` and `settings.json` land tagged `IO_REPARSE_TAG_CLOUD`, so Windows returns `ERROR_ACCESS_DENIED` on `CreateProcess` for the exe and on serde's `File::open` for the config. A new `_stage()` helper copies both files to `%LOCALAPPDATA%\konnect\cache\` (never OneDrive-synced) before launch; no mtime gating needed since the copy is ~50ms once per Start.

5. **PID file for server tracking** -- the module-level `_server_process` global reset on every KiCAD `Run()` re-import, so `is_server_running()` always reported `False` after the first click even with konnect still alive. Orphans from a task-killed KiCAD also held port 3000 and made the next Start silently fail. A `_PID_FILE` at `%LOCALAPPDATA%\konnect\cache\server.pid` is written after `Popen` and removed in `finally`; `is_server_running()` proxies via file existence. `_kill_tracked()` uses `taskkill` on Windows, not `os.kill(pid, 0)` -- on Windows any signal value invokes `TerminateProcess`, so a "probe" would actually kill the process. `start_server()` preflights `_kill_tracked()` to clean orphans before spawning.

6. **Retry loop respawned konnect after Stop** -- `_run_server` had a `for attempt in range(3)` loop that retried on any `wait()` return. Stop Server killed konnect, `wait()` returned normally (no exception), `finally` ran, the loop iterated, and konnect respawned. The user couldn't actually stop the server. The retry loop is deleted; if konnect crashes on startup the user clicks Start again.

Also collapsed two staging helpers (`_resolve_binary` + `_stage_settings`) into a single `_stage()` used inline in `_run_server` -- same behavior, half the code.

## Test plan

Verified locally on KiCAD 10 / Windows:

- [x] Start Server -- konnect spawns from cache, binds port 3000, PID file written; dialog reopen shows Running.
- [x] Stop Server -- taskkill runs, port freed, PID file removed; dialog reopen shows Stopped.
- [x] Task-kill KiCAD with server running -- konnect survives as a tracked orphan, PID file persists.
- [x] Reopen KiCAD after crash -- dialog reads PID file, shows Running correctly.
- [x] Stop after crash-recovery -- cleanly kills the orphan, all state clean.
- [x] `cargo test --workspace --lib --tests` -- passes (no-op, Python-only change)
- [x] `cargo clippy --workspace -- -D warnings` -- clean (no-op)
- [x] `cargo fmt --all` -- no diff (Python file)

No tool changes, so `tool_count` / `tool-directory.md` are not applicable.


mixelpixx:
G, i thought this merged, i approved others that touched this, so my bad. Needs your love again

gcolonese:
> G, i thought this merged, i approved others that touched this, so my bad. Needs your love again

no worries, all set now!

PR #14, opened 2026-07-09
feat(pcb_board): add import_svg_logo tool

boovaragan:
## Summary
- Add `import_svg_logo` to the `pcb_board` toolset — closing the roadmap's `import_svg_logo` item.
- Parses an SVG file and places its filled paths as silkscreen/copper artwork on a board. Curved paths are flattened into straight-line polygon outlines since KiCAD's board format (`PolyLine`/`gr_poly`) only supports points and circular arcs, not Bezier curves.
- Tries the KiCAD IPC API first, falls back to a direct file edit if KiCAD isn't running — the same dual-path pattern already used by `add_board_text`.

## Motivation
`import_svg_logo` was one of two concrete, unblocked items left on the roadmap's Tools list (the other, hierarchical sheets, is a much larger prerequisite for other features). KiCAD's IPC API already exposes a `PolySet`/`PolyLine` polygon type ready to use; what was missing was SVG parsing and curve-to-polygon conversion.

## Implementation
- **New `crates/konnect-core/src/tools/svg_import.rs`**: uses `usvg` (added as a new dependency) to parse SVG content, then flattens each path's curves into polygon points with hand-written cubic/quadratic Bezier sampling (fixed 16-segment subdivision — sufficient fidelity for silkscreen/copper artwork). The flattening math (`SvgSegment`, `flatten_subpaths`, `cubic_bezier_points`, `quad_to_cubic`) is fully decoupled from `usvg`'s types via a small local enum, so it's unit-testable without any SVG parsing involved.
- **New `board_polygon()`** IPC builder in `crates/konnect-ipc/src/builders.rs`, following the existing `board_segment`/`board_rectangle`/`board_circle` pattern.
- **New `format_gr_poly()`** S-expression formatter in `pcb_board.rs` for the file-fallback path — a standalone filled `gr_poly`, not tied to a net/zone like the existing `format_zone_polygon`.
- Updated `tool_count` (`pcb_board` 10→11, workspace total 175→176) across `registry.rs`, `tool-directory.md`, `DEV.md`. Fixed a pre-existing staleness in `DEV.md`'s file-tree comment (`pcb_export.rs` was still annotated "9 tools" from before #11 merged; corrected to 13 while in the area). Moved the roadmap item to Done.

## Test plan
- [x] `cargo test --workspace --lib --tests` — 118/118 passed
- [x] `cargo clippy --workspace -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] 12 unit tests on the geometry module: Bezier sampling, quadratic-to-cubic elevation, subpath flattening (closed shapes, multiple subpaths, degenerate-input dropping, curve expansion), scale/placement math, and real `usvg` parsing of small SVG strings (both straight-line and curved paths).
- [x] 3 unit tests on the new `board_polygon` IPC builder.
- [x] 4 handler-level tests exercising the file-fallback path end-to-end against a real temporary board + SVG file (happy path, empty-SVG rejection, missing required argument).

### Manual / real-world verification
- Built a release binary and ran the tool against a real production board (not a synthetic fixture) with a multi-shape test SVG (mixing straight lines and curves) — confirmed visually in KiCAD's PCB editor that the artwork renders correctly on F.Silkscreen.
- **Round-tripped the generated board through real `kicad-cli`** (`pcb export svg`) to confirm the emitted `gr_poly` syntax is genuinely valid KiCAD format, not just visually plausible — succeeded with no errors, and the exported SVG contains real rendered path data.
- Re-tested against a real, geometrically complex production logo (a stag's head — lots of fine curve detail) on a real board: 17 flattened polygons, clean visual result in KiCAD.
- **Live IPC path**: attempted to verify the IPC-first branch independently by pointing a standalone test harness at a live, already-open KiCad instance. This surfaced a genuine constraint documented in KiCAD's own developer docs: a process not launched directly *by* KiCAD has no reliable way to discover the running instance's IPC socket path (only `KICAD_API_SOCKET`, set for processes KiCAD itself launches). This is a pre-existing characteristic of every IPC-touching tool in this file, not something introduced here — `import_svg_logo`'s IPC branch reuses the exact same `with_ipc`/`create_items` machinery `add_board_text`, `add_zone`, etc. already rely on. Since a live connection couldn't be independently exercised from the outside, verification for that branch rests on it being the same well-established code path as the rest of the file, plus the file-fallback path being thoroughly proven above (which is what runs whenever IPC isn't reachable, matching this whole verification session).

## Known limitation (candidate follow-up)
Overlapping/nested SVG shapes are not converted into holes — each subpath becomes its own separate filled polygon rather than detecting winding-rule holes via `PolygonWithHoles.holes`. A logo with a true cutout (e.g. a letter "O") will render as two overlapping filled shapes rather than one shape with a hole punched through it. Happy to follow up on this if there's interest.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -30,6 +36,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "arrayref"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
+
+[[package]]
 name = "async-trait"
 version = "0.1.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -45,6 +63,12 @@ name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
+name = "autocfg"
+version = "1.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
 [[package]]
 name = "axum"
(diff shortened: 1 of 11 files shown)


mixelpixx:
Merged — verified live: a rect+circle test SVG imported as 2 filled polygons on F.SilkS via the file-fallback path, and kicad-cli DRC loads the modified board. The layered, unit-testable flattening pipeline is exactly the right shape. Thanks for closing out this roadmap item.

PR #16, opened 2026-07-10
feat: sheet pin lifecycle for sch_hierarchy (PR-B of 2)

boovaragan:
## Summary
- **PR-B of 2** for ROADMAP.md's "Hierarchical sheets" item — base branch is `feat/hierarchical-sheets` (#15), not `main`, since this depends directly on the `Sheet`/`SheetPin` types added there. This diff should only show the new sheet-pin work; the sheet-object diff from #15 is reviewed separately.
- Adds 5 tools to the `sch_hierarchy` toolset (7→12): `import_sheet_pins`, `add_sheet_pin`, `edit_sheet_pin`, `delete_sheet_pin`, `validate_sheet_pins`.
- With this, "Hierarchical sheets" moves to `ROADMAP.md`'s Done section — both halves are complete.

## Motivation
PR-A gave sheets a physical box on the parent canvas; without pins, that box has no electrical connection to the rest of the design — it's cosmetic. This PR adds the connection points, completing the feature.

## Implementation

**`import_sheet_pins` is the primary path**, not `add_sheet_pin`. It scans the child sheet's `hierarchical_label`s and auto-generates matching pins on the parent, skipping names that already have a pin — mirroring KiCAD's own "Import Sheet Pins" command rather than requiring every pin to be paired to a label by hand (the error-prone workflow KiCAD's own UI avoids). New pins are placed along one edge of the sheet box (`right` by default, `left` optional), stacked below any existing pins using the sheet's own current pin count as the starting offset — so repeated calls as labels get added incrementally don't overlap previously-placed pins.

`add_sheet_pin`/`edit_sheet_pin`/`delete_sheet_pin` round out manual CRUD for the cases `import_sheet_pins` doesn't cover — placing a pin before its label exists, renaming, retyping, repositioning, or removing one. `pin_type` is validated against KiCAD's real electrical types (`input`, `output`, `bidirectional`, `tri_state`, `passive`) at the handler level — a typo gets a structured error listing the valid values instead of silently writing invalid data into the schematic file.

**`validate_sheet_pins`** is read-only: it walks the whole sheet tree and reports hierarchical labels with no matching parent pin, and pins with no matching child label — the "check first" companion to `import_sheet_pins`, useful as a pre-ERC sanity check or to catch drift after manual edits. Reuses the same cycle-safe DFS traversal pattern already established by `get_sheet_hierarchy`/`renumber_sheet_pages` in PR-A.

One drive-by cleanup: the first draft of `validate_sheet_pins`/its internal helper carried a `project_name` parameter for symmetry with the other tools, but pin/label name matching isn't project-scoped (only page numbers are) — `clippy`'s `only_used_in_recursion` lint caught it as dead weight passed through every recursive call with no actual use, so it was removed from the schema and the helper rather than suppressed.

## Test plan
- [x] `cargo test --workspace --lib --tests` — 128/128 passed
- [x] `cargo clippy --workspace -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] 10 new handler tests: `import_sheet_pins` creates matching pins from real hierarchical labels; a second `import_sheet_pins` call skips already-imported names instead of duplicating; `add_sheet_pin` rejects a duplicate pin name and an invalid `pin_type`; `edit_sheet_pin` renames + retypes, and errors when no fields are given; `delete_sheet_pin` removes a pin and errors on a missing one; `validate_sheet_pins` reports a deliberately mismatched label/pin pair by name, then reports zero issues once `import_sheet_pins` has synced them.

### Domain validation
Not yet performed — pending review, same as PR-A. Planned: draw real hierarchical labels in a KiCAD 10 sub-sheet, run `import_sheet_pins`, confirm the result matches KiCAD's own "Import Sheet Pins" output for the same labels; deliberately mismatch a pin and confirm KiCAD's ERC flags it the same way `validate_sheet_pins` does.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -51,7 +51,7 @@ Konnect/
 │   │           ├── sch_analysis.rs   # 15 tools (union-find net graph, connectivity)
 │   │           ├── sch_batch.rs      # 10 tools (single-read/single-write atomic operations)
 │   │           ├── sch_export.rs     # 7 tools (SVG/PDF/netlist/ERC)
-│   │           ├── sch_hierarchy.rs  # 7 tools (typed Sheet model, add/edit/move/delete/duplicate + hierarchy/page queries)
+│   │           ├── sch_hierarchy.rs  # 12 tools (typed Sheet model, sheet CRUD + hierarchy/page queries + pin lifecycle)
 │   │           ├── pcb_board.rs      # 10 tools (S-expr file editing, IPC fallback)
 │   │           ├── pcb_components.rs # 13 tools (IPC real-time via NNG+protobuf)
 │   │           ├── pcb_routing.rs    # 12 tools (traces, vias, nets, netclasses)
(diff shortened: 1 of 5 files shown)


mixelpixx:
Merged — reviewed and verified live against real KiCAD: built a parent/child hierarchy through the new tools, imported a hierarchical label as a sheet pin, validate_sheet_pins reported clean, and kicad-cli loads the resulting hierarchy. That closes out the hierarchical-sheets roadmap item — excellent pair of PRs. Note: counts in tool-directory.md were resolved to the registry-derived totals during the merge (your branch predated this week's count corrections).

PR #22, opened 2026-07-13
feat(library): produce complete, DRC-clean footprints and symbols

boovaragan:
## What this does

Makes `create_footprint` and `create_symbol` produce **complete, DRC-clean, placeable
parts** instead of stubs. Today both emit only the electrical geometry — a footprint gets
pads and nothing else (**no courtyard, so it fails KiCAD DRC**; no silkscreen, fab outline,
reference/value text, pin-1 marker, or 3D model), and a symbol gets pins with no body
rectangle. So "create a custom part" produced something you had to hand-finish in KiCAD
before it was usable.

This advances the **"Symbol & footprint creation"** roadmap item to production quality.
(Happy to move it to `## Done` if you'd like — I left `ROADMAP.md` untouched since the tools
pre-existed and this completes rather than introduces them.)

## Footprints now include

Derived from the pad bounding box:

- **Courtyard** (`F.CrtYd`) — the DRC-required keep-out. Clearance follows package type:
  **0.5 mm** through-hole, **1.0 mm** BGA, **0.15 mm** for parts smaller than 0603,
  **0.25 mm** other SMT. Overridable via `courtyard_clearance` or `package_type`.
- **Silkscreen** (kept off the pads) and **fab** outlines. The fab uses the component body
  when `body_width`/`body_height` are supplied, else the pad envelope.
- **Reference/value** text on the standard layers.
- **Pin-1 marker** — a silk **dot beside pin 1's pad** (not at the footprint corner, where
  it's ambiguous with the last pin) plus a **chamfered fab corner**.
- An optional **3D model** — new `model` arg emits `(model …)` with offset/scale/rotate; the
  path is passed through verbatim (absolute or `${KICAD*_3DMODEL_DIR}/…`).

## Symbols now include

- A **body rectangle** whose edges pass through the pin roots, so each pin's far end touches
  the border and its connection point sits **outside**, with **spacing beyond the outermost
  pins** on edges that have none.
- Pin names/numbers **shown by default** (new `show_pin_names`/`show_pin_numbers` flags).
- The correct **`konnect`** generator string (was a stale `kicad-mcp`).

## Design / compatibility

All the geometry (pad bbox, courtyard-clearance selection, pin-1 corner, chamfered outline,
pin root, symbol body) is pure, unit-tested functions; only the final S-expression assembly
touches strings. Every new argument is **optional and back-compatible** — existing callers
keep working and simply get complete output.

## Test plan

- `cargo test --workspace --lib --tests` → **214/214** (205 + 9 new)
- `cargo clippy --workspace -- -D warnings` → clean
- `cargo fmt --all -- --check` → clean

The 9 new tests cover the clearance rule across all cases, pin-1 placement, chamfer geometry,
pin-root direction, the symbol body with end-spacing, and 3D-model emission — plus two
round-trip tests that generate a real footprint and symbol and assert the courtyard, pin-1,
model, and body nodes are present and that the output re-parses via `konnect-sexp`.

Domain-validated by generating a real QFN-16 footprint (mapped to a KiCAD QFN 3D model) and a
multi-pin IC symbol and opening them in KiCAD's Footprint/Symbol editors: courtyard passes
the footprint checker, the pin-1 dot sits beside pin 1, the 3D model renders in the 3D viewer,
and the symbol body encloses the pins with the connection points outside the border.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -40,6 +40,21 @@ pub fn tools() -> Vec<ToolDef> {
                             },
                             "required": ["number", "type", "shape", "x", "y", "width", "height"]
                         }
+                    },
+                    "body_width": { "type": "number", "description": "Physical component body width in mm (optional; used for silk/fab outlines). Falls back to the pad envelope if omitted." },
+                    "body_height": { "type": "number", "description": "Physical component body height in mm (optional)." },
+                    "package_type": { "type": "string", "description": "'smd' (0.25mm courtyard), 'through_hole' (0.5mm), 'small' (0.15mm, <0603), or 'bga' (1.0mm). Sets courtyard clearance when courtyard_clearance is not given." },
+                    "courtyard_clearance": { "type": "number", "description": "Explicit courtyard clearance in mm (overrides package_type / auto-detection)." },
+                    "model": {
+                        "type": "object",
+                        "description": "Optional 3D model to associate with the footprint.",
+                        "properties": {
+                            "path": { "type": "string", "description": "Path to the 3D model file (.step/.wrl); absolute or a KiCAD env-var path like ${KICAD9_3DMODEL_DIR}/..." },
+                            "offset": { "type": "object", "description": "{x,y,z} in mm (default 0,0,0)" },
+                            "scale": { "type": "object", "description": "{x,y,z} (default 1,1,1)" },
+                            "rotate": { "type": "object", "description": "{x,y,z} in degrees (default 0,0,0)" }
+                        },
+                        "required": ["path"]
                     }
                 },
                 "required": ["output", "name", "pads"]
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Reviewed and validated locally — approving.

**Code review:** geometry helpers are pure and well-tested; all new args are optional and verified back-compatible (existing calls now simply emit the full graphics); courtyard-clearance rule matches the description; pin-1 dot placement beside the pad (not the corner) is a good call; `generator` fix confirmed harmless (nothing parses the old `kicad-mcp` string).

**Local verification (Windows, KiCAD 10.0.4):**
- `cargo test --workspace --lib --tests`: 214/214, clippy/fmt clean.
- Drove the branch-built server over MCP stdio: generated a QFN-16 footprint (with 3D model ref) and a 5-pin symbol via the real tool calls.
- `kicad-cli fp/sym export svg` both parse and plot cleanly, including the `(pin_numbers)` empty-visibility token.
- **DRC A/B** with `missing_courtyard = error`: the new footprint passes with 0 violations; an old-style pads-only stub fails with `[missing_courtyard]`. Exactly the fix as advertised.

One nit, not blocking: the `create_footprint` response reports `"courtyard": true` even when `pads` is empty and no graphics are emitted — will tidy that up on main.

PR #24, opened 2026-07-13
Add Docker packaging for the Konnect MCP server

gcolonese:
## What

A multi-stage `Dockerfile` so the server runs with no local Rust toolchain -- easy to hand to a colleague or host centrally, dependencies included.

- **Builder**: `rust:1-bookworm` + protoc + cmake + libprotobuf-dev (for `konnect-ipc`'s prost-build and the statically-built nng).
- **Runtime**: `debian:bookworm-slim` carrying just the compiled binary, non-root user.
- **stdio mode** (default entrypoint): `docker run --rm -i -v <proj>:/work konnect` -- plugs into an MCP client the same way markitdown does.
- **HTTP mode**: `docker compose up -d` uses `docker/konnect.toml` and publishes to host loopback (`127.0.0.1:3000`), with a `/health` healthcheck.
- Pre-creates the first-run install marker so the host-install step (meaningless in a container) is skipped.

## Security-conscious default

The MCP server is unauthenticated and its tools edit files / run `kicad-cli`, so compose publishes the port to loopback only; the container binds `0.0.0.0` internally (required for published ports) but the host exposure is `127.0.0.1`. Widening it is documented as requiring an auth proxy on a trusted network.

## Tests (per CONTRIBUTING.md)

- **In the cargo gate**: a `config.rs` unit test parses the *actual shipped* `docker/konnect.toml` and asserts it yields HTTP bound to all interfaces -- guards against config-schema drift. No Docker needed, so Docker does not become a dev dependency.
- **Out of the cargo gate**: `docker/smoke-test.sh` builds the image and runs a real MCP `initialize` handshake over both stdio and HTTP. Kept as a script (not a `#[test]`) on purpose; transport logic stays covered by the existing `tests/protocol_{http,stdio}.rs`.

## Verified

- Image builds clean; `smoke-test.sh` green on both transports.
- `cargo fmt --all --check` clean; `cargo clippy -p konnect --lib --tests -- -D warnings` clean.

Happy to open a tracking issue first if you'd prefer to discuss the approach per CONTRIBUTING.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+# Keep the build context small and reproducible.
+/target/
+crates/*/target/
+crates/schematic-viewer/
+.git/
+.github/
+.claude/
+CLAUDE.md
+*.exe
+*.dll
+*.so
+*.dylib
+dist/
+.playwright/
+.playwright-cli/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,50 @@
+# Konnect MCP server -- containerized so it can be run without a local Rust
+# toolchain (e.g. `docker run --rm -i konnect`, the same way markitdown ships).
+#
+# Two stages: a builder with the Rust toolchain + protoc + cmake (nng builds
+# libnng via cmake, konnect-ipc runs prost-build via protoc), and a slim
+# Debian runtime that carries only the compiled binary.
+
+# ---- Builder ----------------------------------------------------------------
+FROM rust:1-bookworm AS builder
+
+# libprotobuf-dev ships the well-known protos (google/protobuf/*.proto) under
+# /usr/include; konnect-ipc's build.rs derives that include dir from $PROTOC.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends protobuf-compiler libprotobuf-dev cmake \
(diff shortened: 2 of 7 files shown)


mixelpixx (approved):
Reviewed — approving.

**What I checked locally (no Docker on this machine, so the image build itself is verified by the green CI only):**
- Dockerfile/compose/config/smoke-test read-through: multi-stage build is correct for this workspace (protoc + libprotobuf-dev for prost-build's well-known types, cmake for nng), runtime is non-root on bookworm-slim, and the pre-created install marker is the right call for a containerized server.
- Security posture is sound and well-documented: compose publishes to `127.0.0.1` only, and I verified the claim in `docker/konnect.toml`'s comment — `transport/http.rs` really does 403 non-localhost `Origin` headers, so the bind-all inside the container is defensible.
- The only Rust change is additive (a test): `shipped_docker_config_serves_http_on_all_interfaces` passes and usefully pins the shipped config against schema drift.
- README section is accurate, including the caveat that IPC/kicad-cli tools need a local KiCAD rather than the container.

Nice touch keeping the smoke test out of the cargo gate so Docker doesn't become a dev dependency.

PR #25, opened 2026-07-15
feat(library): multi-unit symbol support in create_symbol

boovaragan:
## What this does

Adds **multi-unit symbol** support to `create_symbol`. Today it can only author single-unit
symbols, so multi-unit parts — dual/quad op-amps, hex-inverter gate banks, relays, multi-bank
connectors — can't be created (you'd hand-build the extra units in the Symbol Editor).

Builds directly on the completed-symbol/footprint work.

## New arguments

- **`units`** — an array, one entry per unit (each with its own `pins`). Each becomes a unit
  (A, B, C…) — sub-symbols `NAME_1_1 … NAME_N_1` — with its own rectangular body sized to that
  unit's pins (reusing the edge-aware `symbol_body_rect` geometry, so each unit's pins touch its
  border with the connection points outside).
- **`power_pins`** — shared power pins (V+/V-, VCC/GND). They become a **dedicated final "power
  unit"** (`NAME_<N+1>_1`), following KiCAD's own convention — e.g. **74LS00** has the four gates
  as units 1–4 and VCC/GND as **unit 5**. This is deliberate: putting the power pins in one unit
  means they appear on exactly one placed unit, instead of being drawn on *every* unit (which
  would make each duplicate need wiring to pass ERC).

KiCAD infers the unit count from the highest unit index, so no explicit count is emitted.

## Compatibility

**Fully back-compatible.** With no `units`, the single-unit `pins` path is byte-for-byte
unchanged — verified by a regression test (a single-unit symbol still emits only `NAME_0_1`, no
`NAME_1_1`). The per-unit builder is factored out and shared by both paths.

## Rectangular bodies for now

Each unit gets a rectangular body (consistent with the single-unit tool). Conventional glyph
shapes (op-amp triangles, real logic-gate bodies) are a planned follow-up — they need a shape
library and are a much larger change; the parts here are functionally correct (right pins on the
right units, power on its own unit).

## Test plan

- `cargo test --workspace --lib --tests` → **227/227** (2 new multi-unit tests)
- `cargo clippy --workspace -- -D warnings` → clean
- `cargo fmt --all -- --check` → clean

Two new tests: a single-unit regression, and a dual op-amp asserting two signal units + a
dedicated power unit, that V+/V- appear exactly once, one body rectangle per unit, no `_0_1`, and
that the output re-parses via `konnect-sexp`.

Domain-validated by generating a dual op-amp (2 gates + power unit) and a quad NAND (4 gates +
power unit, same unit layout as 74LS00) and opening them in KiCAD's Symbol Editor: the unit
selector shows the signal units plus a separate power unit, and the power pins appear only on that
power unit.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -118,7 +118,10 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "create_symbol",
-            "Create a new KiCAD schematic symbol and append it to a .kicad_sym library file.",
+            "Create a new KiCAD schematic symbol and append it to a .kicad_sym library file. \
+             Supports single-unit symbols (via `pins`) and multi-unit parts like dual/quad \
+             op-amps or gate banks (via `units` + optional `common_pins`). Each unit gets a \
+             rectangular body sized to its pins.",
             json!({
                 "type": "object",
                 "properties": {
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Approving. Validated live on Windows/KiCAD 10.0.4 with the branch merged onto current main: generated a dual op-amp (2 signal units + shared V+/V− as `power_pins`) and a single-unit part through the real MCP tool. The library carries exactly `DUAL_OPAMP_1_1/_2_1/_3_1` with the power pins appearing once (in the dedicated final unit, matching the 74xx convention you cite), the single-unit symbol still emits only `_0_1`, and `kicad-cli sym export svg` renders per-unit SVGs for all of it — KiCAD's own parser agrees with the structure.

Factoring the shared `build_symbol_unit` out of the single-unit path (rather than duplicating the body/pin assembly) was the right move, and the schema docs on `units` vs `pins` vs `power_pins` are clear enough for an LLM caller to get right first try. Full local gate green (266 tests), CI green.

Rectangular bodies for op-amps/gates as a follow-up sounds right — functional beats pretty for now.

Issue #26, opened 2026-07-15
macOS: schematic tools silently skip symbol embedding (lib_symbols left empty) — symbol dirs and kicad-cli not found in the app bundle


tascsystems-andrew:
## Environment

- macOS 26.5, Apple Silicon (M-series)
- KiCad 10.0.3, standard DMG install at `/Applications/KiCad/KiCad.app`
- Konnect built from source at d70b561 (main, 2026-07-14)

## What happens

On macOS, `find_symbol_dirs()` (`konnect-schematic-editor/src/library.rs`) and `find_kicad_symbol_dirs()` (`konnect-core/src/tools/mod.rs`) take the non-Windows branch, which only probes the Linux paths `/usr/share/kicad/symbols` and `/usr/local/share/kicad/symbols`. KiCad on macOS ships its libraries inside the app bundle at `/Applications/KiCad/KiCad.app/Contents/SharedSupport/symbols`, so no symbol directory is ever found.

The nasty part is that it fails silently: `ensure_lib_symbol()` returns early when resolution fails, so `add_schematic_component` reports success —

```json
{"added": "Device:R", "reference": "R1", ...}
```

— but the written schematic has an empty `lib_symbols` section. KiCad then opens it with missing symbol references, and `kicad-cli sch erc` can't make sense of it. From an MCP client (Claude Code/Desktop) this looks like Konnect working fine while quietly producing broken schematics.

Repro:

```bash
printf '%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"t","version":"0"}}}' \
  '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
  '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"load_toolset","arguments":{"name":"sch_components"}}}' \
  '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"create_schematic","arguments":{"path":"t.kicad_sch"}}}' \
  '{"jsonrpc":"2.0","id":4,"method":"tools/call","params":{"name":"add_schematic_component","arguments":{"schematic":"t.kicad_sch","lib_id":"Device:R","x":100,"y":80}}}' \
  | ./konnect
grep -c '(symbol "Device:R"' t.kicad_sch   # 0 — no embedded definition
```

[...]

PR #27, opened 2026-07-15
docs: add macOS install section to README

tascsystems-andrew:
## What

Adds a macOS subsection under Installation, and updates the Requirements line to match reality now that release tarballs for both Apple archs exist.

Covers:
- Release tarballs (`aarch64` / `x86_64-apple-darwin`) + the `xattr -d com.apple.quarantine` step browser downloads need while binaries are unsigned
- Build from source, including the **cmake** prerequisite — `nng-sys` compiles the NNG C library with cmake, which never surfaces in CI because GitHub's runners preinstall it, but a fresh Mac fails with `is 'cmake' not installed?` (added to the Windows/Linux lines too, same reason)
- `~/Library/Application Support/konnect/config.toml` example with the app-bundle `kicad-cli`/`kicad` paths and the KiCad 10 IPC socket (`ipc:///tmp/kicad/api.sock`)
- Claude Desktop config path on macOS + Claude Code `.mcp.json` pointer
- Honest gaps: PCM zip currently bundles only `konnect.exe`; the schematic viewer builds and launches on macOS (Tauri 2 → WKWebView, no WebView2 needed) but has less mileage than on Windows

## Verified on

macOS 26.5 (Apple Silicon), KiCad 10.0.3, Rust 1.97, repo at d70b561:
- `cargo build --release -p konnect` ✓ (after `brew install protobuf cmake`)
- `cargo test --workspace --lib --tests` → 225 passed, 0 failed
- MCP stdio handshake, `load_toolset` → `tools/list_changed` → tools callable ✓
- Schematic pipeline (create project, add/wire components, ERC via kicad-cli, SVG export) ✓
- Live IPC against a running KiCad 10 (`open_project` → `kicad_ui_running: true`) ✓
- `cd crates/schematic-viewer && cargo build --release` ✓

Docs-only change; no code touched. Related: #26 tracks the macOS symbol-path bug (fix PR incoming) — this README text is written to stay correct whether or not that lands, since `KICAD10_SYMBOL_DIR`/config workarounds aren't mentioned here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,10 +110,61 @@ Verify: open the **PCB Editor** → **Tools → External Plugins** → you shoul
 ### Build from source
 
 ```bash
-# protoc is required (protobuf code generation)
-# Windows: choco install protoc / macOS: brew install protobuf / Linux: apt install protobuf-compiler
+# protoc is required (protobuf code generation), and cmake (the nng crate
+# compiles the NNG C library with it).
+# Windows: choco install protoc cmake
+# macOS:   brew install protobuf cmake
+# Linux:   apt install protobuf-compiler cmake
 cargo build --release -p konnect
 ```
+
+### macOS
+
+The [Releases](https://github.com/mixelpixx/Konnect/releases) page ships
+standalone server binaries for both Apple Silicon (`aarch64-apple-darwin`) and
+Intel (`x86_64-apple-darwin`). They are not yet code-signed, so if you download
+one through a browser, clear the quarantine flag before first launch:
+
+```bash
+tar xzf konnect-v*-aarch64-apple-darwin.tar.gz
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Approving. Fact-checked the load-bearing claims against the repo rather than taking the prose on faith: the v0.1.3 release does ship both darwin tarballs (aarch64 + x86_64), `~/Library/Application Support/konnect/config.toml` is exactly what `dirs_config_path()` reads on macOS, the config keys match the schema (including the `kicad_binary`/`kicad_cli` split), and the PCM-is-Windows-only caveat is accurate today (#45/#46 will obsolete that sentence — whoever merges #46 should touch it up). The cmake addition to the build prereqs is a good catch, nng does require it.

PR #28, opened 2026-07-15
fix(macos): find KiCad symbol libraries and kicad-cli inside the app bundle

### LINKED ISSUE #26: macOS: schematic tools silently skip symbol embedding (lib_symbols left empty) — symbol dirs and kicad-cli not found in the app bundle

tascsystems-andrew:
Fixes #26.

## What

On macOS both symbol-library resolvers took the generic non-Windows branch and probed only `/usr/share/kicad/symbols` + `/usr/local/share/kicad/symbols`, so a stock KiCad install (libraries inside `KiCad.app`) was never found and `add_schematic_component` / `add_power_symbol` silently wrote schematics with an empty `lib_symbols` section (details + repro in #26).

Three symmetrical changes, no behavior change on Windows or Linux:

- `konnect-schematic-editor/src/library.rs::find_symbol_dirs()` — add a `#[cfg(target_os = "macos")]` branch probing `/Applications/KiCad/KiCad.app/Contents/SharedSupport/symbols`, `/usr/local/share/kicad/symbols`, and the per-user `~/Applications/KiCad/KiCad.app/...` variant. This is the resolver the schematic tools actually use.
- `konnect-core/src/tools/mod.rs::find_kicad_symbol_dirs()` — same fix, keeping the two copies consistent.
- `konnect/src/install.rs::detect_kicad()` — per-platform candidate lists so `konnect init`/`konnect status` recognize a standard macOS install instead of printing "Not found in standard locations" (Linux gets `/usr/bin` + `/usr/local/bin` candidates; previously there were no non-Windows candidates at all). The Windows list and registry probing are unchanged.

`KICAD10_SYMBOL_DIR` still takes priority in both resolvers, and the app-bundle path follows the precedent `conformance_test.rs::demo_dirs()` already set for the demos corpus.

## Verified

macOS 26.5 (Apple Silicon), KiCad 10.0.3:

- Before: `add_schematic_component` with `Device:R`, no env vars → reports success, `grep -c '(symbol "Device:R"'` on the output = 0
- After: same call → embedded definition present (= 1), `kicad-cli sch erc` clean on the produced schematic, power symbols embed too
- `konnect status` now prints `[+] Found: /Applications/KiCad/KiCad.app/Contents/MacOS/kicad-cli`

Checklist: `cargo test --workspace --lib --tests` (225 passed, 0 failed) / `cargo clippy --workspace -- -D warnings` clean / `cargo fmt --all` applied. No tools added or removed, so `tool_count` / `tool-directory.md` are untouched.

### DIFF
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -544,7 +544,29 @@ fn find_kicad_symbol_dirs() -> Vec<std::path::PathBuf> {
             }
         }
     }
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(target_os = "macos")]
+    {
+        // KiCad on macOS ships its libraries inside the app bundle.
+        let mut candidates = vec![
+            std::path::PathBuf::from(
+                "/Applications/KiCad/KiCad.app/Contents/SharedSupport/symbols",
+            ),
+            std::path::PathBuf::from("/usr/local/share/kicad/symbols"),
+        ];
+        if let Ok(home) = std::env::var("HOME") {
+            // Per-user install (KiCad.app dragged into ~/Applications)
+            candidates.push(
+                std::path::PathBuf::from(home)
+                    .join("Applications/KiCad/KiCad.app/Contents/SharedSupport/symbols"),
+            );
+        }
+        for p in candidates {
+            if p.is_dir() && !dirs.contains(&p) {
+                dirs.push(p);
+            }
+        }
+    }
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
     {
         let candidates = ["/usr/share/kicad/symbols", "/usr/local/share/kicad/symbols"];
         for c in &candidates {
--- a/crates/konnect-schematic-editor/src/library.rs
+++ b/crates/konnect-schematic-editor/src/library.rs
@@ -217,7 +217,28 @@ pub fn find_symbol_dirs() -> Vec<PathBuf> {
         }
     }
 
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(target_os = "macos")]
+    {
+        // KiCad on macOS ships its libraries inside the app bundle.
+        let mut candidates = vec![
+            PathBuf::from("/Applications/KiCad/KiCad.app/Contents/SharedSupport/symbols"),
+            PathBuf::from("/usr/local/share/kicad/symbols"),
+        ];
+        if let Ok(home) = std::env::var("HOME") {
+            // Per-user install (KiCad.app dragged into ~/Applications)
+            candidates.push(
+                PathBuf::from(home)
+                    .join("Applications/KiCad/KiCad.app/Contents/SharedSupport/symbols"),
+            );
+        }
+        for p in candidates {
+            if p.is_dir() && !dirs.contains(&p) {
+                dirs.push(p);
+            }
+        }
+    }
+
+    #[cfg(not(any(target_os = "windows", target_os = "macos")))]
     {
         let candidates = ["/usr/share/kicad/symbols", "/usr/local/share/kicad/symbols"];
         for c in &candidates {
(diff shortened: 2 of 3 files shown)


mixelpixx (approved):
Approving. I can't exercise the macOS paths from this Windows machine, so the mac-side claims rest on your documented verification (which is thorough — the before/after `grep -c` on the embed plus clean ERC is exactly the right evidence) and green macOS CI. What I did verify locally:

- Windows regression check: the standard-path list and registry probe are byte-identical, and a live `add_schematic_component` on the branch build still embeds `Device:C` into `lib_symbols` on Windows.
- The cfg restructure is right: macOS gets its own branch, Linux keeps the old generic one and gains sensible `detect_kicad` candidates it previously lacked entirely.
- Full workspace gate green on the branch merged with current main (264 tests), CI green on all three platforms.

Nice symmetry keeping both resolver copies and `detect_kicad` consistent.

Issue #30, opened 2026-07-15
pin_endpoint() returns the pin body end, not the electrical tip — connect_pins wires miss every pin by one pin-length


tascsystems-andrew:
**Severity: critical** — every consumer of pin coordinates inherits a 1-pin-length offset (1.27 mm for Device:R).

## Repro (d70b561, KiCad 10.0.3)

Place `Device:R` at (100.33, 80.01), rotation 0, and call `get_schematic_pin_locations`: Konnect reports pins at y = **77.47 / 82.55**. KiCad's own ERC report for the same file places them at y = **76.20 / 83.82** (the `(at)`-derived tips, ±3.81 from center). `connect_pins` between two stacked resistors draws its wire between the 82.55/… points — one pin-length short at each end — so the true pin tips sit mid-span on the wire, which KiCad never joins (a pin connects only at its endpoint, verified: ERC still reports all pins unconnected with the wire present).

## Cause

`crates/konnect-sexp/src/schematic.rs:293-299`:

```rust
// The wire-connection point is at pin_origin + length in pin direction
let tip_x = pin.local_x + pin.length * angle_rad.cos();
```

The premise is inverted. In the KiCad symbol format the pin's `(at x y angle)` **is** the connection tip; the angle points from the tip toward the body, and the drawn pin line extends `length` mm inward. Adding length lands exactly on the body outline (checked against Device:R's body rectangle), i.e. the wrong end.

Consumers inheriting the offset: `connect_pins` / `resolve_pin_endpoint` (sch_wiring.rs), `get_schematic_pin_locations` + batch variant, `connect_to_net` snapping (sch_export.rs:338), and the design-review audits (design_review.rs, 6 call sites).

## Fix

PR incoming: drop the length term (`transform_pin(pin.local_x, pin.local_y, t)`), plus regression tests pinning the eeschema-verified coordinates. There was no existing test coverage of `pin_endpoint`, which is how this survived.

Composes with #29: with both fixed, `connect_pins` between two resistors produces `Net-(R1-Pad2)` containing both pins; with either alone, the netlist stays broken.

PR #36, opened 2026-07-15
fix(sch): write root UUID and eeschema-style instance paths so KiCAD forms wire-only nets

### LINKED ISSUE #29: Konnect-created schematics are netlist-dead: no root UUID + unresolvable instance paths suppress every wire-only net
**Severity: critical** — schematics build, export, and look correct, but KiCad forms no wire-only nets from them at all.

## Repro (Konnect d70b561, KiCad 10.0.3, kicad-cli ground truth)

Tool calls: `create_project` → `add_schematic_component` Device:R ×2 (R1, R2) → a wire joining R1.2–R2.1 (placed on the true pin tips; see note below). All calls succeed. Then:

```
kicad-cli sch export netlist out.net <file>.kicad_sch
```

→ components R1/R2 are listed, but the `(nets)` section is **completely empty** — no joined net, not even `unconnected-*` stubs. `kicad-cli sch erc --severity-all` reports all 4 pins as `pin_not_connected` plus dangling-wire noise. Placing a `(label "MID")` on the same wire makes net `/MID` appear with both pins — which is why this survives casual testing: label- and power-named nets work, so schematics that name everything look fine while every unnamed net is silently dropped.

## Cause

Two cooperating defects (bisected on file copies — each alone still yields an empty netlist):

1. The blank-schematic templates write no root `(uuid ...)`: `project.rs:387` (`blank_kicad_sch`), `sch_components.rs:286` (`create_schematic`), `sch_hierarchy.rs:259` (`create_blank_schematic`). (`schematic_builder.rs:53-59` does emit one — the templates are the inconsistent path.)
2. `add_schematic_component` hardcodes `(instances (project "" (path "/" ...)))` (`sch_components.rs:392-409`); `add_power_symbol` does the same (`sch_wiring.rs:941-956`); and `link_sheet` patches child symbols with `(path "/<sheet-uuid>/")` — trailing slash, missing root prefix (`sch_hierarchy.rs:315`).

KiCad ≥7 resolves symbol instances against the hierarchy path `"/<root-uuid>[/<sheet-symbol-uuid>…]"`. When no instance entry resolves, KiCad 10's netlister silently drops every subgraph that would get an auto-generated name. Verified matrix: uuid-only → empty; path-only → empty; **uuid + `path "/<root-uuid>"` → nets form**. The `project` name string is ignored by kicad-cli (wrong/empty names still netlist); `sheet_instances` is not required.

## Fix

PR incoming: root uuid in all three templates (shared helper), instance entries written via the existing `Symbol::set_instance_path` as `(project "<file-stem>" (path "/<root-uuid>" ...))` matching eeschema, eeschema-format sheet paths (`/<root>/<sheet-uuid>`, no trailing slash) in `link_sheet`/page entries, and in-place repair of older Konnect files (root uuid assigned on next edit).

Note: `connect_pins` output *still* won't netlist after this fix alone, because of an independent defect — `pin_endpoint()` returns the pin body end instead of the electrical tip (filed separately) — the two compose into "electrically dead schematics".

Found while building a real tube-amp PA sheet with Konnect on macOS; all analysis against unpatched upstream main. Related: #26.

tascsystems-andrew:
Fixes #29.

## What

Konnect-written schematics had no root `(uuid ...)` and hardcoded `(project "" (path "/" ...))` instance entries, so KiCAD's netlister silently dropped every wire-only net (full analysis + bisection matrix in #29). This PR makes every write path emit what eeschema emits:

- **Templates**: `create_project`, `create_schematic`, and `create_blank_schematic` now share one `blank_schematic_template()` helper (tools/mod.rs) that emits a root uuid — the same thing `schematic_builder.rs` already did on its own path.
- **Symbol + power-symbol instances**: written via the existing `Symbol::set_instance_path` as `(project "<file-stem>" (path "/<root-uuid>" (reference ...) (unit 1)))`. Files from earlier Konnect versions (no root uuid) are repaired in place on their next edit (`ensure_root_uuid`).
- **Hierarchy**: `link_sheet` writes child-symbol paths as `/<root-uuid>/<sheet-uuid>` (eeschema format — previously `/<sheet-uuid>/` with a trailing slash and no root). Sheet page entries and `renumber_sheet_pages` use the same hierarchical prefixes instead of `"/"`, threaded through the recursive walk.
- **project_name defaults**: now the schematic's file stem instead of `""` — kicad-cli ignores the key (verified), but eeschema keys instance data by it and writes the stem for standalone root sheets.

## Verification

- Ground truth, before → after on the identical tool-call sequence (two `Device:R` + a wire, no labels): `kicad-cli sch export netlist` goes from a **completely empty `(nets)` block** to `Net-(R1-Pad2)` containing R1.2 + R2.1, with correct `unconnected-*` stubs for the open pins. (Wire placed at the true pin tips; `connect_pins` output additionally needs the independent #30 fix.)
- 3 new regression tests: template carries uuid; instance path is `/<root-uuid>` under the file-stem project; legacy files without a root uuid get repaired on edit. Hierarchy tests updated to assert the eeschema path format they previously asserted the broken format for.
- `cargo test --workspace --lib --tests` 229 passed / 0 failed; clippy `-D warnings` clean; fmt applied.

## Notes for review

- Behavior change: files edited after this PR carry instance entries under the file-stem project name; pre-existing entries under `""` are left in place (eeschema prunes stale entries on save, and kicad-cli matches by path, not name). `next_free_page` scans under the new default name — a hierarchy previously built entirely with `""` entries would restart page numbering; those files were netlist-dead anyway.
- Root `sheet_instances` (`(path "/" (page "1"))`) is *not* added — verified not required by the netlister; eeschema adds it on first save. Can add for byte-parity if you prefer.

### DIFF
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -293,6 +293,41 @@ pub fn get_path(args: &Value, key: &str) -> anyhow::Result<std::path::PathBuf> {
     Ok(std::path::PathBuf::from(s))
 }
 
+/// Project name used in symbol/sheet `(instances (project "..." ...))` entries:
+/// the schematic's file stem, matching what eeschema writes when it saves a
+/// standalone root sheet.
+pub fn project_name_for(sch_path: &std::path::Path) -> String {
+    sch_path
+        .file_stem()
+        .and_then(|s| s.to_str())
+        .unwrap_or("")
+        .to_string()
+}
+
+/// Minimal valid blank schematic, with a freshly generated root `(uuid ...)`.
+/// The root UUID is mandatory: KiCAD's netlister resolves symbol instance
+/// paths against it and silently forms no wire-only nets when it's missing.
+pub fn blank_schematic_template() -> String {
+    format!(
+        "(kicad_sch\n\t(version 20250610)\n\t(generator \"konnect\")\n\t(generator_version \"10.0\")\n\t(uuid \"{}\")\n\t(paper \"A4\")\n\t(lib_symbols\n\t)\n)\n",
+        konnect_sexp::writer::new_uuid()
+    )
+}
+
+/// Root UUID of a loaded schematic, assigning a fresh one when the file
+/// predates Konnect writing root UUIDs — the file is repaired on its next
+/// overwrite. Instance paths are built as "/<root-uuid>[/<sheet-uuid>…]".
+pub fn ensure_root_uuid(sch: &mut konnect_schematic_editor::Schematic) -> String {
+    match &sch.uuid {
+        Some(u) => u.clone(),
+        None => {
+            let u = konnect_sexp::writer::new_uuid();
+            sch.uuid = Some(u.clone());
+            u
+        }
+    }
+}
+
 #[cfg(test)]
 mod arg_helper_tests {
     use super::*;
--- a/crates/konnect-core/src/tools/project.rs
+++ b/crates/konnect-core/src/tools/project.rs
@@ -384,8 +384,8 @@ fn blank_kicad_pro(name: &str) -> String {
     )
 }
 
-fn blank_kicad_sch() -> &'static str {
-    "(kicad_sch\n\t(version 20250610)\n\t(generator \"konnect\")\n\t(generator_version \"10.0\")\n\t(paper \"A4\")\n\t(lib_symbols\n\t)\n)\n"
+fn blank_kicad_sch() -> String {
+    crate::tools::blank_schematic_template()
 }
 
 fn blank_kicad_pcb() -> &'static str {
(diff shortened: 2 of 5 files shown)


mixelpixx (approved):
Approving. Reproduced your ground-truth A/B locally on Windows/KiCAD 10.0.4 with the branch merged onto current main (so #37's tip fix is in on both sides): identical MCP sequence (create project, two Device:R, `connect_pins` R1.2→R2.1, `kicad-cli sch export netlist`) —

- main: `(nets)` block completely empty
- this PR: `Net-(R1-Pad2)` with both R1.2 and R2.1 as nodes, plus correct `unconnected-*` stubs for the two open pins

Code review: consolidating the three template copies into `blank_schematic_template()` and reusing `Symbol::set_instance_path` instead of the hand-built instances node is exactly the right shape; `ensure_root_uuid` repairing legacy files on next edit is a good touch, and the behavior-change notes (file-stem project name, no root `sheet_instances`) are well-reasoned — agreed on both, no need for byte-parity. Hierarchy path format `/<root-uuid>/<sheet-uuid>` matches what eeschema writes.

Full local gate green (235 tests on the merge result), CI green all platforms. This was the most user-impacting schematic bug in the tracker — thanks for the bisection matrix in #29, it made the fix easy to trust.

PR #37, opened 2026-07-15
fix(sexp): pin_endpoint returned the pin body end, not the electrical connection tip

### LINKED ISSUE #30: pin_endpoint() returns the pin body end, not the electrical tip — connect_pins wires miss every pin by one pin-length

tascsystems-andrew:
Fixes #30.

## What

One-expression fix in `konnect-sexp::schematic::pin_endpoint`: stop adding `length` along the pin direction. In the KiCAD symbol format the pin's `(at x y angle)` **is** the wire-connection point — the angle points from that tip toward the body and the drawn line extends `length` mm inward — so the old code landed on the body-attachment end, one pin-length (1.27 mm for `Device:R`) from where KiCAD joins wires.

Every consumer wants the tip and inherits the fix: `connect_pins` (wires now land on the pins instead of overshooting them), `get_schematic_pin_locations` + batch variant, `connect_to_net` snapping, and the design-review audits (6 call sites).

## Evidence the (at) point is the tip

- eeschema's own ERC JSON reports pin positions at the `(at)`-derived coordinates (Device:R at y=80.01 → pins at 76.20/83.82; the old code produced 77.47/82.55).
- `(at) + length·direction` lands exactly on the symbol body outline for Device:R — i.e. the *other* end.
- With this + #36 applied, `connect_pins` between two resistors netlists as `Net-(R1-Pad2)` containing both pins; on main the wire misses both pins and ERC reports them unconnected.

## Tests

`pin_endpoint` had **zero** test coverage (all 225 upstream tests pass unmodified before *and* after this change) — added regression tests pinning the eeschema-verified tip coordinates at rotation 0 and 90. Full suite 227 passed / 0 failed; clippy `-D warnings` clean; fmt applied.

### DIFF
--- a/crates/konnect-sexp/src/schematic.rs
+++ b/crates/konnect-sexp/src/schematic.rs
@@ -291,11 +291,14 @@ fn parse_lib_pin(node: &SexpNode) -> Option<LibPin> {
 /// Compute the schematic-space pin endpoint (where wires connect) for a lib pin
 /// given a component's placement transform.
 pub fn pin_endpoint(pin: &LibPin, t: PinTransform) -> (f64, f64) {
-    // The wire-connection point is at pin_origin + length in pin direction
-    let angle_rad = pin.rotation.to_radians();
-    let tip_x = pin.local_x + pin.length * angle_rad.cos();
-    let tip_y = pin.local_y + pin.length * angle_rad.sin();
-    transform_pin(tip_x, tip_y, t)
+    // In the KiCAD symbol format the pin's (at x y angle) IS the electrical
+    // connection point: the angle points from that tip TOWARD the symbol body,
+    // and the drawn pin line extends `length` mm inward. Adding length here
+    // would land on the body-attachment end — 1 pin-length away from where
+    // KiCAD actually joins wires (eeschema's ERC reports pin positions at the
+    // (at) point, and pin tips land exactly on the body outline only after
+    // adding length — verified against Device:R in the KiCAD 10 libraries).
+    transform_pin(pin.local_x, pin.local_y, t)
 }
 
 // ─── T-Junction detection ─────────────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Approving. Independently verified the format claim against the installed KiCAD 10 libraries: `Device.kicad_symdir/R.kicad_sym` has pins `(at 0 ±3.81 270/90) (length 1.27)` and the body outline edge sits at ±2.54 — exactly `(at) − length`, so `(at)` is unambiguously the electrical tip and the old `+ length·direction` landed on the body end. Your eeschema-ERC coordinate evidence matches the same numbers.

Locally: full workspace gate green (227 tests), clippy/fmt clean. Nice catch that this had zero test coverage before — the pinned eeschema-verified coordinates are exactly the right regression tests. Will do the end-to-end netlist A/B when reviewing #36, which builds on this.

PR #38, opened 2026-07-15
fix(cli): run_erc always returned 0 violations — parse sheets[].violations per the ERC schema

### LINKED ISSUE #31: run_erc always reports 0 violations — parser reads top-level "violations" but ERC reports nest them under sheets[].violations
**Severity: major** — an ERC wrapper that can never report a violation green-lights broken designs (observed reporting 0 where kicad-cli finds 48).

## Repro (d70b561, KiCad 10.0.3)

Any schematic with violations — e.g. two resistors with dangling pins. `run_erc {"severity":"error"}` (or "warning", or omitted) → `{"errors": 0, "total": 0, "violations": []}`. Running the identical command Konnect runs (`kicad-cli sch erc --output x.json --format json <file>`) and reading the same report file shows the violations are present.

## Cause

`crates/konnect-core/src/tools/cli.rs:114-118` — `parse_erc_json` does `raw.get("violations")` on the report **root**. KiCad's ERC schema (https://schemas.kicad.org/erc.v1.json) has no top-level `violations` key — they're nested per sheet at `sheets[].violations[]`, with positions on the affected `items[]`. (Top-level `violations` is the **DRC** report's shape — drc.v1.json — so `run_drc`'s parser is fine; the ERC one looks copy-pasted from it.) The `None` arm is taken unconditionally → `Vec::new()` for every input, 100% deterministic.

## Fix

PR incoming: flatten `sheets[].violations`, carry the sheet path, take pos from the violation's first item, and append the item description so "Pin not connected" becomes "Pin not connected: Symbol R1 Pin 1 [Passive, Line]". Unit tests pin the real report shape (and that DRC-shaped input yields nothing).

Found in the same session as #29 — the two masked each other nicely: the netlist was empty *and* the ERC that would have flagged it always said clean.

tascsystems-andrew:
Fixes #31.

## What

`parse_erc_json` read a top-level `"violations"` array, but KiCAD's ERC report (https://schemas.kicad.org/erc.v1.json) nests violations per sheet under `sheets[].violations`, with positions on the affected `items[]` — top-level `violations` is the **DRC** schema's shape (`run_drc`'s parser is correct and untouched). The `None` arm was taken unconditionally, so `run_erc` reported `{"errors": 0, "total": 0}` for every schematic ever checked.

Now: flattens `sheets[].violations`, carries the sheet path, takes the position from the violation's first item, and appends the item description — `"Pin not connected"` becomes `"Pin not connected: Symbol R1 Pin 1 [Passive, Line]"`, actionable without opening eeschema (happy to drop that last part if you'd rather keep the description verbatim).

## Verification

- Live, KiCad 10.0.3: divider with two open pins — `run_erc` now returns both violations with named offenders; upstream main returns 0. Report file cross-checked against what the wrapper returns.
- Unit tests pin the real captured report shape, plus guards: empty report, empty sheets, and DRC-shaped input all yield no violations.
- `cargo test --workspace --lib --tests` 227 passed / 0 failed; clippy `-D warnings` clean; fmt applied.

### DIFF
--- a/crates/konnect-core/src/tools/cli.rs
+++ b/crates/konnect-core/src/tools/cli.rs
@@ -112,24 +112,51 @@ pub async fn run_erc(cli: &str, schematic: &Path) -> Result<Vec<ErcViolation>> {
 }
 
 fn parse_erc_json(raw: &serde_json::Value) -> Vec<ErcViolation> {
-    let arr = match raw.get("violations").and_then(|v| v.as_array()) {
-        Some(a) => a,
-        None => return Vec::new(),
+    // KiCAD's ERC report (https://schemas.kicad.org/erc.v1.json) nests
+    // violations per sheet — { "sheets": [ { "path": …, "violations": […] } ] }
+    // — with positions on the affected items. There is no top-level
+    // "violations" key (that's the DRC report's shape), so reading one here
+    // silently returned zero violations for every schematic.
+    let Some(sheets) = raw.get("sheets").and_then(|s| s.as_array()) else {
+        return Vec::new();
     };
 
-    arr.iter()
-        .map(|v| ErcViolation {
-            severity: v["severity"].as_str().unwrap_or("error").to_string(),
-            description: v["description"].as_str().unwrap_or("").to_string(),
-            sheet: v["sheet"].as_str().map(String::from),
-            pos: v.get("pos").and_then(|p| {
-                Some(ErcPos {
-                    x: p["x"].as_f64()?,
-                    y: p["y"].as_f64()?,
-                })
-            }),
-        })
-        .collect()
+    let mut out = Vec::new();
+    for sheet in sheets {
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Approving. Reproduced end-to-end on Windows/KiCAD 10.0.4: drove the branch-built server over MCP stdio, created a project, placed a Device:R with floating pins, and `run_erc` now returns both `Pin not connected` violations with named offenders (`Symbol R1 Pin 1/2 [Passive, Line]`), sheet path, and positions — where main structurally cannot (the top-level `violations` key it reads doesn't exist in ERC reports, only DRC ones). The schema reference and captured-report unit tests pin exactly the right shape, and the alien-input guards (empty/DRC-shaped) are a nice touch. Keeping the appended per-item description — actionable beats verbatim here.

Bonus: the live ERC output places R's pins at ±3.81 from center, independently re-confirming #37's tip fix. Full local gate green (227 tests), CI green on all platforms.

Issue #42, opened 2026-07-16
batch_connect_to_net and connect_passthrough write (net_label) blocks that make the schematic unopenable in KiCAD


tascsystems-andrew:
**Severity: critical — data loss.** Two tools leave the schematic in a state KiCAD refuses to load.

## Repro (d70b561, KiCad 10.0.3)

`format_net_label()` (`crates/konnect-sexp/src/schematic.rs:353`) emits:

```
(net_label "SIGNAL"
  (at 100 80 0)
  …)
```

There is no `net_label` in KiCAD's schematic format — a plain net label is `(label …)`. Controlled test, two files identical but for that one tag:

```
$ kicad-cli sch export netlist good.kicad_sch      # (label "MIDNET" …)
  → net /MIDNET formed, exit 0
$ kicad-cli sch export netlist konnect.kicad_sch   # (net_label "MIDNET" …)
  Failed to load schematic                          # exit != 0, no netlist
```

Callers: `batch_connect_to_net` (`sch_batch.rs:329`) and `connect_passthrough` (`sch_batch.rs:657`). Any project either tool touches stops opening — in eeschema too, not just the CLI.

`add_schematic_net_label` is unaffected: it goes through konnect-schematic-editor, whose `Label::to_sexp` correctly writes `label`. The two label representations drifted apart.

## Blast radius beyond the corruption

The same phantom tag is in two readers, so it fails in the other direction as well:

- `extract_labels()` (`schematic.rs:112-118`) matches `net_label` → **every plain label is invisible** to everything built on it: the union-find net graph in `sch_analysis` (so label-joined nets simply don't exist for connectivity analysis), the batch tools, `list_schematic_labels`.
- `design_review.rs:832` and `:953` search for the literal `(net_label "` in power-rail audits → those checks can never fire.
- `delete_schematic_net_label` matches it → can't delete the label type the add tool writes (that's #33).

So Konnect can round-trip its own invalid labels, but can neither read real ones nor produce a loadable file from `batch_connect_to_net`.

## Fix

PR #41 (stacked on #40) switches the writer, both readers, and the design-review literals to `label`, with a format → extract round-trip test. Filed separately from #33 because the corruption is the more severe half and worth its own tracking; the deleter can't be made correct while the writer still emits an invalid tag, so they're fixed together.

[...]

Issue #33, opened 2026-07-15
delete_schematic_net_label can never delete plain (label) blocks and can silently delete the wrong same-named global/hierarchical label


tascsystems-andrew:
**Severity: major** for AI-driven workflows — silent deletion of the wrong object.

## Repro (d70b561)

1. `add_schematic_net_label {"net":"VCC","x":100,"y":100}` and again at (200,100) — default `label_type` writes plain `(label "VCC" ...)` blocks.
2. `delete_schematic_net_label {"net":"VCC","x":200,"y":100}` → **error "Label 'VCC' not found"**, file unchanged. The delete tool can never remove the label type its own add tool writes by default.
3. With **global** labels the coordinate matching works correctly (closest same-named label is deleted — tested). But: when a same-named global/hierarchical label exists elsewhere, and the requested name+coords don't resolve inside a matching block, the backwards string-walk can select an unrelated block — we observed a delete that reported `{"deleted_label":"VBUS","at":{"x":300,"y":300}}` while actually removing the global label at (100,100).

## Cause

`sch_wiring.rs:725` — `handle_delete_net_label`'s block-start pattern list is `["(net_label", "(global_label", "(hierarchical_label"]`. It omits `"(label"`, the actual KiCad token for plain net labels (`(net_label` is a Konnect-internal token that KiCad never writes). Plus the backwards search from a name-string occurrence is unbounded — it can land on a block that doesn't actually contain the matched name.

## Suggested fix

Parse instead of string-walk: extend `konnect_sexp::schematic::extract_labels` (schematic.rs:115-118) to include the `label` tag, pick the name match nearest the requested (x,y) within a tolerance (error otherwise — never delete a global minimum far away), and delete by that block's uuid. Minimal patch: add `"(label"` to the pattern list and verify the candidate block actually encloses the name occurrence.

Same session as #29-#31.

Issue #32, opened 2026-07-15
edit_schematic_component silently no-ops on eeschema-saved (tab-indented) files; batch variant errors "Field not found"


tascsystems-andrew:
**Severity: major** — edits to any schematic last saved by eeschema (i.e. the primary interop case) are silently dropped.

## Repro (d70b561, KiCad 10.0.3)

Copy any eeschema-authored schematic (e.g. the bundled `template/Arduino_Nano/Arduino_Nano.kicad_sch`, generator "eeschema", tab-indented) and call:

- `edit_schematic_component {"reference":"J1","value":"AnalogEdited"}` → `isError: false`, `"changes": []`, file byte-identical. No error surfaced anywhere.
- `batch_edit_schematic_components` with the same edit → `isError: false`, `updated_count: 0`, errors: `["Field 'Value' not found on 'J1'"]` — misleading, since J1's Value property is plainly present.

Control: the identical calls succeed on a Konnect-authored (2-space-indented) file.

## Cause

Both edit paths locate the enclosing symbol block by scanning backwards for the literal string `"\n  (symbol"` — newline + exactly two spaces (`sch_components.rs:473-477` in `handle_edit_schematic_component`'s `update_field`, and the batch path). eeschema writes tabs, so `rfind` returns `None` and the lookup dead-ends — single-edit swallows it, batch mislabels it as a missing field. There look to be ~a dozen call sites sharing the `"\n  (symbol"` literal.

## Suggested fix

Indentation-agnostic anchor (`\n[ \t]+\(symbol`) or — better — resolve the symbol block span through the existing konnect-sexp parser instead of text heuristics, factored into one shared helper. Independently: `edit_schematic_component` should report an error when a requested change couldn't be applied instead of returning success with `changes: []`.

Same session as #29/#30/#31.

PR #44, opened 2026-07-16
fix(sch): derive label justify from rotation — 180°/270° labels rendered backwards

### LINKED ISSUE #43: Labels render backwards at 180°/270°: add_schematic_net_label writes no (effects), and rotate_schematic_label never touches justify
**No Konnect path currently produces a correctly-rendering 180°/270° label.** Two tools have to be wrong together for that, and they are.

## Background

A label's `(at … ROT)` aims the connection arrow; `(justify …)` inside `(effects)` decides which way the **text** runs. Out of step, the label attaches correctly and draws its text back over whatever it points at. KiCAD's own SVG export, two 180° global labels with anchor at x=100:

```
(justify right)   →  <text x="98.5712"  text-anchor="end">    text runs WEST, away  ✓
no (effects)      →  <text x="101.4288" text-anchor="start">  text runs EAST, backwards  ✗
```

## The two bugs

1. **`add_schematic_net_label` writes no effects block.** `Label::new` / `GlobalLabel::new` / `HierarchicalLabel::new` set `effects: None` (`konnect-schematic-editor/src/schematic/label.rs`), and the handler only sets `at.rotation` after. So every label it creates takes KiCAD's left-justified default and renders backwards at 180°/270°.

2. **`rotate_schematic_label` has no justify code at all.** Its tool description says *"Rotate a net label to a new angle and update its justify direction accordingly"* (`sch_wiring.rs:147`), but the handler only rewrites `(at …)`. Before the fix, the string `justify` appeared exactly **once** in the whole codebase — in that description. So it does nothing about justify whether the block is present or absent, which also means it can't repair anything the add tool created.

Additionally, `format_net_label` hardcoded `(justify left)` for every rotation and omitted the `bottom` alignment eeschema gives plain labels.

## Ground truth for the mapping

Sampled from 692 labels across KiCAD 10-authored schematics — the pairing is exact, no counter-examples:

| rotation | justify | n |
|---|---|---|
| 0 | left | 297 |
| 90 | left | 6 |
| 180 | right | 298 |
| 270 | right | 5 |

(plain `label` blocks additionally carry `bottom`: `(justify left bottom)` / `(justify right bottom)`)

## Fix

PR #44 (stacked on #41): one shared `label_justify(rotation)` in konnect-sexp, `set_rotation()` on the three label types in konnect-schematic-editor that builds-or-refaces `(effects)` while preserving an existing font, and a `rotate_schematic_label` that inserts the block where eeschema puts it (before the uuid) when absent. Re-applying a label's current rotation repairs legacy labels in place, uuid preserved.

Found while building a real amp schematic: three global labels were electrically correct but rendered their text east into the relay bodies they pointed at, only visible by rendering the sheet.

tascsystems-andrew:
Fixes #43. **Stacked on #41** (which is stacked on #40) — merge those first and this becomes a 3-file diff.

## What

A label's `(at … ROT)` aims the connection arrow; `(justify …)` inside `(effects)` decides which way the **text** runs. Konnect never kept them in step, so there was no way to make a correctly-rendering 180°/270° label:

- **`add_schematic_net_label`** created labels with `effects: None` — no block at all → KiCAD's left-justified default → text runs backwards over whatever the label points at.
- **`rotate_schematic_label`** promises *"update its justify direction accordingly"* in its own description but had **no justify code**: it only rewrote `(at …)`. Before this PR the string `justify` appeared exactly once in the codebase — in that description.
- **`format_net_label`** hardcoded `(justify left)` at every rotation, and omitted the `bottom` eeschema gives plain labels.

## Proof, from KiCAD's renderer

Two 180° global labels, anchor at x=100:

```
(justify right)   →  <text x="98.5712"  text-anchor="end">    runs WEST, away  ✓
no (effects)      →  <text x="101.4288" text-anchor="start">  runs EAST, backwards  ✗
```

## The mapping

`konnect_sexp::schematic::label_justify`: 0/90 → `left`, 180/270 → `right` — what eeschema itself writes across 692 sampled labels (0→left ×297, 90→left ×6, 180→right ×298, 270→right ×5, no counter-examples). Plain labels also get `bottom`.

## Changes

- `set_rotation()` on `Label` / `GlobalLabel` / `HierarchicalLabel` sets the anchor and builds-or-refaces `(effects)`; an existing block keeps its font, only the justify token is swapped. `add_schematic_net_label` uses it.
- `rotate_schematic_label` now updates justify — replacing an existing one, adding one to an `(effects)` that lacks it, or inserting the whole block where eeschema puts it (before the uuid, at that line's indentation). **This repairs labels written by earlier versions**: re-applying a label's current rotation adds the missing block, uuid preserved.
- Its lookup moves to the structured `find_label_blocks` scan from #41, so it inherits those fixes too: plain `(label)` blocks are findable at all, and exact position matching means a same-named label elsewhere is never rotated by mistake.

## Verified

macOS 26.5 / KiCad 10.0.3, by rendering: labels added at 180°, labels rotated 0°→180° through the tool, and a legacy no-effects label repaired by re-applying its rotation all render `text-anchor="end"`. 12 handler tests + 3 mapping tests; full suite 254 passed / 0 failed; clippy `-D warnings` clean; fmt applied.


mixelpixx (approved):
Approving all three commits in this stack (this supersedes #40 and #41, which I'll close as included here).

**Verified live on Windows/KiCAD 10.0.4** with the branch-built server over MCP stdio:
- `add_schematic_net_label` at 0/90/180/270 → file carries `(label …)` (no `net_label` anywhere) with justify left/left/right/right exactly matching your 692-label survey; `kicad-cli sch export svg` loads and renders the file, which the old `net_label` tag made impossible (#42).
- `delete_schematic_net_label` removes exactly the targeted plain label at its coordinates (#33).
- `rotate_schematic_label` to 180° updates both the rotation and the justify to `right` in one step (#43).
- The tab-indent lookup (#32) is pinned by the eeschema-style tab-indented fixtures in the new unit tests.

**Two commits added on top** (thanks for flipping maintainer-edit on): restored the four `crates/schematic-viewer/gen/schemas/*.json` files — they're Tauri build artifacts tracked on main and their removal is out of scope for label fixes — and a `cargo fmt` touch-up on an import split from the rebase. Everything else is yours as pushed. Appreciate the proactive rebase onto the #36 merge, it landed while I was mid-review.

Full local gate green (264 tests), CI green on all platforms.

PR #46, opened 2026-07-16
feat(packaging): ship macOS and Linux PCM packages

### LINKED ISSUE #45: PCM package ships only Windows binaries — the recommended install gives macOS and Linux users a plugin that can't run
The README calls the Plugin and Content Manager the **recommended** install method. The package it points at contains only Windows binaries:

```
$ unzip -l konnect-pcm-v0.1.3.zip
 15542272  plugins/bin/konnect.exe
  8401408  plugins/bin/schematic-viewer.exe
```

So a macOS or Linux user who follows the documented instructions installs 24 MB of `.exe` and gets a plugin that cannot run. Nothing warns them, because `metadata.json` declares no `platforms` — which PCM reads as "compatible with every OS".

This is the gap left behind #4 ("Can't install on KiCad 10.0 macOS"): that report was closed once the schema-invalid metadata was fixed, so the *install* succeeds now — it just doesn't produce a working plugin on anything but Windows.

## Cause

`packaging/build-pcm.sh` — the macOS/Linux packager that came in with #7 — is never called. `.github/workflows/release.yml`'s `pcm-package` job runs only on `windows-latest`, only invoking `build-pcm.ps1`. And neither builder stamps `platforms`, though KiCAD's schema supports it:

```json
"platforms": { "type": "array", "items": { "enum": ["windows", "macos", "linux"] } }
```

## Fix

PR #46: per-OS matrix in the release job so all three packages get built with the right builder; universal (arm64 + x86_64) macOS binaries via `lipo`, since PCM has no architecture dimension and an arm64-only "macos" package would break Intel Macs; `platforms` stamped by both builders; and a `validate-pcm.py` gate asserting a package's bundled binary matches the platform it declares — which fails against the current v0.1.3 asset:

```
FAIL: version 0.1.3 declares no 'platforms' — PCM would offer this native
      package to every OS
```

Verified end to end on macOS 26.5 / KiCad 10.0.3: universal binaries, correct entrypoint, passes validation, binary runs.

This is the "release packaging" half of the macOS/Linux roadmap item — signing/notarization is separate and not addressed here.

tascsystems-andrew:
Fixes #45. Independent of my other PRs — this one is packaging/CI only, no Rust changes.

## Why

`konnect-pcm-v0.1.3.zip` contains only `konnect.exe` + `schematic-viewer.exe`, and the README calls PCM the recommended install. Every macOS and Linux user who follows it gets 24 MB of Windows binaries and a plugin that can't run — with no warning, because the package declares no `platforms` and PCM reads that as "every OS".

`packaging/build-pcm.sh` (from #7) already does the macOS/Linux packaging correctly. It was simply never wired in: the release job runs only `build-pcm.ps1` on `windows-latest`.

## What

- **`pcm-package` → per-OS matrix** (windows-latest / macos-latest / ubuntu-latest), each using the builder for its platform. `fail-fast: false` so one platform's failure doesn't mask the others.
- **Universal macOS binaries.** The job builds both Apple targets and `lipo`s them. PCM has no architecture dimension — only platform — so an arm64-only "macos" package would silently break Intel Macs.
- **`platforms` stamped** by both builders (`--platform` / `-Platform`). A package bundles one native binary, so it's never valid for all platforms.
- **Per-platform names**: `konnect-pcm-v<version>-<platform>.zip`. kicad-addons needs a distinct `download_url` per platform package anyway, and the three assets can't collide in a release. ⚠️ **This renames the Windows asset** (was `konnect-pcm-v<version>.zip`) — happy to keep the old name and only suffix the new ones if you'd rather.
- **A gate that would have caught this**: `validate-pcm.py` now requires exactly one declared platform and checks the bundled binary's magic bytes (PE / Mach-O / ELF, incl. universal) against it. Against the current release:
  ```
  FAIL: konnect-pcm-v0.1.3.zip: version 0.1.3 declares no 'platforms' —
        PCM would offer this native package to every OS
  ```
- **CI now exercises `build-pcm.sh`** (it's load-bearing for two of three packages), with stub binaries carrying real executable magic, plus a negative test asserting a Windows binary in a Linux package is rejected. The existing Windows stub is now `MZ`-prefixed so it passes the new format check.

## Verified

macOS 26.5 (Apple Silicon) / KiCad 10.0.3, building the macOS package exactly as the workflow does:

```
plugins/bin/konnect          → x86_64 arm64   (universal, +x, `--version` runs)
plugins/bin/schematic-viewer → x86_64 arm64
metadata.json                → platforms: ["macos"]
plugin.json                  → entrypoint bin/konnect
validate-pcm.py              → OK
```

The new CI packaging steps were run locally against this repo (metadata validation, Linux stub package, and the negative test) — all pass. The release-workflow matrix itself can only run on a tag; the packaging commands it invokes are the ones verified above.

Signing/notarization is the other half of the macOS roadmap item and isn't touched here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,10 +63,38 @@ jobs:
       - run: pip install jsonschema
       - name: Validate repository metadata
         run: python packaging/validate-pcm.py --metadata packaging/metadata.json
-      - name: Assemble PCM zip with a stub binary and validate structure
+
+      # Stub binaries carry each platform's executable magic: the validator
+      # checks a package's bundled binary against the platform it declares,
+      # which is what stops a Windows-only zip from being offered to every OS.
+      - name: Assemble Windows PCM zip with a stub binary and validate structure
         shell: pwsh
         run: |
-          New-Item -ItemType File -Path stub-konnect.exe | Out-Null
-          Set-Content stub-konnect.exe "stub"
-          ./packaging/build-pcm.ps1 -Version 0.0.0 -BinaryPath stub-konnect.exe -OutDir dist
-          python packaging/validate-pcm.py --zip dist/konnect-pcm-v0.0.0.zip
+          [System.IO.File]::WriteAllBytes("stub-konnect.exe", [byte[]](0x4D, 0x5A, 0x90, 0x00))
+          ./packaging/build-pcm.ps1 -Version 0.0.0 -BinaryPath stub-konnect.exe `
+            -Platform windows -OutDir dist
+          python packaging/validate-pcm.py --zip dist/konnect-pcm-v0.0.0-windows.zip
+
+      # build-pcm.sh is what the macOS and Linux release packages are built
+      # with, so it gets the same structural check.
+      - name: Assemble Linux PCM zip with a stub binary and validate structure
+        run: |
+          set -euo pipefail
+          printf '\177ELF\002\001\001\000' > stub-konnect
+          ./packaging/build-pcm.sh --version 0.0.0 --binary stub-konnect \
+            --platform linux --out dist
+          python packaging/validate-pcm.py --zip dist/konnect-pcm-v0.0.0-linux.zip
+
+      - name: Reject a package whose binary contradicts its platform
+        run: |
+          set -euo pipefail
+          # The exact mistake v0.1.3 shipped: a Windows binary in a package
+          # non-Windows users are offered. The gate must fail here.
+          printf 'MZ\220\000' > stub-konnect-win
+          ./packaging/build-pcm.sh --version 0.0.0 --binary stub-konnect-win \
+            --platform linux --out dist-bad
+          if python packaging/validate-pcm.py --zip dist-bad/konnect-pcm-v0.0.0-linux.zip; then
+            echo "::error::validator accepted a Windows binary in a linux package"
+            exit 1
+          fi
+          echo "OK: mismatched binary/platform correctly rejected"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -70,12 +70,29 @@ jobs:
           path: ${{ env.BINARY_NAME }}-${{ github.ref_name }}-${{ matrix.target }}.*
 
   pcm-package:
-    name: KiCAD PCM plugin package
-    runs-on: windows-latest
+    name: KiCAD PCM plugin package (${{ matrix.platform }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: windows
+            os: windows-latest
+          - platform: macos
+            os: macos-latest
+          - platform: linux
+            os: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
       - uses: dtolnay/rust-toolchain@stable
+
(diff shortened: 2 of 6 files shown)


tascsystems-andrew:
### Verified on a real Mac: KiCAD's PCM accepts the macOS package

Built the macOS package with `build-pcm.sh` exactly as the workflow in this PR does, and fed it to **Plugin and Content Manager → Install from File** on macOS 26.5 (Apple Silicon), KiCad 10.0.3. PCM parsed it, rendered the package, and marked it:

| Version | Install Size | Compatible | Status |
|---|---|---|---|
| 0.1.3 | 53.7 MB | ✔ | stable |

That ✔ is PCM evaluating `kicad_version: 10.0` + `platforms: ["macos"]` against the host and accepting it — the check that today's `konnect-pcm-v0.1.3.zip` can't make, because it declares no platforms at all.

Package contents:

```
plugins/bin/konnect           x86_64 arm64   -rwxr-xr-x   (universal, runs: konnect 0.1.3)
plugins/bin/schematic-viewer  x86_64 arm64   -rwxr-xr-x
metadata.json                 platforms: ["macos"], install_size 53718843
plugins/plugin.json           entrypoint: bin/konnect
validate-pcm.py               OK
```

The zip stores Unix modes (`3.0 unx`, `-rwxr-xr-x`), so extraction preserves the executable bit — worth stating explicitly since a macOS/Linux package whose binary lands without `+x` would install "successfully" and still be unusable.

**Scope of this test:** the package was staged in PCM but Apply Pending Changes wasn't run, so this verifies PCM's acceptance/compatibility evaluation and the package payload — not the extraction step itself. 53.7 MB install is the universal binaries carrying both architectures, not bloat.

### One fix since the first push

`build-pcm.sh` was stamping placeholder download fields into the shipped metadata (`download_size: 1`, an all-zeros `download_sha256`, an `example.invalid` URL) — PCM displayed the first of those to the user as **"Download Size: 1 B"**. `build-pcm.ps1` deliberately omits those fields ("placeholder values are lies"); the shell builder now does the same, so the two agree and users see honest metadata. Caught by looking at the actual PCM dialog.

mixelpixx (approved):
Approving. Verified the Windows leg end-to-end locally: `build-pcm.ps1 -Platform windows` on a real binary produces a package that passes the new validator; an ELF stub smuggled into a windows-declared package is rejected with a precise message; and — the best part — the validator run against the actual shipped `konnect-pcm-v0.1.3.zip` fails with the no-platforms error, so this gate really would have caught #45. The macOS universal-binary reasoning (PCM has platform but no architecture dimension) is correct, and `fail-fast: false` on the matrix is the right call. Workflow security looks fine: plain `pull_request` events, no secret exposure beyond setup-protoc's token.

Keeping the per-platform rename including Windows — consistency beats asset-name continuity here, and the release job's `merge-multiple` download picks the new names up fine.

Two release-time notes, neither blocking: `packaging/metadata.json`'s stamped `download_url` still points at the old single-asset name (accurate for v0.1.3; the next release's stamping step must switch to `-windows`), and the macOS section #27 just added says the PCM zip is Windows-only — I'll touch that sentence up on main now that this fixes it.

macOS/Linux legs verified by your documented run + the new CI steps exercising `build-pcm.sh` on ubuntu (green).

PR #47, opened 2026-07-17
fix(config): resolve KICAD_API_SOCKET reliably regardless of launcher

### LINKED ISSUE #39: IPC API never connects (socket live, env var set, KiCAD running)
## Summary

Konnect 0.1.3 cannot reach the KiCAD IPC API even when:

- KiCAD 10.0 is running with the IPC API enabled
- The unix socket `/tmp/kicad/api.sock` exists, is world-connectable, and accepts connections from a manual Python test
- `KICAD_API_SOCKET=/tmp/kicad/api.sock` is present in the Konnect process environment

Every IPC-dependent call fails. `open_project` returns `kicad_ui_running: false` with `dur_ms: 0` (never attempts a connection); `save_project` errors with "IPC socket path not configured".

## Environment

- Konnect: 0.1.3 (`/home/USER/.local/share/kicad/10.0/3rdparty/plugins/com_github_mixelpixx_konnect/bin/konnect --version` → `konnect 0.1.3`)
- KiCAD: 10.0 (paths under `~/.local/share/kicad/10.0/...`)
- Client: Claude Code CLI
- OS: Linux 6.17.0-1028-oem (Ubuntu)
- MCP transport: stdio, project-local `.mcp.json`

## `.mcp.json`

```json
{
  "mcpServers": {
    "konnect": {
      "command": "/home/USER/.local/share/kicad/10.0/3rdparty/plugins/com_github_mixelpixx_konnect/bin/konnect",
      "env": { "RUST_LOG": "info", "KICAD_API_SOCKET": "/tmp/kicad/api.sock" }
    }
  }
}
```

## Reproduction

1. Start KiCAD 10.0 with the IPC API enabled. Confirm the socket is live and connectable:

   ```
   $ ls -la /tmp/kicad/api.sock
   srwxrwxr-x 1 USER USER 0 /tmp/kicad/api.sock
   $ python3 -c "import socket; s=socket.socket(socket.AF_UNIX); s.settimeout(2); s.connect('/tmp/kicad/api.sock'); print('OK')"
   OK
   $ ss -lx | grep kicad
   u_str LISTEN 0 128 /tmp/kicad/api.sock 258919 * 0
   $ pgrep -af kicad
   41122 /usr/bin/kicad
   ```

2. Spawn Konnect via Claude Code with `KICAD_API_SOCKET=/tmp/kicad/api.sock` in the env. Confirm the var reaches the process:

   ```
   $ cat /proc/$(pgrep -n konnect)/environ | tr '\0' '\n' | grep -i kicad
   KICAD_API_SOCKET=/tmp/kicad/api.sock
   ```

3. Call `open_project`:

   ```json
   {
     "ipc_address": "",
     "kicad_ui_running": false,
     "message": "KiCAD IPC is not reachable. Start KiCAD and enable the IPC API, or work in file-only mode."
   }
   ```

4. Call `save_project` to force a real IPC round-trip:

   ```json
   {
     "error": {
       "kind": "handler_error",
       "reason": "KiCAD IPC socket path not configured. To fix: (1) in KiCAD, enable Edit > Preferences > Plugins > 'Enable KiCad API' and copy the listed ipc:// address; (2) paste it into the 'IPC Socket' field of the Konnect settings dialog ... Alternatively set ipc_socket_path in konnect-settings.json or launch via KiCAD (which sets KICAD_API_SOCKET). Full guide: https://github.com/mixelpixx/Konnect/blob/main/docs/TROUBLESHOOTING.md"
     }
   }
   ```

## Expected

Konnect connects to the unix socket at `$KICAD_API_SOCKET` and `open_project` / `save_project` succeed.

## Actual

- `open_project` returns `kicad_ui_running: false` with `dur_ms: 0` — it never attempts a socket connection (no latency, no error path).
- `save_project` errors with "IPC socket path not configured" despite `KICAD_API_SOCKET` being present in the process environment.

## Notes

- The env var name `KICAD_IPC_SOCKET` referenced in some docs/examples is **not** read by 0.1.3; the recognised name is `KICAD_API_SOCKET` (per the error string itself).
- A second Konnect process is spawned by KiCAD itself with `--config settings.json`. Both run concurrently — only the Claude-Code-spawned one is the MCP stdio server. Coexistence of the two is fine (the socket allows multiple clients), but worth confirming Konnect isn't trying to bind a lock the KiCAD-spawned instance already holds.

   ```
   $ ps -ef | grep konnect | grep -v grep
  USER  95633 ... konnect                                          # MCP stdio server (child of claude)
  USER 101633 ... konnect --config .../settings.json               # spawned by kicad
   ```

- KiCAD confirmed running: `pgrep -af kicad` → `41122 /usr/bin/kicad`.

Robin-OwO:
## What

Config::load() only checked ./konnect.toml, ./settings.json (relative to CWD) and the OS-standard config dir. It never checked the plugin directory, which is where the Settings dialog actually writes settings.json. That only worked before because KiCAD itself passes --config when it launches the server. Any other launcher had no way to find that file and fell through to Config::default().

There's a second problem with the KICAD_API_SOCKET fallback. It's a serde field default, so it only fires when the key is missing from the file entirely, not when it's present but blank. A settings.json with ipc_address set to "" permanently blocks the env var from being used.

## Proof, from issue #39

KICAD_API_SOCKET was confirmed present in the process environment (checked via /proc/PID/environ), yet open_project still returned:

    { "ipc_address": "", "kicad_ui_running": false, "message": "KiCAD IPC is not reachable..." }

dur_ms was 0, meaning no connection was ever attempted, because ipc_address was already empty before any socket code ran.

## Changes

Config::load() now also checks <exe_dir>/settings.json and <exe_dir>/../settings.json. This covers the actual KiCAD PCM install layout (<plugin_dir>/bin/konnect sitting next to <plugin_dir>/settings.json) regardless of the caller's working directory.

A blank ipc_address is now always rechecked against KICAD_API_SOCKET before falling back to nothing, whether the blank came from Config::default() or from a loaded file.

## Verified

Tested on Windows 11 with KiCad 10.0.4 against a real running instance. open_project now returns the correct ipc_address and kicad_ui_running: true.

cargo test --workspace --lib --tests passes. clippy with -D warnings is clean. fmt applied with no changes needed.

Fixes #39.

### DIFF
--- a/crates/konnect/src/config.rs
+++ b/crates/konnect/src/config.rs
@@ -79,20 +79,33 @@ fn default_log_level() -> String {
 impl Config {
     /// Load config from the default search path.
     pub fn load() -> Result<Self> {
-        let config_paths = [
+        let mut config_paths = vec![
             PathBuf::from("konnect.toml"),
             PathBuf::from("settings.json"),
-            dirs_config_path(),
         ];
+        config_paths.extend(exe_relative_settings_paths());
+        config_paths.push(dirs_config_path());
 
+        let mut config = None;
         for path in &config_paths {
             if path.exists() {
-                return Self::load_from(path);
+                config = Some(Self::load_from(path)?);
+                break;
             }
         }
 
-        // No config file found; use defaults
-        Ok(Config::default())
+        let mut config = config.unwrap_or_default();
+
+        // Env var wins over an unset/blank ipc_address either way.
+        if config.ipc_address.is_empty() {
+            if let Ok(sock) = std::env::var("KICAD_API_SOCKET") {
+                if !sock.is_empty() {
+                    config.ipc_address = sock;
+                }
+            }
+        }
+
+        Ok(config)
     }
 
     /// Load config from a specific file path. Auto-detects JSON vs TOML by extension.
@@ -127,6 +140,20 @@ impl Default for Config {
             log_level: default_log_level(),
         }
     }
+}
+
+/// settings.json next to the binary, and one dir up (covers <plugin_dir>/bin/konnect).
+fn exe_relative_settings_paths() -> Vec<PathBuf> {
+    let mut paths = Vec::new();
+    if let Ok(exe) = std::env::current_exe() {
+        if let Some(exe_dir) = exe.parent() {
+            paths.push(exe_dir.join("settings.json"));
+            if let Some(parent_dir) = exe_dir.parent() {
+                paths.push(parent_dir.join("settings.json"));
+            }
+        }
+    }
+    paths
 }
 
 fn dirs_config_path() -> PathBuf {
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Approving — verified live on Windows in a replica of the PCM layout (`plugin/bin/konnect.exe` + `plugin/settings.json`, server launched from a foreign cwd):

- blank `ipc_socket_path` in the plugin settings.json + `KICAD_API_SOCKET` set → `open_project` reports the env socket (the exact #39 failure, fixed)
- explicit file value + env set → file wins, as it should
- blank + no env → empty, falls through to the guidance error

Case 2 also proves the exe-relative lookup itself works: the file was picked up from one directory above the binary with the cwd elsewhere, which `main` could never do.

One gap I'm fixing in a follow-up commit on main rather than blocking you: the `--config <path>` branch in `main.rs` calls `Config::load_from` directly, so a KiCAD-launched server (`--config` + blank `ipc_socket_path` + `KICAD_API_SOCKET` in env) still misses the fallback. I'll hoist your env-recheck into a shared `Config` method applied on both paths, and point the second unit test at the real method instead of the inlined copy of the logic.

Full gate green locally (229 tests), CI green on all platforms. Thanks for the clean diagnosis on #39 — the `/proc/PID/environ` evidence made this easy to trust.

PR #48, opened 2026-07-17
Remove Docker packaging

gcolonese:
## Why

Konnect's core value is talking to a *running* KiCAD over its IPC API. KiCAD 10's IPC transport is NNG over `ipc://` -- a Unix domain socket on Linux, a Windows named pipe on Windows (`crates/konnect-ipc/src/client.rs`) -- and KiCAD never exposes it over TCP.

A Linux container cannot open the host's Windows named pipe, nor a host Unix socket, the way it can reach a host TCP port -- there's no bridge for local IPC transports the way there is for TCP. So from inside the Docker image:

- All the live-IPC PCB tools fail (no reachable KiCAD).
- `kicad-cli` export tools fail too (no KiCAD install in the image).

That leaves only the file-based schematic-edit tools working in the container -- a small subset of the toolset -- while Docker adds real install overhead (install Docker, build/pull an image, manage volume mounts). It will never be the best install path for this project, so shipping it in-tree just sets up new users to hit a wall. The PCM-installed native binary remains the supported, full-featured install.

## What this removes

- `Dockerfile`, `docker-compose.yml`, `.dockerignore`
- `docker/konnect.toml`, `docker/smoke-test.sh`
- The README "With Docker" install section
- The `config.rs` unit test that parsed the shipped `docker/konnect.toml`

Confirmed no `.github/workflows/*` reference Docker, so no CI cleanup needed.

## What this keeps

`56f2301` ("Harden hosted HTTP default to loopback-only publish"), the immediate follow-up in the same PR, is folded into this removal rather than kept. I checked its diff carefully because a genuine change to the *server's own* default HTTP bind address (e.g. defaulting `http_address` to `127.0.0.1` in Rust) would be a real security fix independent of Docker and worth keeping. It isn't that -- the diff only touches `docker-compose.yml`'s port publish mapping (`3000:3000` -> `127.0.0.1:3000:3000`) and the corresponding README docker instructions. No `crates/konnect/src/config.rs` default changed. Since it's docker-only, it goes with the rest of the packaging.

## Testing

- `cargo test --workspace --lib --tests` -- passes
- `cargo clippy --workspace -- -D warnings` -- clean
- `cargo fmt --all` -- no changes
- Grepped the repo for `Dockerfile`, `docker-compose`, `docker/`, `.dockerignore`, `smoke-test` -- no dangling references remain outside this diff.


mixelpixx (approved):
Approving. You're right that this can never be a good install path: KiCAD's IPC is NNG over a local-only transport (Unix socket / Windows named pipe) with no TCP option, so a container can't reach a host KiCAD, and without KiCAD in the image the kicad-cli exporters are dead too — that leaves only the file-based schematic subset. Better to remove it cleanly now, three days after it landed, than to let users discover the wall.

Verified locally: the diff is the exact mirror of #24 plus its loopback follow-up (−218, no strays), repo-wide grep finds no dangling docker references, and the workspace gate passes (223 tests — exactly the docker config test removed, no behavior change). Your call-out on why `56f2301` goes with it (compose-only, no Rust default changed) checks out against the diff.

Thanks for being willing to revert your own feature — that's good stewardship.

PR #49, opened 2026-07-18
fix(docs): mark the tool! macro usage example as ignore — doctest cannot compile

tascsystems-andrew:
`cargo test -p konnect-core` currently fails on main:

```
test crates/konnect-core/src/tools/mod.rs - tools::tool (line 216) ... FAILED
error: cannot find macro `tool` in this scope
```

The `tool!` usage example in the macro's doc comment is fenced as plain `rust`, so cargo tries to compile it as a doctest, where the macro isn't in scope. Fencing it as `rust,ignore` keeps it rendered in rustdoc while exempting it from compilation.

Found while rebasing the label-fix PRs after today's merges — thanks for those! With this, `cargo test -p konnect-core` is fully green on main again.

### DIFF
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -213,7 +213,7 @@ mod query_cache_tests {
 /// Shorthand for building a ToolDef with a typed async handler function.
 ///
 /// Usage:
-/// ```rust
+/// ```rust,ignore
 /// tool!(
 ///     "tool_name",
 ///     "Description of what it does.",


mixelpixx (approved):
Approving — reproduced the doctest failure on main and confirmed the branch fixes it (`cargo test -p konnect-core --doc`: FAILED → ok/1 ignored). Good catch; our CI and local gates run `--lib --tests`, which is exactly why this slipped through. Thanks for the quick turnaround on all the rebases today.

PR #54, opened 2026-07-22
fix(symbol-lib): rebuild KiCad 10 library parsing on PR #51 base

presire:
## Summary

Rebuild KiCad 10 library parsing on top of the structural S-expression implementation from PR #51,
while preserving the branch's PCB and IPC fixes.

## Changes

- Support CRLF and TAB-indented KiCad 10 library files.
- Resolve nested `type="Table"` library references.
- Preserve UTF-8 paths during KiCad URI expansion.
- Preserve unresolved environment-variable tokens safely.
- Parse symbols, pins, properties, and library tables structurally.
- Preserve PCB footprint placement improvements and IPC response validation.
- Add handler-to-handler regression coverage for nested library resolution.

## Testing

- `cargo fmt --all -- --check`
- `cargo test -p konnect-core`
- `cargo test -p konnect-ipc`

All tests pass.

PR #65, opened 2026-07-25
fix(library): read KiCad 10's real lib-tables

emolitor:
## The symptom

On a stock KiCad 10 install, every library listing came back empty:

```
list_footprint_libraries -> {"count": 0}
list_symbol_libraries    -> {"count": 0}
search_footprints        -> no results
```

Because nothing resolved, `place_component` could not turn any `Library:Footprint`
id into a file, so PCB placement was unusable on a default install.

## Root cause

Three separate things had to be wrong at once, which is why the earlier partial
fix did not help.

**1. The table was never parsed.** `parse_lib_table` scanned for the literal
needle `"\n  (lib "` — a newline plus exactly two spaces. KiCad writes
`fp-lib-table` and `sym-lib-table` TAB-indented with CRLF line endings, so the
needle never matched and every table read as empty.

This is the same class of bug already fixed for symbol libraries in
`list_symbols_parses_kicad10_crlf_tab_format`; `parse_lib_table` and the
duplicate scan inside `search_footprints` were missed. Both now use
`find_block_starts`, which `konnect-sexp` already provides precisely because
"KiCAD's own writers indent with tabs while this crate's writer uses two
spaces".

**2. KiCad 10 stopped inlining its bundled libraries.** The default global table
no longer lists the ~155 shipped libraries. It holds one indirection entry:

```
(lib (name "KiCad") (type "Table") (uri ".../template/fp-lib-table"))
```

Treating that as an ordinary library hid everything behind it. Nested tables are
now followed, with a depth cap so a self-referencing table terminates rather
than recursing forever.

**3. Shipped entries are addressed by environment variable.** The template table
names libraries as `${KICAD10_FOOTPRINT_DIR}/Resistor_SMD.pretty`. Callers
previously skipped any URI starting with `${` as unresolvable. Those are now
expanded — from the exported variable when KiCad set one, otherwise from the
known install locations. `find_kicad_symbol_dirs` generalises into
`find_kicad_library_dirs` so symbols, footprints and 3D models share one
discovery path.

## Also in this change

- Entries carry a resolved `path` alongside the raw `uri`, so callers stop
  re-deriving it.
- `resolve_footprint_path` centralises `Library:Footprint` lookup. On failure it
  reports how many libraries were seen and names a few, which turns a silent
  empty result into a diagnosable one.

## Compatibility

No public surface changes. Tool names, schema fields and config keys are
untouched; no migration needed.

Library listings gain a `path` field alongside the existing `uri`. That is
additive — existing consumers that read `uri` and `nickname` are unaffected.

Tables that already parsed under the old two-space scan still parse: the switch
to `find_block_starts` widens what is accepted rather than narrowing it, and
`parse_lib_table_still_reads_two_space_indentation` pins that.

## Validation

**Against a real install.** Stock KiCad 10.0, counts before -> after:

| | before | after |
|---|---|---|
| footprint libraries | 0 | 157 |
| symbol libraries | 0 | 225 |
| `search_footprints` | no results | finds bundled parts |

**Regression tests.** Parser-level tests pin the format handling, and two
handler-level tests reproduce what users actually hit:

- `list_footprint_libraries_reads_a_table_kicad_wrote` drives the real handler
  with a table in the exact shape KiCad writes (CRLF, tab-indented). It asserts
  up front that the fixture does *not* contain the old `"\n  (lib "` needle, so
  it genuinely fails against the previous implementation rather than passing by
  accident.
- `list_footprint_libraries_expands_a_nested_table_of_env_var_uris` covers the
  two constructs that kept the bundled libraries invisible even once the table
  parsed — a `(type "Table")` indirection and a `${KICAD10_FOOTPRINT_DIR}` URI —
  asserting both resolve to a real directory.
- `flatten_lib_table_stops_at_a_self_referencing_table` pins the depth cap.

`cargo test --workspace --lib --tests` passes; `cargo clippy --workspace -- -D
warnings` is clean; `cargo fmt --all` applied.

Not exercised here: the live-KiCad e2e suite, which needs a running KiCad and is
environment-dependent. The install-level numbers above were measured by hand
against KiCad 10.0 on Windows; other platforms' install roots are covered only
by `find_kicad_library_dirs`' existing path list.

## Risk and rollback

Read-only change — this code discovers and parses lib-tables and never writes
them, so no user file can be damaged by it and there is nothing to migrate back.
The blast radius is confined to which libraries are *visible*.

The realistic failure mode is over-inclusion: following nested tables could
surface a library a user expected to stay hidden, or an unusual `${VAR}` could
expand to an unintended directory. Both are visible in `list_*_libraries`
output. Recursion is bounded by `MAX_LIB_TABLE_DEPTH` and a self-reference test.

Rollback is a plain revert of the two commits; nothing persists outside the
process.

## Review feedback addressed

- **Test isolation.** The two tests that set `KICAD10_FOOTPRINT_DIR` now take a
  guard that serializes them on a mutex — the way `sch_components`'
  `SYMBOL_DIR_ENV` already does for `KICAD10_SYMBOL_DIR` — and restores the
  previous value on drop. Rust runs tests as threads of one process, so the two
  were a real race: whichever set the variable second decided what the other
  saw. The prior value is read with `var_os` rather than `var`, so a value this
  process cannot decode as UTF-8 is restored rather than deleted.
- **An unreadable lib-table read as zero libraries.** A regression this branch
  introduced: centralising the table read folded every `io::Error` into an empty
  list, where the handlers previously used `read_to_string(&table).await?` and
  surfaced it. `{"count": 0}` is the exact symptom this PR fixes, so an
  unreadable table became indistinguishable from the bug being back.
  `read_lib_table_checked` now separates absent (normal, empty) from unreadable
  (an error naming the file).
- **KiCad 8 install roots were missing.** Also this branch's doing: generalising
  `find_kicad_symbol_dirs` widened the environment lookup to majors 10/9/8 but
  left `kicad_share_roots` at 10.0 and 9.0, so major 8 was visible only where
  KiCad had exported its variable — precisely the machines this PR is about.
- **A missing footprint file did not name itself.** `resolve_footprint_path`
  documents that it returns a human-readable message on failure, but the
  filesystem-path branch returned `Ok` without checking anything, so
  `get_footprint_info` surfaced a typo as `The system cannot find the file
  specified. (os error 2)` — no filename, no hint that a `Library:Footprint` id
  was an option. It now checks (`is_file`, so a `.pretty` directory is rejected
  here rather than confusingly at read time) and names the file.
- **Doc accuracy.** `flatten_lib_table`'s comment promised a resolved `path`
  only "when the URI could be turned into an existing directory". Corrected:
  `expand_lib_uri` checks existence only for `${VAR}` expansions, a plain URI
  passes through as written, and either may name a file (`.kicad_sym`) rather
  than a directory.

One suggestion was **not** taken: moving `read_flat_lib_table` onto
`spawn_blocking`. Sync filesystem reads in async handlers are the established
convention here (94 such calls across `tools/*.rs`); `spawn_blocking` is
reserved for IPC sockets and `kicad-cli` invocations. Changing that is a
codebase-wide decision, not something this bug fix should do unilaterally.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -1970,6 +1970,95 @@ mod tests {
         format!("({kind}\r\n\t(version 7)\r\n{body})\r\n")
     }
 
+    #[tokio::test]
+    async fn list_footprint_libraries_reads_a_table_kicad_wrote() {
+        // End-to-end regression for the user-visible symptom: on a stock KiCad
+        // 10 install every library listing returned {"count": 0}, which left
+        // place_component unable to resolve any Library:Footprint id. Drive the
+        // real handler with a table in the exact shape KiCad writes.
+        let tmp = tempfile::tempdir().unwrap();
+        let pretty = tmp.path().join("MyParts.pretty");
+        std::fs::create_dir_all(&pretty).unwrap();
+        let table = kicad_style_table(
+            "fp_lib_table",
+            &[("MyParts", "KiCad", &pretty.to_string_lossy())],
+        );
+        assert!(
+            !table.contains("\n  (lib "),
+            "fixture must be in KiCad's tab format, not the old needle's"
+        );
+        std::fs::write(tmp.path().join("fp-lib-table"), table).unwrap();
+
+        let args = json!({
+            "project": tmp.path().join("board.kicad_pro").to_string_lossy(),
+            "scope": "project",
+        });
+        let res = handle_list_footprint_libraries(&args, &test_ctx())
+            .await
+            .unwrap();
+        assert!(!res.is_error, "handler errored: {:?}", res.content);
+
+        let out: serde_json::Value = serde_json::from_str(&result_text(&res)).unwrap();
+        assert_eq!(out["count"], 1, "library not found: {out}");
+        assert_eq!(out["libraries"][0]["nickname"], "MyParts");
+        assert_eq!(
+            out["libraries"][0]["path"].as_str().map(PathBuf::from),
+            Some(pretty),
+            "the resolved directory should be reported alongside the raw uri"
+        );
+    }
+
+    #[tokio::test]
+    async fn list_footprint_libraries_expands_a_nested_table_of_env_var_uris() {
+        // The two things that kept KiCad's ~155 bundled libraries invisible even
+        // once the table parsed: a `(type "Table")` indirection, and entries
+        // addressed as ${KICAD10_FOOTPRINT_DIR}/Foo.pretty.
+        let tmp = tempfile::tempdir().unwrap();
+        let shipped = tmp.path().join("share");
+        let pretty = shipped.join("Resistor_SMD.pretty");
+        std::fs::create_dir_all(&pretty).unwrap();
+        std::env::set_var("KICAD10_FOOTPRINT_DIR", &shipped);
+
+        let nested = tmp.path().join("template-fp-lib-table");
+        std::fs::write(
+            &nested,
+            kicad_style_table(
+                "fp_lib_table",
+                &[(
+                    "Resistor_SMD",
+                    "KiCad",
+                    "${KICAD10_FOOTPRINT_DIR}/Resistor_SMD.pretty",
+                )],
+            ),
+        )
+        .unwrap();
+        std::fs::write(
+            tmp.path().join("fp-lib-table"),
+            kicad_style_table(
+                "fp_lib_table",
+                &[("KiCad", "Table", &nested.to_string_lossy())],
+            ),
+        )
+        .unwrap();
+
+        let args = json!({
+            "project": tmp.path().join("board.kicad_pro").to_string_lossy(),
+            "scope": "project",
+        });
+        let res = handle_list_footprint_libraries(&args, &test_ctx())
+            .await
+            .unwrap();
+        let out: serde_json::Value = serde_json::from_str(&result_text(&res)).unwrap();
+
+        assert_eq!(out["count"], 1, "nested table not expanded: {out}");
+        assert_eq!(out["libraries"][0]["nickname"], "Resistor_SMD");
+        assert_eq!(
+            out["libraries"][0]["path"].as_str().map(PathBuf::from),
+            Some(pretty),
+            "env-var URI should resolve to a real directory"
+        );
+    }
+
     #[test]
     fn parse_lib_table_reads_kicad10_crlf_tab_format() {
         // Regression: parse_lib_table hard-coded the needle `\n  (lib ` (LF +
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -942,12 +942,35 @@ fn flatten_lib_table(content: &str, depth: usize) -> Vec<serde_json::Value> {
     out
 }
 
-/// Read a lib-table file from disk and flatten it. Returns an empty list when
-/// the file is absent or unreadable.
-fn read_flat_lib_table(path: &Path) -> Vec<serde_json::Value> {
+/// Read a lib-table file from disk and flatten it, reporting a table that is
+/// present but unreadable.
+///
+/// An absent table is normal and yields an empty list: a project without its
+/// own fp-lib-table simply has none, and every caller checks both the global
+/// and project tables. Anything else — a permissions problem, a truncated
+/// file — is not normal, and must not be folded into the same empty list. The
+/// symptom that produces is a bare `{"count": 0}`, which is precisely what the
+/// bug this module fixes looked like, so silence here would make a real
+/// failure indistinguishable from a regression.
+fn read_lib_table_checked(path: &Path) -> Result<Vec<serde_json::Value>, String> {
     match std::fs::read_to_string(path) {
-        Ok(content) => flatten_lib_table(&content, 0),
-        Err(_) => Vec::new(),
+        Ok(content) => Ok(flatten_lib_table(&content, 0)),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(Vec::new()),
+        Err(e) => Err(format!("Cannot read lib-table {}: {}", path.display(), e)),
+    }
+}
+
+/// As [`read_lib_table_checked`], for callers with nowhere to put an error.
+///
+/// The failure is logged rather than dropped in silence. Handlers that can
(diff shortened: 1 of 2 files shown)


emolitor:
Addressed the review feedback:

- **Test env-var race** — both tests that set `KICAD10_FOOTPRINT_DIR` now take a mutex guard (matching `sch_components`' existing `SYMBOL_DIR_ENV`) and restore the prior value on drop, read via `var_os` so a non-UTF-8 value is restored rather than deleted.
- **`flatten_lib_table` doc comment** — corrected; `expand_lib_uri` checks existence only for `${VAR}` expansions, and either form may resolve to a file rather than a directory.

Not taken: moving `read_flat_lib_table` onto `spawn_blocking`. Sync filesystem reads in async handlers are the established convention here (94 such calls across `tools/*.rs`); `spawn_blocking` is reserved for IPC sockets and `kicad-cli`. That is a workspace-wide decision rather than something this bug fix should change unilaterally — happy to open a separate issue if you would like it revisited.

Description rewritten to cover root cause, compatibility, validation, and rollback.

emolitor:
@mixelpixx this is ready for review.

emolitor:
Addressed the new finding:

- **`resolve_footprint_path` returning `Ok` for a missing path** — correct, and it contradicted the function's own documented contract. `get_footprint_info` surfaced a typo as `The system cannot find the file specified. (os error 2)` with no filename. It now checks `is_file` (so a `.pretty` directory is rejected here rather than confusingly at read time) and names the file plus the accepted forms. Three tests added.

emolitor:
Addressed the remaining doc-accuracy point (raised as low-confidence here, and inline on #66):

- **`resolve_symbol_lib_path`'s comment** claimed it returns the entry "whose URI resolves to an existing path". It does not — `expand_lib_uri` checks existence only for `${KICAD*_DIR}` expansions and takes a plain URI as written. This was the same false claim as the `flatten_lib_table` comment corrected earlier in this branch; I fixed one and left its sibling stating it. The wording now matches, and spells out the consequence: a stale global entry still shadows a working project entry of the same nickname.

I did **not** add the existence check the comment implied. That would change symbol-library resolution semantics — skipping stale entries so a shadowed project entry wins — which is a real improvement but a behaviour change beyond this PR's scope, and it deserves its own commit and test rather than arriving inside a review response. Happy to do it here if you would prefer, or as a follow-up.

emolitor:
Both new findings addressed — the first was a regression this branch introduced.

- **`read_flat_lib_table` swallowing read errors** — correct, and a bad trade specifically here. Centralising the table read folded every `io::Error` into an empty list; before, the two list handlers read with `tokio::fs::read_to_string(&table).await?` and surfaced it. `{"count": 0}` is the exact symptom this PR fixes, so an unreadable table became indistinguishable from the bug being back. `read_lib_table_checked` now separates the cases: an absent table stays an empty list (a project without its own fp-lib-table is normal, and the global table is checked too), anything else is an error naming the file. Both list handlers surface it again; `read_flat_lib_table` remains for callers with nowhere to put an error but logs rather than discarding. Three tests added, including the handler-level one.

- **`find_kicad_library_dirs` doc comment** — also correct. `KICAD<major>_<KIND>_DIR` substitutes to `KICAD10_SYMBOLS_DIR`; the real variable is singular, and the 3D one is not a word at all. The doc now lists the three actual names.

emolitor:
Both addressed.

- **Windows install roots missing KiCad 8** — correct, and a mismatch this branch introduced. `find_kicad_symbol_dirs` previously read one variable (`KICAD10_SYMBOL_DIR`); generalising it widened the env lookup to majors 10/9/8 but left `kicad_share_roots` listing only 10.0 and 9.0. A major covered by the env var but not the install roots is invisible on any machine where KiCad did not export it — which is every machine where Konnect was not launched by KiCad, the same gap this PR exists to close. KiCad 8's roots are now listed, with a note tying the two lists together.

- **`uri` loop variable** — renamed to `resolved`. The entries stopped being raw uris when `read_flat_lib_table` took over the expansion.

emolitor:
All three addressed.

- **`std::env::var` in `find_kicad_library_dirs` and `expand_lib_uri`** — correct. `var` reports a non-Unicode value as absent, so discovery would fall back to guessing install roots, and a perfectly good `${KICAD*_DIR}` would go down the same guess path — on exactly the machines where the variable was the only right answer. Both now use `var_os`, matching what this branch's test guard already did.

- **Windows drive-relative paths** — a genuine hole. `C:\libs\R.kicad_mod` was caught by the separator test, but `C:R.kicad_mod` (meaning `R.kicad_mod` in drive C's current directory) carries no separator and is otherwise shaped exactly like a lib id, so it was parsed as nickname `C`. `is_lib_id` now reads a one-letter prefix as a drive letter.

  Flagging the judgment call: nothing actually distinguishes `C:foo` the drive-relative path from `C:foo` the library `C`. A drive letter is much the likelier reading, and guessing the other way means silently hunting for a library named "C" — but the cost is that a single-letter nickname can no longer be written in this form. It stays reachable by path, and the rule applies on every platform so behaviour does not shift under the caller's feet. Happy to invert it if you would rather single-letter nicknames win.

emolitor:
Closing in lieu of https://github.com/mixelpixx/Konnect/pull/70 which is a better approach

mixelpixx:
Merged to main in eeb6b79 (your branch verbatim, plus one maintainer commit on top — the PR shows closed rather than merged only because maintainer-edit was off on the fork, so the merge went through main directly; every commit of yours is in the history with authorship intact).

**Verified live against a stock KiCad 10.0 Windows profile before merging**: `list_footprint_libraries` 0 → **155**, `list_symbol_libraries` 0 → **222**, `search_footprints` returns results, and the global table on this machine is byte-for-byte the tab-indented `(type "Table")` indirection your root-cause analysis describes. The three-things-wrong-at-once diagnosis was exactly right.

**Maintainer commit adds** (78860de): `${KIPRJMOD}` resolution against the table's own directory — project scope is `register_footprint_library`'s default, so project-registered libraries now resolve (verified live: register → list shows the resolved path); Flatpak/Snap//opt/kicad roots on Linux and the Homebrew prefix on macOS; and the sym-lib-table scaffold root fix (credit to PR #54's diagnosis — a fresh `sym-lib-table` now starts with `(sym_lib_table`, verified live).

Thanks for the eight rounds of review-hardening iterations on this — the unreadable-table-vs-empty distinction in particular is the kind of detail that prevents the next mystery report.

PR #66, opened 2026-07-25
fix (pcb): Place real footprints instead of reporting phantom success

### LINKED ISSUE #61: place_component returns success but never places the footprint (same bug class as #58)
### Summary

The `place_component` MCP tool returns a success-shaped response echoing the input args (footprint/reference/x/y/rotation/layer), but **no footprint is ever placed** on the board — nothing appears in the live KiCAD session, and `get_component_list` / `find_component` never see it afterward.

### Environment

- KiCAD 10.0.5, macOS
- Konnect v0.2.0, installed via KiCAD PCM
- IPC at `ipc:///tmp/kicad/api.sock`, PCB Editor open with the board loaded
- Observed 2026-07-23

### Reproduction

1. Open a board in KiCAD with the IPC API enabled, PCB Editor frame open.
2. Call `place_component` with a footprint already used elsewhere on that same board, e.g.:
   ```
   place_component(board="<path>.kicad_pcb", footprint="Resistor_SMD:R_0603_1608Metric", reference="TESTR", x=124.25, y=92.725, rotation=0)
   ```
3. Observe the tool returns success JSON, but the `placed` field is an **empty string** rather than the reference:
   ```json
   {"footprint":"Resistor_SMD:R_0603_1608Metric","layer":"F.Cu","placed":"","rotation":0.0,"x":124.25,"y":92.725}
   ```
4. Call `find_component(board=..., reference="TESTR")` or `get_component_list(board=...)`.

### Expected

`TESTR` exists on the board at (124.25, 92.725) and shows up in `get_component_list`.

### Actual

`find_component` errors with `Footprint 'TESTR' not found`, and `get_component_list` doesn't include it. Reproduced 3 times with different footprints/libraries to rule out library-table resolution as the cause:

- `Button_Switch_SMD:SW_SPST_EVQPE1` (library registered project-scope via `register_footprint_library` first) — failed
- `Resistor_SMD:R_0603_1608Metric` (already resolvable — 4 other footprints on the same board use this exact string successfully) — failed
- `Resistor_SMD_test:R_0603_1608Metric` (same directory, freshly registered global-scope) — failed

Same board, same session: `delete_component` on existing footprints (`RV1`, `RV2`) worked correctly and persisted in `get_component_list`. So the IPC session/board handle is healthy — this is specific to the placement path.

### Possibly related

This looks like the same bug class as #58 (`add_via` returns success but never creates the via): a "success" response fabricated from the input args rather than from the actual `CreateItemsResponse`, likely via the same `ParseAndCreateItemsFromString`-based path that silently no-ops rather than the protobuf `create_items` path known to work for `add_track`/`delete_component`. The empty `"placed": ""` field (rather than echoing the reference designator, which the handler clearly has) is consistent with a response built before/without checking whether the IPC call actually created anything.

### Affected versions

Konnect v0.2.0 against KiCAD 10.0.5.

emolitor:
Fixes #61.

> Stacked on #65 — that PR's lib-table fix is what makes `Library:Footprint`
> ids resolve at all, so this one needs it. Review the top two commits here.

## The symptom

`place_component` reported success on every call while the board stayed empty.
Two "successful" placements followed by a save produced a `.kicad_pcb`
containing zero footprints.

## Root cause

Two independent failures that combined to hide each other.

**The payload had no body.** The tool synthesised a three-line stub — a
`(footprint "Lib:Fp")` with a layer and a position and nothing else. KiCad's
parser needs the real pad and graphic definitions; given a body-less stub it
parses it to nothing and creates no item.

**The response was discarded.** `ParseAndCreateItemsFromString` returns a
`CreateItemsResponse` whose own documentation warns that the overall status
"may return IRS_OK even if no items were created". The per-item results are the
only trustworthy signal, and they were never read — so a payload KiCad rejected
outright still looked like success at the transport level.

This is the same failure the `(via …)` path had before it moved to a typed
message; see `add_via`.

## The fix

**Send the real definition.** A library `.kicad_mod` is already a complete
footprint sitting at the origin with a `REF**` placeholder. Placing it means
renaming it to the full library id, stamping in position, rotation and a fresh
UUID, and substituting the real reference designator — then letting KiCad's own
parser handle the pads and graphics. The whole definition is forwarded rather
than reconstructed.

**Believe the response.** `place_footprint` now fails when KiCad reports zero
created items, with an error that says the S-expression was rejected by the
parser rather than claiming success.

**Rotation reaches the children.** KiCad stores each pad's and text item's
*absolute* orientation while their positions stay in unrotated footprint-local
coordinates — a part placed at -90° keeps `(at -0.775 0 270)` on pad 1. Omitting
this leaves pad shapes unrotated relative to the body and trips KiCad's own
`lib_footprint_mismatch` check. Text is additionally kept readable, matching
KiCad's 180° flip for labels that would otherwise be upside down.

**IPC first, then the file.** KiCad 10.0 answers
`ParseAndCreateItemsFromString` with an empty response and creates nothing, so
in practice this always falls through to editing the board file. The IPC attempt
is kept so the tool starts working the moment KiCad implements the command. When
KiCad is reachable but did nothing, the result carries a warning that the board
was edited on disk and must be reopened, since an open editor would otherwise
overwrite it on next save.

## Compatibility

No tool names, schema fields or config keys change, so nothing to migrate.

Two behaviour changes are deliberate and user-visible:

- **`place_component` can now fail where it previously "succeeded".** That is
  the point of the fix — the old success was a false report over an empty board.
  Any caller that treated a success response as proof a footprint exists was
  already wrong; it now gets an actionable error instead.
- **Results carry a `source` field** (`"ipc"` or `"file"`) and, on the file
  path with KiCad reachable, a `warning`. Both are additive.

Board files are written in KiCad's own format — tab-indented, definition
forwarded verbatim from the `.kicad_mod` — so boards stay readable by KiCad 9
and 10 alike. No `.kicad_pcb` schema version is bumped.

## Validation

`place_component_puts_a_real_footprint_on_the_board` is the direct reproduction:
it asserts the board file gains a footprint *with pads and pad geometry*. Under
the old code the count is zero — exactly the empty board the bug produced.

Also covered: rotation reaching pads (270) and staying readable on text (90),
the error path leaving the board byte-for-byte untouched, the `ipc`/`file`
source field, and the array path writing every instance. Every test that writes
a board re-checks that it is still balanced.

The IPC response handling is covered against a mock NNG server (no KiCad
required), including the rejected-item, empty-response, all-created and mixed
cases.

Each new assertion was confirmed to fail against the pre-fix code, not merely
pass against the new code — the fixes were temporarily reverted to check that
the four new assertions go red, then restored.

The review-response commit was additionally put through an independent
adversarial review, which is what turned up the nickname-derivation hole above
(and the `.pretty` case sensitivity, the path-like declared name, and the
weak-balance-check gaps). Those findings are fixed here; the two it raised that
are pre-existing and workspace-wide — the shared `write_atomic` temp path and
comment-unaware paren counting — are recorded under Risk below rather than
quietly expanded into this PR.

`cargo test --workspace --lib --tests` passes; `cargo clippy --workspace -- -D
warnings` is clean; `cargo fmt --all` applied.

Not exercised here: the live-KiCad e2e suite, which needs a running KiCad. The
IPC branch specifically cannot be proven green until KiCad implements
`ParseAndCreateItemsFromString` — against KiCad 10.0 it returns an empty
response, which is why the file path is what actually runs today and what the
tests cover.

## Risk and rollback

This PR writes to `.kicad_pcb` files, so that is where the risk sits.

- Writes go through `write_atomic` (write temp, fsync, rename), so a single
  writer interrupted mid-run leaves the original board intact.
- `insert_into_board` re-parses the result and refuses to write anything that is
  not one complete `(kicad_pcb …)` form, so malformed input fails without
  touching the file.
- The error path is asserted to leave the board byte-for-byte identical.

Two limits worth stating plainly rather than leaving implied:

- **Concurrency.** `write_atomic` derives its temp file from the destination
  name, so two writers targeting one board can collide. That is pre-existing
  and shared by every write in the workspace, so fixing it belongs in
  `konnect-sexp` rather than here — but it does mean the atomicity guarantee
  above is a single-writer one.
- **Comments.** The paren counting treats every paren as syntax, including
  inside a `#` comment. KiCad does not write comments into `.kicad_pcb` and no
  reader in this workspace understands them, so the assumption is consistent —
  but it is an assumption, and it is called out in the doc comment.

The residual risk is a *semantically* wrong but well-formed board — a footprint
placed with the wrong rotation folded into its pads, say. That is why rotation
is asserted at the pad and text level rather than only on the footprint header.

**The one thing to know:** when KiCad is running with the board open, the file
edit is invisible to it and will be overwritten on its next save. The tool
detects that case and returns a warning saying to close without saving and
reopen. It cannot prevent it — only KiCad implementing the IPC command fixes
that properly.

Rollback is a plain revert of the top two commits; boards already written stay
valid, since they are ordinary KiCad footprints.

## Review feedback addressed

**A filesystem path could become the board's library id.** `resolve_footprint_path`
deliberately accepts either a `Library:Footprint` id or a bare path to a
`.kicad_mod`, but the header was written with whichever it was given. Placing by
path produced:

```
(footprint "C:\...\Resistor_SMD.pretty\R_0805_2012Metric.kicad_mod"
```

which is not a library identifier — KiCad opens the board with the part flagged
as a broken library link, and the id no longer round-trips through the IPC
lookups that expect `nickname:name`. This was not hypothetical: the end-to-end
test placed by path precisely to stay hermetic, so the regression test itself
was writing an absolute path into its fixture board.

`board_lib_id` is now total — every branch returns something that is not a path.
In preference order:

1. the caller's own `Library:Footprint` id, verbatim;
2. the nickname the **fp-lib-table** maps to the containing directory;
3. a conventional `<nickname>.pretty/` directory;
4. failing all of those, a bare footprint name.

Step 2 is the one that matters, and it is not what the first version of this fix
did. A nickname is *not* derivable from a path: KiCad lets a table map any
nickname to any directory, so `MyParts` may well live in `vendor.pretty`.
Guessing `vendor` from the path would silently mislink the part — worse than an
obviously broken link, because it looks right. Only the table can answer it, so
`footprint_lib_nickname_for_dir` asks it, comparing canonicalised paths so a
symlinked entry still matches. The `.pretty` convention remains as the fallback
for a library that is not registered at all, matched case-insensitively since
Windows and macOS filesystems are.

Step 4 keeps the name the library file declares, unless that is itself
path-like — a malformed file naming itself `/tmp/R.kicad_mod` must not smuggle a
path in through the fallback — in which case the file stem is used, which cannot
contain a separator.

The "id or path" rule moved to `library::is_lib_id` so the two call sites cannot
disagree about a Windows drive letter's colon. The board fixture now lays
footprints out in `Resistor_SMD.pretty/` the way KiCad does, so the main path is
what the tests exercise, and asserts no `.kicad_mod` appears anywhere in the
written board.

**The zero-created check counted the wrong thing.** This one undermined the fix
itself, and is the most important change in the PR. `place_footprint` failed
only when `created_items` was *empty* — but that field is documented as "status
of each item **to be** created", so a rejected item still occupies a slot,
carrying a code like `ISC_INVALID_DATA` and no `item`. A wholesale rejection
therefore arrives as a one-element vector, `len()` reports 1, and the call is
treated as success: precisely the phantom success this PR exists to remove,
reintroduced by the check meant to remove it.

Only `ISC_OK` entries now count (plus an entry with a defaulted status that
nonetheless carries an item, since some paths populate one and leave the other).
`parse_and_create_items` returns a `CreateOutcome` rather than a bare count, so
the per-item reasons survive into the error — "KiCAD created nothing" becomes
"KiCAD rejected it — IscInvalidData: `<KiCad's own message>`".

Four mock-server tests cover this, including the mixed case where one item is
created and one rejected; the rejection tests fail against the old `len()` count.

**`duplicate_component` had no file fallback.** It was the one caller still
invoking `place_footprint` bare. Now that `place_footprint` correctly fails on
KiCad 10.0, that tool would have gone from silently doing nothing to hard-failing
with no way through — the only tool left behind by the IPC-then-file strategy.
It now follows the same path as `place_component`. Reaching that point already
proves KiCad is running, since the source footprint is read over IPC, so the
file edit carries the same reopen warning.

**Inserting mixed line endings into a board.** `indent_block` rejoined with a
bare `\n`, so placing into a CRLF board left two conventions in the file. This
is the normal Windows path: checked against a KiCad 10.0 install, the bundled
footprints are CRLF throughout — `Reverb_BTDR-1H.kicad_mod` has 535 CRLF endings
and zero bare LF — as is the shipped `fp-lib-table`. So placing any stock
footprint on Windows mixed them, in a crate whose module doc opens by saying it
preserves KiCAD's exact formatting. `insert_into_board` now takes the
destination's convention and imposes it on what it inserts, in both directions.

**The balanced-parens check was claimed but not implemented.** It was only ever
asserted in tests; the write path itself trusted `rfind(')')` to find the
board's closing paren. Both file-edit writes now go through `insert_into_board`.

The check verifies the result is one *complete* `(kicad_pcb …)` form, not merely
that a balanced block exists somewhere — `find_balanced_block` skips whatever
precedes the first paren, so the weaker check would have accepted
`garbage(kicad_pcb …)`, and would equally have accepted a well-formed
expression that is not a board at all. A truncated board now fails loudly
instead of being saved in a state KiCad cannot open.

One suggestion was **not** taken: checking `any.type_url` in
`parse_and_create_items` before decoding. The `type_url` is already logged at
debug alongside the decoded outcome, and the zero-created path now reports
KiCad's own per-item rejection text, so the extra branch buys little.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -1002,6 +1002,31 @@ pub(crate) fn is_lib_id(reference: &str) -> bool {
     !(nick.len() == 1 && nick.as_bytes()[0].is_ascii_alphabetic())
 }
 
+/// The nickname the fp-lib-table gives to the library living in `dir`, if any.
+///
+/// This is the inverse of `resolve_footprint_path` and exists because a
+/// nickname is *not* derivable from the directory name: KiCad lets a table map
+/// any nickname to any path, so `MyParts` may well point at `vendor.pretty`,
+/// and two nicknames may share one directory. Only the table can answer it.
+///
+/// Paths are compared canonicalised so a symlinked or non-normalised entry
+/// still matches, falling back to a literal comparison when canonicalisation
+/// fails (a directory that no longer exists, say).
+pub(crate) fn footprint_lib_nickname_for_dir(dir: &Path) -> Option<String> {
+    let canonical = std::fs::canonicalize(dir).ok();
+    let same = |candidate: &Path| -> bool {
+        match (&canonical, std::fs::canonicalize(candidate).ok()) {
+            (Some(a), Some(b)) => a == &b,
+            _ => candidate == dir,
+        }
+    };
+
+    read_flat_lib_table(&global_fp_lib_table())
+        .into_iter()
+        .find(|lib| lib["path"].as_str().is_some_and(|p| same(Path::new(p))))
+        .and_then(|lib| lib["nickname"].as_str().map(str::to_string))
+}
+
 /// Resolve a footprint reference to an on-disk `.kicad_mod` path.
 ///
 /// Accepts either a direct filesystem path or KiCad's `Library:Footprint`
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -291,7 +291,8 @@ pub fn tools() -> Vec<ToolDef> {
             json!({
                 "type": "object",
                 "properties": {
-                    "footprint_path": { "type": "string", "description": "Path to .kicad_mod file, OR 'Library:Footprint' identifier" }
+                    "footprint_path": { "type": "string", "description": "Path to .kicad_mod file, OR 'Library:Footprint' identifier" },
+                    "project": { "type": "string", "description": "Path to .kicad_pro, so project-registered libraries resolve too (optional)" }
                 },
                 "required": ["footprint_path"]
             }),
@@ -1030,9 +1031,21 @@ pub(crate) fn footprint_lib_nickname_for_dir(dir: &Path) -> Option<String> {
 /// Resolve a footprint reference to an on-disk `.kicad_mod` path.
 ///
 /// Accepts either a direct filesystem path or KiCad's `Library:Footprint`
-/// form, which is looked up in the global fp-lib-table. Returns a
-/// human-readable message on failure so callers can surface it verbatim.
-pub(crate) fn resolve_footprint_path(reference: &str) -> Result<PathBuf, String> {
+/// form. Returns a human-readable message on failure so callers can surface it
+/// verbatim.
+///
+/// A lib id is looked up in `project_dir`'s fp-lib-table first, then the global
+/// one. Project-first matches KiCad, where a project entry shadows a global one
+/// of the same nickname, and it is the only order that makes
+/// `register_footprint_library` useful — it writes to the project table by
+/// default, so a global-only lookup cannot see anything it registers.
+///
+/// (`resolve_symbol_lib_path` still searches global-first for symbols; that
+/// asymmetry is pre-existing and noted on that function.)
+pub(crate) fn resolve_footprint_path(
+    reference: &str,
+    project_dir: Option<&Path>,
+) -> Result<PathBuf, String> {
     if !is_lib_id(reference) {
         // Check here rather than leaving it to the caller's read: an unchecked
         // path reaches the reader as a bare io::Error, which surfaces as
@@ -1050,15 +1063,25 @@ pub(crate) fn resolve_footprint_path(reference: &str) -> Result<PathBuf, String>
     }
 
     let (nick, fp_name) = reference.split_once(':').expect("checked above");
-    let table = global_fp_lib_table();
-    if !table.exists() {
+
+    let global = global_fp_lib_table();
+    let project = project_dir.map(|d| d.join("fp-lib-table"));
+    if !global.exists() && project.as_ref().is_none_or(|p| !p.exists()) {
         return Err(format!(
-            "Global fp-lib-table not found at {}",
-            table.display()
+            "No fp-lib-table found (looked for {}{})",
+            global.display(),
+            project
+                .map(|p| format!(" and {}", p.display()))
+                .unwrap_or_default()
         ));
     }
 
-    let libs = read_flat_lib_table(&table);
+    let mut libs = Vec::new();
+    if let Some(project) = &project {
+        libs.extend(read_flat_lib_table(project));
+    }
+    libs.extend(read_flat_lib_table(&global));
+
     let Some(lib) = libs.iter().find(|l| l["nickname"].as_str() == Some(nick)) else {
         let known: Vec<&str> = libs
             .iter()
@@ -1865,8 +1888,13 @@ async fn handle_get_footprint_info(
     let fp_path_str =
         require_str(args, "footprint_path").map_err(|e| anyhow::anyhow!("{:?}", e))?;
 
-    // Resolve "Library:Footprint" form via global fp-lib-table
-    let path = match resolve_footprint_path(fp_path_str) {
+    // Resolve "Library:Footprint" against the project's fp-lib-table as well as
+    // the global one, when the caller says which project they mean.
+    let project_dir = args["project"]
+        .as_str()
+        .map(PathBuf::from)
+        .and_then(|p| p.parent().map(Path::to_path_buf));
+    let path = match resolve_footprint_path(fp_path_str, project_dir.as_deref()) {
         Ok(p) => p,
         Err(msg) => return Ok(CallToolResult::error(msg)),
     };
@@ -2311,6 +2339,32 @@ mod tests {
         assert!(!is_lib_id("R_0402.kicad_mod"));
     }
 
+    #[tokio::test]
+    async fn a_project_registered_library_resolves() {
+        // register_footprint_library writes to the project fp-lib-table by
+        // default, so a global-only lookup could not see anything it
+        // registered — the default workflow resolved to "library not found".
+        let tmp = tempfile::tempdir().unwrap();
+        let pretty = tmp.path().join("MyProjLib.pretty");
+        std::fs::create_dir_all(&pretty).unwrap();
+        std::fs::write(pretty.join("Foo.kicad_mod"), "(footprint \"Foo\")").unwrap();
+        std::fs::write(
+            tmp.path().join("fp-lib-table"),
+            kicad_style_table(
+                "fp_lib_table",
+                &[("MyProjLib", "KiCad", &pretty.to_string_lossy())],
+            ),
+        )
+        .unwrap();
+
+        assert_eq!(
+            resolve_footprint_path("MyProjLib:Foo", Some(tmp.path())).unwrap(),
+            pretty.join("Foo.kicad_mod")
+        );
+        // Without the project dir it is invisible, which is the bug.
+        assert!(resolve_footprint_path("MyProjLib:Foo", None).is_err());
+    }
+
     #[test]
     fn a_windows_drive_relative_path_is_not_a_library_id() {
         // `C:R.kicad_mod` means R.kicad_mod in drive C's current directory. It
@@ -2374,7 +2428,7 @@ mod tests {
         // point of the test.
         let tmp = tempfile::tempdir().unwrap();
         let missing = tmp.path().join("nope.kicad_mod");
-        let err = resolve_footprint_path(&missing.to_string_lossy())
+        let err = resolve_footprint_path(&missing.to_string_lossy(), None)
             .expect_err("a nonexistent path must not resolve");
         assert!(err.contains("nope.kicad_mod"), "must name the file: {err}");
         assert!(
@@ -2388,7 +2442,7 @@ mod tests {
         // is_file, not exists — a .pretty directory would otherwise resolve and
         // fail confusingly at read time.
         let tmp = tempfile::tempdir().unwrap();
-        assert!(resolve_footprint_path(&tmp.path().to_string_lossy()).is_err());
+        assert!(resolve_footprint_path(&tmp.path().to_string_lossy(), None).is_err());
     }
 
     #[test]
@@ -2397,7 +2451,7 @@ mod tests {
         let file = tmp.path().join("R_0805.kicad_mod");
         std::fs::write(&file, "(footprint \"R_0805\")").unwrap();
         assert_eq!(
-            resolve_footprint_path(&file.to_string_lossy()).unwrap(),
+            resolve_footprint_path(&file.to_string_lossy(), None).unwrap(),
             file
         );
     }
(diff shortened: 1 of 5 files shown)


emolitor:
Addressed the review feedback:

- **Filesystem path reaching the board's library id** — real, and the end-to-end test was itself writing an absolute path into its fixture board. `board_lib_id` is now total. Worth noting the first fix for this was *wrong*: deriving the nickname from the `.pretty` directory name is a guess, and KiCad lets a table map any nickname to any directory, so `MyParts` may live in `vendor.pretty`. Guessing would silently mislink the part — worse than an obviously broken link. It now asks the fp-lib-table first (`footprint_lib_nickname_for_dir`, canonicalised path comparison) and only falls back to the `.pretty` convention for libraries that are not registered at all.
- **Balanced-parens check** — added for real; it was only ever asserted in tests. It verifies one complete `(kicad_pcb …)` form rather than merely "a balanced block exists", since `find_balanced_block` skips leading bytes and would have accepted `garbage(kicad_pcb …)` or a non-board expression.

Not taken: checking `any.type_url` in `parse_and_create_items`. The `type_url` is already logged at debug on the following line and the zero-created path gives a specific error, so the extra branch buys little.

Two related issues surfaced that are pre-existing and workspace-wide, so I have left them out of this PR rather than expanding its scope — both are now recorded under Risk in the description:

- `write_atomic` derives its temp file from the destination name (`konnect-sexp/src/writer.rs:97`), so concurrent writers targeting one board can collide. Affects every write in the workspace.
- Paren counting treats parens inside `#` comments as syntax. No reader in this workspace understands comments, so it is at least consistent, but it is an assumption.

Happy to open issues for either. Description rewritten to cover root cause, compatibility, validation, and rollback.

emolitor:
@mixelpixx this is ready for review.

emolitor:
Both new findings were correct, and the first one was serious — thank you.

- **`created_items.len()` counting attempts, not successes** — this undermined the fix itself. The field is documented as "status of each item **to be** created", so a rejected item still occupies a slot with a code like `ISC_INVALID_DATA` and no `item`. A wholesale rejection arrives as a one-element vector, `len()` reports 1, and the call reads as success: the exact phantom success this PR exists to remove, reintroduced by the check meant to remove it. Only `ISC_OK` entries now count (plus a defaulted status that still carries an item). `parse_and_create_items` returns a `CreateOutcome` so the per-item reasons reach the error — "KiCAD created nothing" becomes "KiCAD rejected it — IscInvalidData: `<its own message>`". Four mock-server tests cover it, including the mixed created/rejected case; the rejection tests fail against the old `len()` count.

- **`duplicate_component` with no file fallback** — also correct. It was the one caller still invoking `place_footprint` bare, so once `place_footprint` started failing honestly on KiCad 10.0 that tool went from silently doing nothing to hard-failing with no way through. It now uses the same IPC-then-file strategy as `place_component`; reaching that point already proves KiCad is running (the source footprint is read over IPC), so the file edit carries the same reopen warning.

Still not taken: the `any.type_url` check. The `type_url` is logged at debug alongside the decoded outcome, and the zero-created path now reports KiCad's own rejection text, so the extra branch buys little.

emolitor:
Both new points addressed — both were comment/documentation accuracy rather than behaviour:

- **`client.rs` created-item comment** — correct, and the comment was the wrong half rather than the code. Accepting an explicit `ISC_UNKNOWN` alongside a missing status is deliberate: protobuf cannot distinguish "unset" from "explicitly zero". The comment now states both, and why an item is required either way — without one there is nothing to have created, so it is a failure regardless of status. Added a mock-server test pinning both halves, since this was previously described but not exercised.

- **`resolve_symbol_lib_path` doc comment** — also correct, and fixed on #65 where the claim originates. It is the same false claim as the `flatten_lib_table` comment corrected earlier; I fixed one and left its sibling stating it. I did not add the existence check it implied, since that changes symbol-resolution semantics beyond this PR's scope — noted on #65 with an offer to do it there or as a follow-up.

emolitor:
Addressed, and it was a better catch than it looks — this is the normal Windows path, not an edge case.

`indent_block` rejoined with a bare `\n` and `insert_into_board` separated blocks the same way, so placing into a CRLF board left two conventions in the file. I checked against a real KiCad 10.0 install: the bundled footprints are CRLF throughout (`Reverb_BTDR-1H.kicad_mod` has 535 CRLF endings and zero bare LF), as is the shipped `fp-lib-table`. So placing any stock footprint on Windows mixed them — in a crate whose module doc opens by saying it preserves KiCAD's exact formatting.

`insert_into_board` now takes the destination's convention and imposes it on what it inserts, in both directions: a CRLF library footprint no longer drags `\r` into an LF board either.

Worth recording: the tests now normalise their fixtures explicitly. Git gives this source file CRLF in a Windows working tree, so a multi-line literal in it is CRLF locally and LF on CI — `EMPTY_BOARD` was silently already CRLF here, which is what made my first version of the LF test fail against correct code.

emolitor:
Addressed, and it is worse than the comment states — which makes it the best catch of this round.

`register_footprint_library` does not merely *support* `scope: "project"`, it **defaults** to it. So the default registration target was precisely the one place `place_component` could not see: register a library, then place from it, and the answer was `Library 'MyProjLib' not found`.

`resolve_footprint_path` now takes the project directory and searches the project table first, then the global one — project-first matching KiCad, where a project entry shadows a global one of the same nickname. The three placement handlers pass the board's own directory, which they already had; `get_footprint_info` gains an optional `project` argument, additive and matching what `list_footprint_libraries` already accepts.

`resolve_symbol_lib_path` still searches global-first for symbols. That asymmetry is pre-existing, and I have documented it on that function rather than changing symbol resolution semantics inside a PCB placement PR — happy to align it as a follow-up.

emolitor:
Closing in lieu of https://github.com/mixelpixx/Konnect/pull/70 which is a better approach

mixelpixx:
Following up now that #70 is merged (78ea80f): your work is in main with credit — the CreateOutcome/ISC_OK-only counting and its five mock tests (ported in the merge, commit message credits this PR), `board_lib_id`/`footprint_lib_nickname_for_dir`, the line-ending preservation, and the whole file-edit path (`board_footprint_sexp`/`insert_into_board`/`check_single_board_form`) which now serves as the offline fallback when IPC is unconfigured or unreachable — gated by a typed `TransportUnreachable` classification rather than the substring heuristic Copilot flagged. Between this, #65, and #72, a large share of today's placement fix is your code. Thanks for the graceful close-in-favor call too.

Issue #55, opened 2026-07-22
[Bug Report] S-expression output bugs in .kicad_sch files


jimmypluswang-ux:
# [Bug Report] S-expression output bugs in .kicad_sch files

**Konnect version:** v0.2.0 (also present in v0.1.3)
**KiCad version:** 10.0.4
**File format:** `(version 20250610)`

## Summary

Konnect writes invalid KiCad S-expressions in several places when creating/editing `.kicad_sch` schematic files. These cause KiCad GUI to fail loading schematics with parser errors.

---

## Bug 1: Wrong pin electrical type enum — `not_connected` should be `no_connect`

**Location:** `add_schematic_component` → pin electrical type

Konnect writes `(pin not_connected line ...)` but KiCad expects `(pin no_connect line ...)`.

**KiCad valid values:** `input`, `output`, `bidirectional`, `tri_state`, `passive`, `unspecified`, `power_in`, `power_out`, `open_collector`, `open_emitter`, `free`, `no_connect`

**Affects:** All components with NC (no-connect) pins.

---

## Bug 2: Dual electrical types in pin — `(pin output bidirectional line ...)`

**Location:** `add_schematic_component` → pin definition

Konnect writes `(pin output bidirectional line ...)` — two electrical types (`output` + `bidirectional`) where KiCad expects exactly one.

**Fix:** Should be `(pin bidirectional line ...)` (the more specific type).

**Affects:** Components with bidirectional pins in `.kicad_sch` files.

---

## Bug 3: Trailing `.00` in numeric coordinate values

**Location:** `add_schematic_component` → wire routing coordinates

Konnect writes coordinates like `(xy 117.46 108.00)` where `108.00` has trailing zeros.

The KiCad S-expression parser (GUI version) tokenizes `108.00` as two separate tokens (`108` and `.00`) instead of one token (`108.00`), causing `(xy x y)` to have the wrong arity. This manifests as an "expected ')'" parse error.

**Fix:** Remove trailing `.00` from all numeric values (or use integers directly).

**Affects:** Wire coordinates with `.00` trailing decimals.

---

## Bug 4: Multi-segment wires in a single `(wire (pts ...))` block

**Location:** `sch_wiring` tools

Konnect writes multi-segment wires as:

```lisp
(wire (pts (xy 80 115) (xy 80 108) (xy 117 108)) (stroke ...) (uuid "..."))
```

KiCad requires each segment to be its own wire with exactly 2 XY points:

```lisp
[...]

PR #69, opened 2026-07-25
fix(schematic): make wire deletion indentation-safe

### LINKED ISSUE #64: delete_schematic_wire / batch_delete_schematic_wire corrupt files with tab-indented schematics (hardcoded 2-space search)
## Summary

`handle_delete_wire` and `handle_batch_delete_wire` in `crates/konnect-core/src/tools/sch_wiring.rs`
locate the start of a `(wire ...)` block using a hardcoded 2-space-indent search:

https://github.com/mixelpixx/Konnect/blob/main/crates/konnect-core/src/tools/sch_wiring.rs#L547
https://github.com/mixelpixx/Konnect/blob/main/crates/konnect-core/src/tools/sch_wiring.rs#L580

```rust
let wire_start = before.rfind("\n  (wire").map(|p| p + 1).unwrap_or(0);
```

If the target `.kicad_sch` file uses **tab indentation** instead of two spaces (this is what
KiCad 10's own `eeschema` writes by default — see the file header: `generator_version "10.0"`),
`rfind` never matches, `.unwrap_or(0)` silently falls back to byte offset 0, and
`find_block_with_leading_whitespace` then walks forward from the very start of the file. The
resulting "block" spans from the top of the file to wherever the next balanced close-paren
lands — in practice, most or all of the file — and gets deleted. The write is atomic
(`write_atomic`), so the corruption isn't a partial/torn write; it's a clean, complete
overwrite with the wrong (near-empty) content.

## Impact

Silent, total data loss on a live schematic with no warning — the tool call returns
`"Wire deleted."` (success) while having just wiped the file.

## Repro

1. Open/create a `.kicad_sch` file saved by KiCad 10's `eeschema` (tab-indented — confirm with
   `cat -A file.kicad_sch | head` showing `^I(...)` rather than two spaces).
2. Call `delete_schematic_wire` with a valid `uuid` for any wire in the file.
3. First call may appear to succeed or fail with `"Wire not found"` depending on file layout.
4. A second call (or `batch_delete_schematic_wire` with 2+ UUIDs) reliably truncates the file to
   near-empty (`(kicad_sch\n\n` or similar).

Observed firsthand while merging a generated sub-circuit block into a hand-authored KiCad 10
LDO schematic (tab-indented) — two sequential `delete_schematic_wire` calls corrupted the file
after the first one succeeded normally.

## Suggested fix

Replace the hardcoded whitespace search with one that matches either tabs or spaces, e.g.:

```rust
let wire_start = before
    .rmatch_indices('\n')
    .map(|(i, _)| i + 1)
    .find(|&i| before[i..].trim_start().starts_with("(wire"))
    .unwrap_or(wire_offset); // or return an error instead of silently defaulting to 0
```

More importantly: **never silently fall back to offset `0`** on a failed indent-search — that's
what turns a "can't find this block" bug into a "delete everything" bug. Returning an explicit
error (`"Cannot locate wire block start"`) instead of `.unwrap_or(0)` would have made this a
no-op instead of data loss, even before fixing the underlying indent assumption.

I'd guess other handlers in this file (and possibly other `tools/*.rs` files) that search for
`"\n  ("` patterns have the same latent bug — happy to grep for and audit the rest if useful.

## Environment
- Konnect: built from `main` @ commit a060b59 (2026-07-25)
- KiCad: 10.0
- OS: Windows 10

perara:
### Summary

- replace fixed-indentation wire matching with string-aware balanced
  S-expression range discovery
- delete wires by UUID or by complete endpoints, including reversed endpoint
  order, without consuming adjacent forms
- reuse the same safe deletion path for wire splitting and batch operations
- reject malformed or structurally ambiguous edits instead of partially
  rewriting the schematic
- add regression tests for tabs, alternate indentation, escaped strings,
  reversed endpoints, duplicate requests, and protected parent structures

Closes #64. Related to #55.

### Root cause

The previous deletion code searched backward for a literal newline plus two
spaces before `(wire ...)`. Tabs or a different indentation width could move the
start offset into a neighboring form, corrupting the schematic while the tool
reported success.

### Implementation

`konnect-sexp` now exposes balanced, string-aware helpers for locating a form's
parent and direct-child ranges. Schematic wiring and batch tools use those
ranges, deduplicate targets, and fail closed when the expected structure cannot
be proven. Endpoint deletion compares the whole coordinate pair in either
direction rather than matching a partial textual fragment.

### Compatibility

Existing MCP tool names and request/response schemas are unchanged. Correctly
formatted legacy schematics remain supported; the change broadens safe support
to tabs and arbitrary indentation.

### Validation

- `git diff --check`
- `cargo test --locked -p konnect-core -p konnect-sexp`
  - `konnect-core`: 152 unit tests plus integration and conformance tests
  - `konnect-sexp`: 36 unit tests plus property tests
- targeted strict library Clippy passed for the affected crates

The upstream baseline's all-target Clippy currently reports unrelated existing
warnings (`fp_y` is unused and two tests hold a lock across an await), so those
baseline findings are not bundled into this focused fix.

### Risk and rollback

Risk is concentrated in schematic text-range selection. Coverage includes the
reported tab-indentation reproducer and adversarial nested/string cases. The
commit is self-contained and can be reverted without data migration.

### DIFF
--- a/crates/konnect-core/src/tools/sch_batch.rs
+++ b/crates/konnect-core/src/tools/sch_batch.rs
@@ -449,16 +449,22 @@ fn sexp_tag(block: &str) -> &str {
     &after_open[..end]
 }
 
+// Blocklist of structural forms, not an allowlist of item kinds: deleting a
+// drawing item (text, bus, sheet, image, polyline, …) by UUID has always
+// worked and must keep working — only the schematic's skeleton is protected.
 fn is_deletable_schematic_item(block: &str) -> bool {
-    matches!(
+    !matches!(
         sexp_tag(block),
-        "wire"
-            | "label"
-            | "global_label"
-            | "hierarchical_label"
-            | "junction"
-            | "no_connect"
-            | "symbol"
+        "version"
+            | "generator"
+            | "generator_version"
+            | "uuid"
+            | "paper"
+            | "title_block"
+            | "lib_symbols"
+            | "sheet_instances"
+            | "symbol_instances"
+            | "embedded_fonts"
     )
 }
 
--- a/crates/konnect-core/src/tools/sch_batch.rs
+++ b/crates/konnect-core/src/tools/sch_batch.rs
@@ -16,9 +16,13 @@ use konnect_sexp::{
         extract_labels, extract_lib_pins, extract_symbol_instances, extract_wires,
         format_net_label, format_wire, pin_endpoint, read_schematic,
     },
-    writer::{apply_edits, find_block_with_leading_whitespace, new_uuid, write_atomic, SexpEdit},
+    writer::{
+        apply_edits, find_block_with_leading_whitespace, find_enclosing_direct_child_block,
+        new_uuid, write_atomic, SexpEdit,
+    },
 };
 use serde_json::json;
+use std::collections::HashSet;
 
 // Re-use the crate-internal net-graph primitives from sch_analysis.
 use super::sch_analysis::build_net_graph;
@@ -363,6 +367,7 @@ async fn handle_batch_delete(
     let mut edits: Vec<SexpEdit> = Vec::new();
     let mut deleted: Vec<String> = Vec::new();
     let mut errors: Vec<String> = Vec::new();
+    let mut delete_ranges: HashSet<(usize, usize)> = HashSet::new();
 
     // Delete by UUID — walk back from uuid node to enclosing top-level block
     if let Some(uuids) = args["uuids"].as_array() {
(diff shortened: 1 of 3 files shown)


mixelpixx (approved):
Approving — this was the most urgent item in the queue, and the fix is exactly right.

**Reproduced the disaster on main first**: `delete_schematic_wire` by UUID on a tab-indented (eeschema-saved) schematic reported "Wire deleted." while truncating the file from 472 bytes to a single `\n` — the `rfind("\n  (wire").unwrap_or(0)` delete-from-top path. With this branch, the same fixture: targeted wire removed cleanly, neighbor wire and `sheet_instances` intact, a missing UUID fails closed with the file untouched, and deleting a structural node by UUID is rejected with a clear error naming the protected form.

The design is right too: reusing `find_enclosing_block` (the #40 machinery) rather than re-deriving indentation scanning, and putting `find_direct_child_blocks`/`find_enclosing_direct_child_block` in `konnect-sexp` where every caller inherits them — those are the correct general primitives for "top-level schematic item", and the depth/string state machine reads correct. The coordinate-matching rewrite (full endpoint pair, both directions, versus the old KiCad-8-era `(start x y)` tag that never matched anything) and the fail-closed batch semantics with per-UUID errors are exactly what #64 asked for.

**Two maintainer edits pushed to the branch** (thanks for maintainer-edit access): inverted `is_deletable_schematic_item` from an allowlist to a structural-forms blocklist — the allowlist would have silently broken previously-working deletions of text/bus/sheet/image items — and documented the all-four-coordinates contract in the tool schema, since agents read that text. All 295 workspace tests green on the branch merged with current main.

Also noting for the record: the same `rfind`/hardcoded-indent class survives at `sch_export.rs:396`, `integration.rs:631`, and `schematic_builder.rs:96-190` (non-delete paths, lower severity) — I'll file a follow-up issue rather than grow this PR.

Issue #23, opened 2026-07-13
`move_component` tool only moves footprint anchor, detaching pads


TUGTEN:
**Description:**
When using the `move_component` MCP tool to change the position of a footprint via IPC, only the footprint's anchor/origin `(at X Y)` is moved. The physical pads, silkscreen, and properties remain locked at their original absolute coordinates. This causes the footprint to "break apart," with the anchor moving to the new coordinates while the physical copper and silkscreen stay behind. 

Furthermore, running "Update PCB from Schematic" (F8) in KiCAD after this occurs does *not* snap the pads back to the origin, leaving the footprint permanently corrupted in the `.kicad_pcb` file.

**Steps to Reproduce:**
1. Open a blank KiCAD PCB project.
2. Place any standard footprint (e.g., `Resistor_SMD:R_0805_2012Metric`) at coordinate `X: 100, Y: 100` and assign it reference `R1`.
3. Start the `konnect` MCP server and ensure KiCAD IPC is active.
4. Send the following tool call to move the component:
   ```json
   {
     "ToolName": "move_component",
     "Arguments": {
       "reference": "R1",
       "x": 50,
       "y": 50
     }
   }
   ```
5. Trigger a board save or `save_project`.
6. Look at the KiCAD UI or inspect the `.kicad_pcb` file.

**Expected Behavior:**
The entire physical footprint (anchor, pads, silkscreen, and text) should move to `50, 50`.

**Actual Behavior:**
The footprint anchor moves to `50, 50` (indicated by the purple crosshair in KiCAD). The pads and silkscreen remain physically located at `100, 100`. 
If you inspect the raw `.kicad_pcb` file, the footprint's origin updates to `(at 50 50)`, but the internal pad lines and shapes get assigned absolute offsets (e.g., `start 50 50`) to keep them at their original absolute location on the board.

<img width="399" height="272" alt="Image" src="https://github.com/user-attachments/assets/3bb295b7-0d94-4bca-b077-6f6474c0f72c" />

**Environment:**
* macOS: 26.1 (25B78)
* KiCAD version: 10.0.4, release build
* Konnect MCP version: v0.1.3 (built from source)

mixelpixx:
Fixed in 37adcaa (on `main`).

[...]

PR #70, opened 2026-07-25
fix(ipc): place footprints through typed KiCad messages

### LINKED ISSUE #61: place_component returns success but never places the footprint (same bug class as #58)

perara:
### Summary

- resolve `Library:Footprint` IDs from project/global tables and standard KiCad
  library locations across Linux, macOS, and Windows
- translate real `.kicad_mod` geometry and pads into typed KiCad IPC messages
  instead of reporting success without a placed footprint
- validate envelope, operation, item-count, and per-item statuses and preserve
  the complete IPC error chain
- verify that a path-bearing request targets the active board before mutation
- wrap multi-step edit, align, and array operations in KiCad transactions with
  rollback on error or panic
- add mock-server, transformation, and real KiCad 10 GUI coverage through both
  the low-level client and the MCP JSON-RPC boundary

Closes #61.

### Root cause

`place_component` did not construct and verify a complete typed footprint from
the selected library definition. A nominal IPC response could therefore become
a successful MCP result even though the PCB editor had no corresponding placed
item.

### Implementation

The tool resolves the selected footprint source, validates its root metadata,
extracts typed fields and pads, creates all requested footprints in one typed
request, commits the transaction, and then verifies the resulting references.
Batch placement avoids querying transaction-staged items, which KiCad does not
expose before commit. Duplicate references are rejected during preflight.

The client now fails closed on missing/malformed statuses, mismatched item
counts, unexpected deletion IDs, inactive-board requests, and failed
transaction commit. Direct file fallback remains available only when IPC is not
configured; a configured IPC mutation failure is never masked by rewriting the
file behind KiCad's back.

### Compatibility

MCP tool names and JSON schemas are unchanged. Calls that previously returned a
false success now either produce a verified placement or an explicit error.
Custom-pad footprints are currently rejected with an actionable message rather
than approximated incorrectly.

### Relationship to open placement work

PRs #54 and #66 also address footprint placement. This draft is offered as a
separate, safety-focused implementation for maintainers to compare:

- it creates typed `FootprintInstance` data and validates the resulting KiCad
  response instead of falling back to editing an actively open board file;
- configured IPC failures remain failures, preventing a hidden file edit that
  KiCad could later overwrite;
- mutations verify the requested active-board path and multi-step operations use
  rollback-capable KiCad transactions;
- the exact branch was exercised through the real KiCad 10 GUI IPC API, not only
  through file-mode or mocked tests.

If maintainers prefer another placement implementation, the response validation,
transaction, active-board, and live-test coverage can be reviewed independently.

### Validation

- `git diff --check`
- `cargo check --locked -p konnect-ipc -p konnect-core -p konnect --all-targets`
- `cargo test --locked -p konnect-ipc -p konnect-core -p konnect --all-targets`
  - IPC unit/integration suites, footprint transforms, and mock transport passed
  - core and MCP suites passed; live tests remain ignored in ordinary runs
- `cargo clippy --locked -p konnect-ipc -p konnect-core -p konnect --lib --bins -- -D warnings`
- real KiCad 10.0.4 under Xvfb:
  - move and rotate preserved child-relative geometry
  - a standard-library resistor footprint was placed with real pads
  - edit, two-item array placement, alignment, save, and on-disk verification passed
  - mutation of a non-active board path was rejected

### Risk and rollback

This is a substantial boundary change because typed protobuf geometry must agree
with KiCad. The tests cover malformed responses, transaction failures, geometry
preservation, and the real GUI/API. Unsupported custom pads fail closed. The
single commit can be reverted without file-format migration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -916,6 +916,7 @@ dependencies = [
  "anyhow",
  "axum",
  "base64",
+ "dirs",
  "konnect-ipc",
  "konnect-schematic-editor",
  "konnect-sexp",
@@ -937,6 +938,7 @@ name = "konnect-ipc"
 version = "0.2.0"
 dependencies = [
  "anyhow",
+ "konnect-sexp",
  "nng",
  "prost",
  "prost-build",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -919,6 +919,8 @@ dependencies = [
  "konnect-ipc",
  "konnect-schematic-editor",
  "konnect-sexp",
+ "nng",
+ "prost",
  "prost-types",
  "reqwest",
  "rusqlite",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -916,7 +916,6 @@ dependencies = [
  "anyhow",
  "axum",
  "base64",
- "dirs",
  "konnect-ipc",
  "konnect-schematic-editor",
  "konnect-sexp",
--- a/crates/konnect-core/Cargo.toml
+++ b/crates/konnect-core/Cargo.toml
@@ -19,7 +19,6 @@ axum.workspace = true
 anyhow.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
-dirs.workspace = true
 uuid.workspace = true
 base64.workspace = true
 reqwest.workspace = true
--- a/crates/konnect-core/Cargo.toml
+++ b/crates/konnect-core/Cargo.toml
@@ -18,11 +18,10 @@ axum.workspace = true
 anyhow.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
+dirs.workspace = true
 uuid.workspace = true
 base64.workspace = true
 reqwest.workspace = true
 rusqlite.workspace = true
 usvg.workspace = true
-
-[dev-dependencies]
 tempfile = "3"
--- a/crates/konnect-core/Cargo.toml
+++ b/crates/konnect-core/Cargo.toml
@@ -25,4 +25,6 @@ base64.workspace = true
 reqwest.workspace = true
 rusqlite.workspace = true
 usvg.workspace = true
+
+[dev-dependencies]
 tempfile = "3"
--- a/crates/konnect-core/Cargo.toml
+++ b/crates/konnect-core/Cargo.toml
@@ -27,3 +27,7 @@ usvg.workspace = true
 
 [dev-dependencies]
 tempfile = "3"
+# The pcb_components fallback tests spawn a mock KiCAD NNG endpoint to prove
+# a reachable-but-rejecting KiCAD never triggers the file-editing fallback.
+nng.workspace = true
+prost.workspace = true
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -1132,7 +1132,7 @@ pub(crate) fn resolve_footprint_path(
     let attempted_list = if attempted.is_empty() {
         "no KiCad library directories were found — set KICAD10_FOOTPRINT_DIR for a \
          non-standard install"
-        .to_string()
+            .to_string()
     } else {
         format!(
             "also looked for {}",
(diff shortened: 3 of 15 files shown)


mixelpixx (approved):
Approving after a maintainer synthesis round on your branch (thanks for enabling maintainer edits) plus full live verification. Your typed-IPC transport was the right call — it's the only approach of the three parallel efforts proven to place a real footprint in a live KiCAD 10 editor, and it follows the create_items precedent #59 set. The transaction wrapper (`run_commit` with rollback + catch_unwind) and per-item response validation are the strongest parts of the PR and are now load-bearing for everything placement does.

**What the synthesis added on top** (11 commits; details in each message):
- Full-definition footprints: courtyard, silk, fab graphics and visible texts now ship through typed IPC (pre-transformed to absolute coords per issue #23's findings) — placed parts pass courtyard DRC and `lib_footprint_mismatch` instead of arriving pads-only.
- #66's CreateOutcome/ISC_OK-only counting semantics + its five mock-server tests, ported with credit (they pin the phantom-success trap the `created_items.len()` check reintroduces).
- Hybrid fallback per the repo decision: typed classification (`TransportUnreachable` marker, no substring matching) — file editing only when IPC is unconfigured/unreachable, never behind a live editor; rejections state the file was untouched.
- Library discovery deduplicated into `library::resolve_footprint_path` (project-table-first, from #65), pad math routed through the canonical `transform_pad` (#57), B.Cu string-swap replaced with an honest rejection, plus the URL/tempfile/2d-view scope-creep reverts.
- Two fixes found only by live verification: placement now targets the *named* board among all open documents (with the user's own project focused, first-document targeting fails or hits the wrong board — a two-board mock now pins this), and Reference/Value fields use the library footprint's own text anchors (the synthesized ±1 mm offset sat on the part's silkscreen — silk_overlap warnings on every placement).

**Live pass bar, all green** on KiCAD 10.0.4 with a second project simultaneously open: editor-visible placement over IPC on the correct board, `transform_pad`-exact pad readback, duplicate/nonexistent/B.Cu failure paths leaving the board byte-identical, courtyard+silk in the saved file, DRC clean for post-fix parts, and the offline fallback producing balanced, warned, convention-preserving file edits.

375 workspace tests + doctests green, clippy/fmt clean.

PR #72, opened 2026-07-26
fix(sexp): give every atomic write its own scratch file

### LINKED ISSUE #71: write_atomic shares one temp file across every file in a project, defeating its own atomicity
### Summary

`write_atomic` derives its temporary file from the destination with
`path.with_extension("kicad_tmp")`. Because `with_extension` **replaces** the
extension rather than appending, every file in a KiCad project collapses to the
same temporary path:

```
proj/board.kicad_pcb  ->  proj/board.kicad_tmp
proj/board.kicad_sch  ->  proj/board.kicad_tmp
proj/board.kicad_pro  ->  proj/board.kicad_tmp
```

That is the standard project layout, not an unusual one. Any two writes to a
project that overlap in time — a schematic and a board, or two board edits — use
one shared scratch file, which defeats the atomicity the function exists to
provide.

`konnect-sexp/src/writer.rs:96`:

```rust
pub fn write_atomic(path: &Path, content: &str) -> Result<(), SexpError> {
    let tmp_path = path.with_extension("kicad_tmp");

    {
        let mut f = std::fs::File::create(&tmp_path)?;
        f.write_all(content.as_bytes())?;
        f.flush()?;
        f.sync_all()?; // fsync — mandatory
    }

    std::fs::rename(&tmp_path, path)?;
    Ok(())
}
```

There are 61 call sites across 12 modules, so this underpins essentially every
file write in the workspace.

### Impact

Two writers sharing `board.kicad_tmp`:

1. Writer A creates and truncates the temp file and begins writing.
2. Writer B creates and truncates *the same* file and begins writing.
3. A renames the temp onto `board.kicad_pcb`.
4. B renames what is left onto `board.kicad_sch`.

The destinations can end up holding interleaved content from both writers, a
truncated prefix, or one file's content written to the other's path. Since
`File::create` truncates, B's truncate can also land between A's write and A's
rename, so A can rename a file that has been emptied underneath it.

The window is small but the payload is a user's schematic or board, and the
failure is silent — the write reports success.

### Reproduction

Concurrency is inherently timing-dependent, but the shared-path precondition is
deterministic and can be shown without threads:

```rust
use std::path::Path;
assert_eq!(
    Path::new("proj/board.kicad_pcb").with_extension("kicad_tmp"),
    Path::new("proj/board.kicad_sch").with_extension("kicad_tmp"),
);
```

Under load, spawn N threads calling `write_atomic` on the same path with
distinct contents and assert the result always equals one input exactly; it
does not reliably hold today.

### Secondary problem

On any failure after the temp file is created — a failed rename, a full disk —
the temp file is left behind. The user finds a stray `board.kicad_tmp` in their
project directory with no indication of what it is.

### Suggested fix

Give each write a temp path unique to the destination *file name* (not just its
stem) and to the writer:

- derive from `file_name()` rather than `with_extension`, so `.kicad_pcb` and
  `.kicad_sch` differ;
- add the process id and a per-process atomic counter, so concurrent writers in
  one process and separate processes differ;
- keep the temp file in the destination's directory, so the rename stays on one
  filesystem and remains atomic;
- remove the temp file if anything fails.

### Out of scope

Full crash durability would additionally require fsyncing the destination
*directory* after the rename. That is a separate concern from this bug: the
rename's visibility to subsequent readers — the property the doc comment
promises — does not depend on it.

### Environment

- Konnect `main` @ a060b59
- Affects all platforms

emolitor:
Fixes #71.

## The problem

`write_atomic` derived its scratch file with
`path.with_extension("kicad_tmp")`. `with_extension` **replaces** the extension
rather than appending, so every file in a KiCad project collapsed to the same
scratch path:

```
proj/board.kicad_pcb  ->  proj/board.kicad_tmp
proj/board.kicad_sch  ->  proj/board.kicad_tmp
proj/board.kicad_pro  ->  proj/board.kicad_tmp
```

That is the standard project layout, so this is not an exotic race. Any two
writes to a project that overlap in time shared one scratch file — which defeats
precisely the atomicity the function exists to provide.

`File::create` truncates, so the interleavings include:

- writer B truncates the scratch file between writer A's write and A's rename,
  so A publishes an empty or partial file;
- A and B interleave their writes, and whichever renames first publishes a
  mixture of both;
- B renames content intended for `board.kicad_sch` onto `board.kicad_pcb`.

In every case the write reports success. With 61 call sites across 12 modules,
this underpins essentially every file write in the workspace, and the payload is
a user's schematic or board.

## The fix

The scratch name is now built from the full **file name** plus the process id
and a per-process atomic counter:

```
board.kicad_pcb.31428.7.kicad_tmp
```

Each component earns its place — the file name separates a project's files, the
pid separates concurrent Konnect processes, and the counter separates concurrent
writes within one process.

It remains a **sibling** of the destination. A rename is only atomic within a
single filesystem, so moving scratch files to a temp directory would silently
degrade the guarantee to a copy on any setup where the project lives on a
different mount.

Separately, a scratch file is no longer left behind when a write fails part-way.
It was previously abandoned in the user's project directory with no indication
of what it was.

## Review feedback

The scratch name is built as an `OsString` rather than through
`to_string_lossy`. Worth stating precisely, since the review that prompted it
gave a reason that does not hold: lossy conversion **cannot** reintroduce
collisions here, because uniqueness comes from the pid and counter rather than
the name — two calls for the same path already get distinct scratch files.
`identical_names_in_one_directory_still_get_distinct_scratch_paths` pins that
rather than leaving it to argument.

The change earns its place for a smaller reason. The name exists so a leftover
from a killed process can be traced back to the file it belonged to, and lossy
conversion mangles exactly the names where that matters. Covered on both
platforms, since the invalid cases differ — a `0xFF` byte on Unix, an unpaired
high surrogate on Windows.

## Compatibility

None. `write_atomic`'s signature and behaviour on success are unchanged, and the
scratch file is an implementation detail that never outlives the call. No tool
names, schema fields or config keys are touched.

The scratch file keeps the `.kicad_tmp` extension, so any existing cleanup or
ignore rule keyed on it still matches.

## Validation

The precondition is stated without threads, since concurrency alone makes a poor
regression test — `a_projects_files_do_not_share_a_scratch_path` asserts the
three project files get distinct scratch paths, *and* asserts that the old
derivation collapses them, so the test records what it is guarding against.

Then it is driven concurrently:

- `concurrent_writers_never_interleave` — eight threads write 400 KB of a
  distinct repeated byte to one path, ten times each. The result must be exactly
  one writer's complete content, checked by asserting the file is uniformly one
  byte. Any interleaving shows up as a mixed file.
- `a_schematic_and_a_board_do_not_overwrite_each_other` — the project-layout
  case from the issue, writing `board.kicad_pcb` and `board.kicad_sch`
  concurrently and asserting neither lands on the other.
- `a_failed_write_leaves_no_scratch_file` — renames onto an existing directory,
  which fails on every platform, and asserts nothing is left behind.

All four fail against the old derivation; the concurrent ones fail with real
`write_atomic` errors as the writers stomp each other's scratch file, not merely
mismatched content. These tests cannot flake with the fix in place: each writer
owns its scratch file, so each rename publishes bytes no other writer touched.

`cargo test --workspace --lib --tests` passes (20 suites);
`cargo clippy --workspace -- -D warnings` is clean; `cargo fmt --all` applied.

Not exercised: the live-KiCad e2e suite, which needs a running KiCad. Verified on
Windows only — the ubuntu and macos legs of the matrix are down to CI.

## Risk and rollback

The change is confined to how one path string is built, plus cleanup on a
failure path that previously did nothing.

The residual risk is scratch-file accumulation if a process is killed between
`File::create` and the rename — the guard cannot run then. That was equally true
before, and is now *less* likely to collide with a subsequent write, since the
names differ. The names remain predictable enough to clean up by glob
(`*.kicad_tmp`).

`tempfile` is added as a dev-dependency of `konnect-sexp`; it was already a
dependency of `konnect-core` and is in the lock file, so no new third-party code
enters the tree.

Rollback is a plain revert. Nothing persists across the call, so no state needs
unwinding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -968,6 +968,7 @@ dependencies = [
  "proptest",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror",
  "uuid",
 ]
--- a/crates/konnect-sexp/Cargo.toml
+++ b/crates/konnect-sexp/Cargo.toml
@@ -16,3 +16,4 @@ uuid.workspace = true
 [dev-dependencies]
 pretty_assertions = "1"
 proptest = "1"
+tempfile = "3"
--- a/crates/konnect-sexp/src/writer.rs
+++ b/crates/konnect-sexp/src/writer.rs
@@ -96,24 +96,65 @@ pub fn apply_edits(mut content: String, mut edits: Vec<SexpEdit>) -> String {
 ///
 /// The scratch file is a sibling so the rename stays within one filesystem,
 /// where it is atomic; a temp directory elsewhere would silently degrade to a
-/// copy. It is removed if any step fails, so a failed write leaves nothing
-/// behind in the user's project directory.
+/// copy. A failed write attempts to remove it — best effort, since the removal
+/// can itself fail on a locked or read-only directory, and a write already
+/// failing is the wrong moment to start reporting a second error.
 pub fn write_atomic(path: &Path, content: &str) -> Result<(), SexpError> {
-    let tmp_path = scratch_path_for(path);
+    let (tmp_path, mut file) = create_scratch_file(path)?;
 
     // Remove the scratch file unless the rename below succeeds.
     let mut cleanup = ScratchGuard(Some(tmp_path.clone()));
 
-    {
-        let mut f = std::fs::File::create(&tmp_path)?;
-        f.write_all(content.as_bytes())?;
-        f.flush()?;
-        f.sync_all()?; // fsync — mandatory
-    }
+    file.write_all(content.as_bytes())?;
+    file.flush()?;
+    file.sync_all()?; // fsync — mandatory
+    drop(file);
 
     std::fs::rename(&tmp_path, path)?;
     cleanup.disarm();
     Ok(())
+}
+
+/// Open `path` for writing, failing if anything is already there.
+///
+/// The whole point is what it does *not* do: `File::create` truncates whatever
+/// it finds and follows a symlink to write wherever it points.
+fn open_exclusive(path: &Path) -> std::io::Result<std::fs::File> {
+    std::fs::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .open(path)
+}
+
+/// How many scratch names to try before giving up.
+///
+/// Reaching the end means something is generating files faster than this can
+/// pick names, or is planting them deliberately. Either way, looping forever
+/// would be worse than failing.
+const SCRATCH_ATTEMPTS: u32 = 16;
+
+/// Create a fresh scratch file, never opening one that already exists.
+///
+/// `File::create` truncates whatever it finds, and follows a symlink to write
+/// wherever it points. In a directory the user does not solely control that is
+/// the classic insecure-temp-file shape — a planted symlink turns this write
+/// into a write somewhere else entirely. `create_new` refuses instead, and a
+/// name that is somehow taken is simply exchanged for another.
+fn create_scratch_file(path: &Path) -> Result<(PathBuf, std::fs::File), SexpError> {
+    let mut last = None;
+    for _ in 0..SCRATCH_ATTEMPTS {
+        let candidate = scratch_path_for(path);
+        match open_exclusive(&candidate) {
+            Ok(file) => return Ok((candidate, file)),
+            Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => last = Some(e),
+            Err(e) => return Err(e.into()),
+        }
+    }
+    Err(last
+        .unwrap_or_else(|| {
+            std::io::Error::new(std::io::ErrorKind::AlreadyExists, "no scratch name free")
+        })
+        .into())
 }
 
 /// A scratch path in `path`'s directory, unique to this file and this write.
@@ -164,7 +205,10 @@ impl ScratchGuard {
 impl Drop for ScratchGuard {
     fn drop(&mut self) {
         if let Some(path) = &self.0 {
-            // Already failing; a failure to clean up is not worth masking it.
+            // Best effort by necessity: this runs while a write is already
+            // failing, and a removal that fails too — a locked file, a
+            // read-only directory — has no better outcome to report than the
+            // error already on its way to the caller.
             let _ = std::fs::remove_file(path);
         }
     }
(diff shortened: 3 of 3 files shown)


emolitor:
Adopted the suggested change, but I want to be precise about the reasoning, because the stated failure mode does not hold.

The review says lossy conversion "can reintroduce scratch-path collisions". It cannot. Uniqueness comes from the pid and the counter, not from the name — two calls for the *same* path already get different scratch files, so two paths that merely flatten to the same string were never at risk. I added `identical_names_in_one_directory_still_get_distinct_scratch_paths` to pin that outright rather than leave it to reasoning.

The change is still worth making, for a smaller reason: the name is in the scratch file purely so a leftover from a killed process can be traced back to the file it belonged to, and lossy conversion mangles exactly the names where that matters. Building it as an `OsString` keeps it byte-for-byte and drops an allocation.

Covered on both platforms, since the invalid cases differ — a `0xFF` byte on Unix, an unpaired high surrogate on Windows. The test first asserts its own fixture actually survives a lossy round-trip *changed*, so it cannot quietly degenerate into proving nothing.

emolitor:
Both addressed.

- **`File::create` clobbering an existing scratch path** — a fair hardening point. It truncates whatever it finds and follows a symlink to write wherever it points, which in a directory the user does not solely control is the classic insecure-temp-file shape. The scratch file now opens with `create_new`, and a taken name is exchanged for another (bounded, so a directory being flooded fails rather than spins).

- **"leaves nothing behind"** — correct, the doc overstated it. Cleanup runs in `Drop` and ignores its own errors, so a locked file or read-only directory can still leave one. It is best-effort by necessity: it runs while a write is already failing, and a removal that fails too has no better outcome to report than the error already on its way to the caller. Comment and guard now both say so.

Worth recording a mistake I made here: my first clobber test squatted the name `write_atomic` would pick next and asserted it survived. That passes whatever the opener does — the counter has already moved past that name by the time the write runs — so it proved nothing. The test now exercises the opener directly and does fail against `File::create`.

mixelpixx (approved):
Approving. The diagnosis is exactly right and independently verified: `with_extension("kicad_tmp")` collapses `board.kicad_pcb`/`board.kicad_sch`/`board.kicad_pro` to one scratch path, and `write_atomic` is the universal write path — so any overlapping writes in a project could publish each other's half-written bytes. Your issue write-up (#71) was a model bug report.

The fix goes beyond what the issue asked for, in the right ways: pid+counter uniqueness on all three axes, `create_new` exclusive creation with bounded retry (which also closes the planted-symlink/insecure-temp shape `File::create` had), same-directory rename preserved so atomicity is real, fsync unchanged, and the RAII guard that never masks the original error. The `OsString` reasoning for non-Unicode names — uniqueness from pid+counter, name only for traceability — is correct, and the test suite states its preconditions unusually well (the test that demonstrates the old derivation collapsing is exactly how regression tests should be written).

Verified locally on Windows: full suite including the threaded interleave tests, 307 workspace tests green on the branch merged with current main (the anticipated `writer.rs` tail conflict with #69 auto-resolved cleanly; branch updated with that merge).

Noted for later rather than blocking: uniquely-named leftovers from killed processes can now accumulate where the old single name self-overwrote — a `*.kicad_tmp` sweep at startup or on project open would tidy that, and is strictly follow-up material.

PR #95, opened 2026-07-27
fix(sch): place schematic field properties near their owners

Nigh:
## Summary
- `add_power_symbol` wrote bare `Reference`/`Value` properties with no `(at …)`, so KiCad defaulted them to `(0,0)` and every `#PWR*` piled up in the top-left corner.
- `Sheet::new` (used by `add_hierarchical_sheet` / `duplicate_sheet`) had the same bug for `Sheetname` / `Sheetfile`.
- Write absolute property positions next to the owner (symbol or sheet box). Hide `#PWR` inside effects (eeschema-compatible); keep Value visible.
- Add regression tests for both paths.

## Test plan
- [x] `cargo test -p konnect-core --lib power_symbol_tests`
- [x] `cargo test -p konnect-schematic-editor --lib new_sheet_places_name_and_file_near_the_box`
- [ ] In KiCad: place a power symbol via `add_power_symbol` and confirm fields are not at sheet origin; Value appears near the symbol
- [ ] In KiCad: `add_hierarchical_sheet` and confirm Sheetname sits above the box and Sheetfile below it
- [ ] Re-open the schematic after save and confirm positions survive

### DIFF
--- a/crates/konnect-core/src/tools/sch_wiring.rs
+++ b/crates/konnect-core/src/tools/sch_wiring.rs
@@ -1156,11 +1156,56 @@ async fn handle_add_power_symbol(
     sym.in_bom = true;
     sym.on_board = true;
     sym.uuid = uuid::Uuid::new_v4().to_string();
-    sym.properties
-        .push(cse::Property::new("Reference", &pwr_ref));
-    sym.properties.push(cse::Property::new("Value", &power_net));
-    sym.properties.push(cse::Property::new("Footprint", ""));
-    sym.properties.push(cse::Property::new("Datasheet", ""));
+
+    // Property (at …) is absolute sheet coords — same as add_schematic_component.
+    // Bare Property::new writes no (at); KiCad then defaults to (0,0) and every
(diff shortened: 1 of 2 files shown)


mixelpixx:
Confirmed the bug on main (properties written with no `(at)`/`(effects)` node, so every `#PWR` reference and value landed at the sheet origin, reference visible) and reviewed the fix — it exactly mirrors the already-correct `add_schematic_component` property placement, with the Reference hidden per eeschema convention for power symbols. The hermetic regression test is the first of its kind for property placement in this repo. Nice first contribution, thank you!

For the record: the earlier red CI on this PR was our flake (an ephemeral-port race in the konnect-ipc mock-server tests, fixed on main in 639bdf9), not anything in this change. Two known follow-ups we'll handle maintainer-side: deduplicating the now-twice-copied `effects_node`/`at_node` closures into a shared helper, and rotation-aware property placement (a pre-existing limitation shared with `add_schematic_component`).

PR #96, opened 2026-07-28
fix(sch): hide power Reference with KiCad 10 property-level hide

Nigh:
## Summary
- Depends on / stacks after #95 (field positions). Unique commit: KiCad 10 **property-level** `(hide yes)` for power symbol Reference.
- Official `power.kicad_sym` puts `(hide yes)` as a sibling of `effects`, not inside it — effects-only hide can leave `#PWR*` visible in KiCad 10.
- `add_power_symbol`: Reference / Footprint / Datasheet use property-level hide; Value stays visible.
- `add_schematic_component`: same when `lib_id` is `power:*` or reference is `#PWR*`.

## Test plan
- [x] `cargo test -p konnect-core --lib power_symbol_tests`
- [x] `cargo test -p konnect-core --lib hides_power_reference`
- [ ] In KiCad: `add_power_symbol` — `#PWR*` invisible, Value visible near symbol
- [ ] In KiCad: `add_schematic_component` with `power:GND` — same
- [ ] After #95 merges, rebase this branch onto `main` so the PR diff is only the hide commit

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/konnect-core/src/tools/sch_components.rs
+++ b/crates/konnect-core/src/tools/sch_components.rs
@@ -360,6 +360,8 @@ async fn handle_add_schematic_component(
     sym.unit = unit;
 
     // Helper: build an effects sub-node  (font (size 1.27 1.27))  with optional (hide yes)
+    // inside effects (legacy). Prefer property-level `(hide yes)` for power refs —
+    // that is what KiCad 10's power.kicad_sym uses.
     let effects_node = |hide: bool| -> cse::sexp::SexpNode {
         let font = cse::sexp::SexpNode::List(vec![
             cse::sexp::atom("font"),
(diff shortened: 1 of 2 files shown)


mixelpixx:
Verified your property-level `(hide yes)` claim against ground truth — both KiCad 10's `power.kicad_symdir/GND.kicad_sym` and instances written by eeschema in the CM5 demo project use exactly `(at …) (hide yes) (effects …)` — so I've adopted it as the canonical form. Since #95 merged, main had refactored both property blocks into a shared `positioned_property` helper, so I resolved the conflict on your branch (thanks for allowing maintainer edits): the helper itself now emits the property-level hide, which upgrades every call site at once. Kept intact from your branch: the power-reference hiding in `add_schematic_component`, the hierarchical-sheet Sheetname/Sheetfile placement fix (nice catch — same bug class), and your stricter hide-before-effects test assertions. Will merge when CI is green.

Issue #35, opened 2026-07-15
Multi-unit and (extends)-derived symbols: no unit selection, empty or superimposed pin lists, extends-stub embeds kicad-cli can't resolve


tascsystems-andrew:
**Severity: major** — opamps, dual triodes, and most transistor variants are effectively unusable via the schematic tools.

## Repro (d70b561, KiCad 10.0.3)

Library facts: `Amplifier_Operational:NE5532` is `(extends "LM2904")` with no pins of its own; `LM2904` has unit sub-symbols `_1_1` (pins 1-3), `_2_1` (5-7), `_3_1` (4,8 power).

1. `add_schematic_component` has **no `unit` argument**; passing `"unit": 2` is silently ignored (no schema rejection). Every placement writes `(unit 1)`.
2. `get_schematic_pin_locations` on NE5532 (derived) → `"pins": []`, `isError: false`; `connect_pins` on it hard-fails ("Pin not found").
3. `get_schematic_pin_locations` on LM2904 (base, multi-unit) → **all three units' pins superimposed** on the single placement (pin 3 and pin 5 at identical coords, etc.).
4. The lib_symbols embed for NE5532 is an `extends` **stub** (parent embedded separately, in full). kicad-cli 10.0.3 can't resolve it: the netlist shows a pinless NE5532 libpart. eeschema itself *flattens* derived symbols when it saves — the stub shape doesn't occur in eeschema-written files.

## Suggested fixes

1. Optional integer `unit` (default 1) on `add_schematic_component` / `replace_component`; validate against the unit count of the resolved symbol; write it to both the symbol and the instances entry.
2. Flatten derived symbols on embed the way eeschema does (deep-copy parent unit sub-symbols renamed `Parent_N_M` → `Derived_N_M`, overlay the child's properties) instead of writing an extends stub.
3. Make pin extraction unit-aware: keep pins from sub-symbols where `N == 0` (common) or `N == instance unit`; follow `(extends)` to the parent for the pin set.

Same session as #29-#34; overlaps with #34 only in that both end in silent `pins: []` — the mechanisms are different (missing symbol vs. unit/extends-unaware pin model).

mixelpixx:
[...]

PR #104, opened 2026-07-29
fix(sch): connect mid-wire pins via junction dots

Nigh:
## Summary

A user reported that a power-symbol/PWR_FLAG pin sitting mid-wire does not connect even with a junction dot (ERC "pin not connected"), requiring a `split_wire_at_point` workaround. Investigation confirmed two bugs (and one misconception):

- Verified with kicad-cli 10.0.4: KiCad connects a mid-wire pin through a junction dot alone — splitting the wire is unnecessary. The "even with a junction" failure was Konnect's own validators.
- **Bug 1:** the union-find `NetGraph` only joined wire *endpoints* and never read junctions, so every connectivity/validation tool reported mid-wire pins as unconnected regardless of junctions.
- **Bug 2:** `find_t_junctions` only handles wire-on-wire T's, so placing a component whose pin lands mid-wire (or drawing a wire across an existing pin) never emitted the junction dot KiCad needs — producing real ERC failures.

## Changes

- `konnect-sexp`: new `extract_junctions`.
- `build_net_graph` takes junctions and unions any label/junction point lying mid-segment on a wire with that wire; all call sites updated.
- `validate_component_connections` accepts pin-mid-wire + junction, still flags the no-junction case (matches KiCad ERC semantics).
- `add_schematic_component` / `add_power_symbol` auto-insert a junction when a placed pin lands mid-wire (reported as `junctions_added`).
- `add_wire` / `batch_add_wire` / `connect_to_net` / `connect_pins` auto-insert a junction for existing pins the new wire passes over.
- Tool descriptions for `split_wire_at_point` / `add_junction` updated to reflect that a junction alone suffices.

## Test plan

- [x] New regression tests: `midwire_pin_connects_with_junction_only`, `add_wire_over_pin_inserts_junction`
- [x] `cargo test -p konnect-core -p konnect-sexp` — 288 passed

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -360,8 +360,8 @@ pub(crate) fn add_pin_midwire_junctions(
 ) -> anyhow::Result<Vec<(f64, f64)>> {
     use konnect_sexp::geometry::{point_on_segment, points_coincident};
     use konnect_sexp::schematic::{
-        extract_junctions, extract_lib_pins, extract_symbol_instances, extract_wires,
-        pin_endpoint, read_schematic,
+        extract_junctions, extract_lib_pins, extract_symbol_instances, extract_wires, pin_endpoint,
+        read_schematic,
     };
     let tol = 0.01;
     let (_, tree) = read_schematic(sch_path)?;
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -328,6 +328,90 @@ pub fn ensure_root_uuid(sch: &mut konnect_schematic_editor::Schematic) -> String
     }
 }
 
+/// All symbol pin connection points in a parsed schematic tree.
+pub(crate) fn all_pin_endpoints(tree: &konnect_sexp::SexpNode) -> Vec<(f64, f64)> {
+    use konnect_sexp::schematic::{extract_lib_pins, extract_symbol_instances, pin_endpoint};
+    let lib_syms = tree
+        .find("lib_symbols")
+        .map(|n| n.find_all("symbol"))
+        .unwrap_or_default();
+    let mut pts = Vec::new();
+    for inst in extract_symbol_instances(tree) {
+        if let Some(sym) = lib_syms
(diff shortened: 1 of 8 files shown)


mixelpixx:
Verified and merged main in on your branch (thanks for allowing maintainer edits) — this needed a real reconcile because #111 landed `batch_place_components`/`batch_connect_pins` in the same three files while this was open.

**On the diagnosis:** both bugs confirmed, and I checked the KiCad-semantics claim independently rather than taking it on trust. `CONNECTION_GRAPH::updateItemConnectivity` has an explicit junction branch calling `GetBusesAndWires(point, aIgnoreEndpoints=true)`, so a junction does register the unsplit wire at that point — your kicad-cli result and the netlister source agree. Worth noting the file shape is non-canonical though: across KiCad's own 115 demo schematics, **0 of 4293 junctions sit mid-segment** — eeschema always splits. Functionally fine, but a user dragging near that point in eeschema will see KiCad normalize it, so I've noted the divergence.

**The under-advertised win:** your `attach` closure also unions *labels* with the segment they lie on. In that same demo corpus, **3712 of 5804 labels (64%) sit strictly mid-segment** — main was silently dropping all of them from net formation. That's a bigger correctness recovery than the junction fix itself and it'll lead the release notes.

**What I changed on top:**
1. **Unit-awareness (the one real defect).** `all_pin_endpoints` and `add_pin_midwire_junctions` used `extract_lib_pins` while the rest of the write path uses `extract_lib_pins_for_unit` (#35). A multi-unit symbol superimposes every unit's pins on one placement, so placing unit 1 of an LM2904 could drop a dot on a phantom unit-2 pin — and where two wires cross, that's a short written into the user's file. Both helpers are unit-aware now.
2. **Reconcile with #111**: `place_one_component` is pure by contract (callers own the read/write), so the junction pass moved to `handle_add_schematic_component` after the write — which also addresses the double-write you'd have had per placement, and leaves the batch path able to do one pass per batch.
3. Deduplicated the two `extract_junctions` calls that the merge left behind.

Full gate green, 403 tests. Merging once CI confirms.

Issue #67, opened 2026-07-25
tools/list not working for toolset?


Myseka:
Maybe this is a me issue but I'm running into errors early when I simply try to use the basic functionality, it will lookup a toolset then tools/list isn't updated.

Myseka:
*think* I got it, looks like it was the tool search in the chrome plugin conflicting.

mixelpixx:
Closing per your own follow-up — glad it resolved. For the record after verifying server-side: the notification path is intact on current main (`tools.listChanged: true` advertised, `notifications/tools/list_changed` emitted after every `load_toolset`/`unload_toolset`, and both transports drain server-initiated notifications — this is the machinery PR #21 fixed and it has regression coverage). The symptom you saw matches a client-side tool-search/filter layer deferring or shadowing the updated list, which is exactly what you found with the Chrome-plugin tool search.

A TROUBLESHOOTING entry covering "tools don't appear after load_toolset" (check that your client honors `tools/list_changed`, disable competing tool-filter layers, re-issue `tools/list`) is queued for the next docs commit so the next person hits a doc instead of filing an issue.

PR #110, opened 2026-07-30
feat(router): load_toolset accepts an array of toolset names

gcolonese:
## Problem / scope

Loading toolsets one at a time fired one `tools/list_changed` notification per call, forcing clients to re-fetch the entire cumulative tool list each time. Measured on a scripted stdio session loading 7 toolsets sequentially: 220,334 bytes (~55K tokens) of `tools/list` traffic; batching the same 7 loads into one call drops that to 53,707 bytes (~13.4K tokens) -- a 4.1x reduction. This measurement comes from batching alone; it doesn't depend on auto-load.

## Root cause / design

`load_toolset`'s `name` argument now also accepts an array, so a client can load several toolsets in one call with a single `tools/list_changed` notification instead of N. Duplicate names are deduped so `tools_added` reflects reality.

**Result shape**, called out explicitly per @mixelpixx's review since tool schemas are public API:
- Single-name (string) input produces a result byte-identical to before this PR: `loaded` is the string, `tools` is `[{name, description}]`.
- Array input is new documented surface: `loaded` is the array of names that loaded, `tools` is the union of every loaded toolset's `{name, description}` entries (same per-item shape as the string form, not bare name strings), and `errors` lists any unknown names -- each carrying the `Call list_toolboxes() to see valid names` recovery hint.
- Partial success (some names load, some are unknown) is not an error.
- Total failure (every name unknown) returns `CallToolResult::error_kind(ToolErrorKind::InvalidArgument { field: "name", .. })` rather than a JSON body with a hand-set `is_error` flag, so `extract_error_kind` still populates the observer's typed column instead of degrading to `handler_error`.

This PR carries no auto-load behavior. An unloaded tool still returns the original `toolset_not_loaded` error naming the owning toolset, unchanged from `main` before this branch -- `dispatch_tool` in `handler.rs` is untouched.

## Compatibility

Additive and backward compatible: `load_toolset(name)` with a single string is unchanged input *and* output. Passing an array is new. `unload_toolset` is untouched (its description text is back to the pre-PR wording -- the auto-reload sentence didn't belong here).

## Tests run

`load_toolset_batch_form_loads_all_and_notifies_once` (batch load + single notification + tools-item shape) and a new `load_toolset_batch_total_failure_is_typed_error` test (asserts the typed `invalid_argument` kind and the `list_toolboxes` hint in the message) in `crates/konnect/tests/protocol_stdio.rs`. `structured_errors_guide_recovery` is back to asserting `toolset_not_loaded` for an unloaded tool, since that behavior is unchanged. `cargo test --workspace --locked --lib --tests` (19 suites on current `main`) and `--doc` pass; `cargo clippy --workspace --locked -- -D warnings` and `cargo fmt --all -- --check` are clean.

## Risk / rollback

Low. Two-commit revert available (feat + docs). The only behavioral change from current `main` is additive (array input); every existing single-name call path is untouched.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/crates/konnect-core/src/router/meta_tools.rs
+++ b/crates/konnect-core/src/router/meta_tools.rs
@@ -1,7 +1,7 @@
 //! The 6 always-visible meta-tools.
 //!
 //! Discovery / routing:
-//!   list_toolboxes()          — show all 17 toolsets with descriptions and load state
+//!   list_toolboxes()          — show all 18 toolsets with descriptions and load state
 //!   load_toolset(name)        — activate a toolset, expose its tools in tools/list
 //!   unload_toolset(name)      — deactivate a toolset, remove its tools from tools/list
 //!   get_active_toolsets()     — list currently loaded toolsets
@@ -14,6 +14,7 @@
 //! baseline context stays small. The LLM reads `list_toolboxes` and calls
 //! `load_toolset(name)` to expose the tools it actually needs for the task.
 
+use crate::mcp::error::ToolErrorKind;
 use crate::mcp::protocol::{CallToolResult, McpToolDescription};
 use crate::tools::ToolContext;
 use serde_json::{json, Value};
@@ -41,14 +42,18 @@ pub fn meta_tool_descriptions() -> Vec<McpToolDescription> {
             description:
                 "Load a toolset by name so its tools appear in tools/list and can be called. \
                  Returns the list of tools that were added. Use list_toolboxes() first to \
-                 see valid names."
+                 see valid names. Pass an array to load several toolsets in one call -- \
+                 cheaper, one tools/list refresh."
                     .to_string(),
             input_schema: json!({
                 "type": "object",
                 "properties": {
                     "name": {
-                        "type": "string",
-                        "description": "Toolset name (e.g. 'sch_components', 'pcb_routing')"
+                        "anyOf": [
+                            {"type": "string"},
+                            {"type": "array", "items": {"type": "string"}}
+                        ],
+                        "description": "Toolset name (e.g. 'sch_components', 'pcb_routing'), or an array of names"
                     }
                 },
                 "required": ["name"]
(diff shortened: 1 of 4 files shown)


mixelpixx:
Thanks for this — the two halves need to go different ways, so I'd like to split it.

**The batch `load_toolset` half is good** and I'd merge it today as its own PR. Taking a list instead of one name is a clean ergonomic win, the dedup is right, and the test is sound.

**The auto-load half I can't take as-is**, for three reasons:

1. **It changes `load_toolset`'s output schema without saying so.** `loaded` goes string → array and `tools` goes `[{name, description}]` → `["name"]`. The PR body calls it backward compatible, which is true of the *input* but not the result. CONTRIBUTING treats tool schemas as public API needing an explicit migration.

2. **It drops the recovery hint.** The old error said `Call list_toolboxes() to see valid names`; the new one is a bare `Unknown toolset 'x'`, built with `CallToolResult::json` + manual `is_error` rather than `error_kind`, so `extract_error_kind` degrades to `handler_error` and the observer loses the typed column.

3. **It inverts what the router is for.** Toolsets accumulate monotonically — any tool name the model emits that exists anywhere in the 187 permanently adds up to ~19 descriptions to every later `tools/list`. The failure it replaces costs one structured error plus one `load_toolset` call; the new one costs ~2 KB of context for the rest of the session, in a router whose entire job is keeping listings near 2 K tokens. `unload_toolset` also becomes advisory — your own doc edit says as much ("unload prunes context, it does not disable").

Credit where due: the `tools/list_changed` notification *is* correctly preserved through `notify_tools_list_changed`, and the new stdio test asserting it fires on auto-load is exactly the right test. I checked #67's history too — that was a client-side tool-filter issue, so nothing here regresses it.

**What I'd merge:** the batch half on its own, plus auto-load behind an opt-in (config key or `initialize` capability) defaulting **off**, with the `load_toolset` result shape preserved and the `list_toolboxes()` hint restored. If you'd rather not do the split, say so and I'll cherry-pick the batch half with credit to you — but I'd rather it stayed your PR.

For rebasing: main has moved a fair bit today (#106, #107, #109, #111, #112, plus an add_via regression fix), and #111 touched `registry.rs` tool counts, so expect a small conflict in `tool-directory.md`.

gcolonese:
Thanks for the review -- split as requested.

This PR is reworked to batch-only: `handler.rs`'s dispatch is back to the original `toolset_not_loaded` arm verbatim, `load_toolset`'s string-form result is byte-identical to before, the array form's `tools` entries are `{name, description}` objects (not bare strings), the `list_toolboxes()` recovery hint is back in every unknown-toolset error, and total failure (all names unknown) now returns a typed `ToolErrorKind::InvalidArgument` via `error_kind` instead of a hand-built `is_error` JSON body. Rebased onto current `main`; resolved the `tool-directory.md` conflict from #111.

Auto-load is resubmitted as a stacked opt-in PR, defaulting off: #132.

PR #136, opened 2026-08-08
fix(library): resolve symbols through sym-lib-table

### LINKED ISSUE #105: add_schematic_component (and ensure_lib_symbol/symbol_unit_count) never resolves custom registered libraries — sym-lib-table is never read
## Summary

`add_schematic_component` (and everything downstream of it: `ensure_lib_symbol`,
`symbol_unit_count`) can never resolve a symbol from a **custom registered library**
— project-scope or global — even when it's correctly registered in `sym-lib-table`.
Only libraries that live directly inside KiCad's own install tree resolve. This
makes it impossible to place any custom/project-specific symbol via MCP; only
stock KiCad libraries (`Device`, `power`, etc.) work.

## Repro

1. Create a custom symbol library, e.g. `MyProject.kicad_sym`, with one symbol `Foo`.
2. Register it at project scope:
   ```
   register_symbol_library(library_path="<abs path>/MyProject.kicad_sym", nickname="MyProject", scope="project")
   ```
   This correctly writes `<project_dir>/sym-lib-table`:
   ```
   (sym_lib_table
     (version 7)
     (lib (name "MyProject") (type "KiCad") (uri "<abs path>/MyProject.kicad_sym") (options "") (descr ""))
   )
   ```
3. `list_symbol_libraries()` correctly reports `MyProject` as registered.
4. `add_schematic_component(schematic="<project_dir>/some.kicad_sch", lib_id="MyProject:Foo", x=10, y=10)` fails with:
   ```
   Library 'MyProject' not found in the installed KiCAD symbol libraries (lib_id 'MyProject:Foo').
   Check the library name, the KiCAD install, or KICAD10_SYMBOL_DIR.
   ```
   This happens regardless of whether the registration is project-scope or global-scope,
   and regardless of restarting KiCad — because KiCad's own state is never involved in
   this resolution path.

## Root cause

`add_schematic_component`'s embed path
(`sch_components.rs::handle_add_schematic_component` → `cse::library::ensure_lib_symbol`
→ `resolve_lib_symbol_flattened_node` → `resolve_lib_symbol_node` → `resolve_lib_symbol`,
the last three all in `crates/konnect-schematic-editor/src/library.rs`) **never reads
`sym-lib-table` at all**, project or global. It only:

1. Checks a `KICAD10_SYMBOL_DIR` (etc.) environment variable, and
2. Scans `find_symbol_dirs()` — a hardcoded list of KiCad's own install locations
   (`C:\Program Files\KiCad\10.0\share\kicad\symbols`, `/usr/share/kicad/symbols`, the
   macOS app bundle path, etc.) — for a directory or file literally named
   `<library_name>.kicad_symdir` / `<library_name>.kicad_sym`.

`register_symbol_library`/`register_footprint_library` correctly write the lib-table
files, and `list_symbol_libraries` correctly reads them back — but nothing in the
actual symbol-placement path ever consults that table. So registration silently
has no effect on `add_schematic_component`, `ensure_lib_symbol`, or `symbol_unit_count`.
This is a missing feature in the resolver, not a caching/KiCad-restart issue, which
is why restarting KiCad's UI (a natural first troubleshooting step) doesn't help —
KiCad's own process state is irrelevant to this code path.

The confusing part while debugging this from the outside: `list_symbol_libraries`
*does* reflect new registrations immediately, and even `get_symbol_info` on the
same `lib_id` works — because those two go through a different lookup than
`add_schematic_component` does. That divergence (two resolution paths, one that
knows about lib-tables and one that doesn't) is what made this look like a stale
cache rather than a missing code path.

## Suggested fix

Add a lib-table-aware resolution path used by the component-placement chain.
Sketch (this is what I implemented locally to unblock myself, happy to open a PR
if useful instead of just describing it):

- `Schematic` already tracks its own file path (`schematic.filepath()`) — that's
  enough to derive "the project directory" without any new parameters on the
  MCP-facing tools.
- Add `resolve_lib_symbol_in(lib_id, project_dir: Option<&Path>)` (and node/flattened/
  unit-count siblings) that:
  1. Looks up `library_name` in `<project_dir>/sym-lib-table` (if given), then in
     each global `sym-lib-table` (`%APPDATA%\kicad\<ver>\`, `~/Library/Preferences/kicad/<ver>/`,
     `~/.config/kicad/<ver>/` per OS, newest version first — mirroring how
     `find_symbol_dirs` already enumerates per-version candidates),
  2. On a match, reads the symbol directly from the registered URI (single-file
     `.kicad_sym`, or `.kicad_symdir`-style directory),
  3. Falls back to the existing `find_symbol_dirs()` scan when there's no lib-table
     hit, so stock-library resolution is unaffected.
- `ensure_lib_symbol(schematic, lib_id)` calls the `_in` variant with
  `schematic.filepath().parent()`; `symbol_unit_count` gets an `_in` sibling used
  the same way in `handle_add_schematic_component`.
- Old function names/signatures stay as thin `None`-project wrappers, so nothing
  else in the codebase needs to change.

Tested locally against a real custom project library (multi-unit symbols
included) — resolves correctly at both project and global scope, with the
existing stock-library behavior unaffected.

## Environment

- Windows, KiCad 10.0
- Built from `main` @ `da78ea8` (just past v0.2.1)

pauliuszaleckas:
Fixes #105.

## Problem

Symbol placement never read a lib table — `resolve_lib_symbol` scanned the KiCad install
dirs for `<nickname>.kicad_sym`. A library registered with `register_symbol_library` was
therefore invisible to `add_schematic_component`: registration reported success, then
placement failed blaming `KICAD10_SYMBOL_DIR`. **No custom symbol could be placed at
all** — the normal case for a project carrying its own parts.

Footprints already resolved correctly (project `fp-lib-table` → global → conventional
layout, following nested `(type "Table")` entries and expanding `${KIPRJMOD}`). Symbols
now use that same machinery.

## Changes

`konnect-schematic-editor` gains a `SymbolLibrarySource` trait and stops probing the
environment; `konnect-core` implements it as `KiCadSymbolSource`. The trait sits in the
lower crate because the dependency runs that way — interface where it is consumed, impl
where the KiCad-config knowledge already lives. Candidates are full paths rather than
directories searched by convention, so a nickname may map to any filename:
`(name "MyParts") (uri ".../vendor.kicad_sym")` is valid KiCad and was previously
unresolvable in principle, not just in practice.

Two further defects share the root cause:

- **Three drifted copies of install-path probing.** The one on the placement path knew
  only `KICAD10_SYMBOL_DIR`, `/usr/share` and `/usr/local/share`, so on flatpak, snap or
  `/opt` it resolved *no* symbol — `Device:R` included — while footprints and
  `get_symbol_info` worked on the same machine. `find_symbol_dirs`, konnect-core's
  duplicate `resolve_lib_symbol` and `find_kicad_symbol_dirs` are deleted; everything
  routes through `find_kicad_library_dirs`.
- **`resolve_symbol_lib_path` searched global before project**, resolving a shared
  nickname to the wrong file. It is now a thin wrapper over `KiCadSymbolSource`, so
  `get_symbol_info` and placement cannot disagree, and it gains the install-dir fallback
  it lacked — which also fixes symdir libraries, previously a `read_to_string` on a
  directory.

`lib_table_target()` is shared by both registrars: project scope writes `${KIPRJMOD}/…`
as KiCad does, and both carry the empty-parent guard —
`Path::new("board.kicad_pro").parent()` is `Some("")`, not `None`, which produced a
corrupt `${KIPRJMOD}//abs/path`. The not-found error now names the tables searched
instead of blaming `KICAD10_SYMBOL_DIR`.

## Review notes

- Follows the plan in #105 (thread the project dir through the editor-crate resolvers,
  try the table before the stock scan, consolidate the two drifted copies) with one
  deliberate deviation: `resolve_symbol_lib_path`'s **global-then-project** order is
  flipped to **project-then-global**, matching KiCad and `resolve_footprint_path`.
  Global-first meant a project entry could not shadow a same-nickname global one, which
  is the case @ncolomer hit in step 3 of the repro.
- Not covered: passing an absolute path as the library half of a lib_id
  (`"/abs/MyLib.kicad_sym:Part"`, step 5 of that repro). That was diagnostic rather than a
  request — lib_ids stay `Nickname:Symbol`, as in KiCad.
- **Beyond the title:** footprint registration now writes `${KIPRJMOD}` URIs for project
  scope where it previously wrote absolute paths (`expand_lib_uri` already understood
  them). Happy to split out if you would rather keep this symbol-only.
- `KiCadSymbolSource` memoises tables and install dirs in `OnceLock`s: the default global
  table indirects to ~200 bundled entries that each re-probe the install roots when
  `${KICAD*_DIR}` is unset, and one placement asks for candidates at least twice.

## Testing

Gate green — `cargo fmt --check`, `cargo clippy --workspace -D warnings`, 484 tests, 3
doctests. `crates/schematic-viewer` untouched.

New coverage: a single-file library whose filename differs from its nickname (the shape
that could not be placed before); project entries preceding install dirs; `${KIPRJMOD}`
expanding against the table's own directory; the empty-`table_dir` regression; both
registrars agreeing on table path and URI.

Also worth a look: `sch_components`' stub library now registers itself through a project
`sym-lib-table`. The global table's `Device` entry had been resolving to the developer's
*real* installed library and silently shadowing the fixture — those tests passed only
because real `Device:R` also has two pins. A `stub://device` marker now asserts the
fixture is the one in use, verified to fail without the fix.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -54,7 +54,7 @@ Konnect/
 │   │       │   ├── registry.rs      # Static toolset metadata + tools_for() dispatcher
 │   │       │   └── meta_tools.rs    # 6 always-visible meta-tools
 │   │       └── tools/
-│   │           ├── mod.rs            # ToolDef, ToolContext, tool! macro, helpers, kicad_config_dir(), resolve_lib_symbol()
+│   │           ├── mod.rs            # ToolDef, ToolContext, tool! macro, helpers, kicad_config_dir()
 │   │           ├── cli.rs            # kicad-cli v10 subprocess wrapper (verified against actual binary)
 │   │           ├── svg_import.rs     # SVG parsing + Bezier flattening for import_svg_logo (usvg-backed)
 │   │           ├── project.rs        # 6 tools (incl. open_schematic_viewer)
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -838,6 +838,73 @@ fn global_sym_lib_table() -> PathBuf {
     super::kicad_config_dir().join("sym-lib-table")
 }
 
+/// Symbol libraries resolved as KiCad does: project `sym-lib-table` (shadowing
+/// same-nickname global entries), then global, then the conventional
+/// `<nickname>.kicad_symdir` / `.kicad_sym` layout. Same order as
+/// [`resolve_footprint_path`]; the last step covers a missing global table.
+///
+/// Both the flattened tables and the install dirs are memoised: placing one
+/// component asks for candidates at least twice, and the default global table
+/// is a `(type "Table")` indirection to ~200 bundled entries, each of which
+/// re-probes the install roots when `${KICAD*_DIR}` is unset.
+pub(crate) struct KiCadSymbolSource {
+    project_dir: Option<PathBuf>,
+    tables: std::sync::OnceLock<Vec<serde_json::Value>>,
+    install_dirs: std::sync::OnceLock<Vec<PathBuf>>,
+}
+
+impl KiCadSymbolSource {
+    pub(crate) fn new(project_dir: Option<PathBuf>) -> Self {
+        Self {
+            project_dir,
+            tables: std::sync::OnceLock::new(),
+            install_dirs: std::sync::OnceLock::new(),
+        }
+    }
+
+    /// For a `.kicad_sch` or `.kicad_pcb` — `sym-lib-table` sits beside it.
+    pub(crate) fn for_file(file: &Path) -> Self {
+        Self::new(file.parent().map(Path::to_path_buf))
+    }
+
+    /// Project entries first so they shadow same-nickname global ones.
+    fn tables(&self) -> &[serde_json::Value] {
+        self.tables.get_or_init(|| {
+            let mut libs = Vec::new();
+            if let Some(dir) = &self.project_dir {
+                libs.extend(read_flat_lib_table(&dir.join("sym-lib-table")));
+            }
+            libs.extend(read_flat_lib_table(&global_sym_lib_table()));
+            libs
+        })
+    }
+}
+
+impl konnect_schematic_editor::library::SymbolLibrarySource for KiCadSymbolSource {
+    fn candidates(&self, nickname: &str) -> Vec<PathBuf> {
+        let mut out = Vec::new();
+
+        for lib in self
+            .tables()
+            .iter()
+            .filter(|l| l["nickname"].as_str() == Some(nickname))
+        {
+            if let Some(path) = lib["path"].as_str() {
+                out.push(PathBuf::from(path));
+            }
+        }
+
+        let install_dirs = self
+            .install_dirs
+            .get_or_init(|| super::find_kicad_library_dirs("symbols"));
+        for base in install_dirs {
+            out.push(base.join(format!("{}.kicad_symdir", nickname)));
+            out.push(base.join(format!("{}.kicad_sym", nickname)));
+        }
+        out
+    }
+}
+
 /// Parse a lib-table S-expression and return list of (nickname, uri, type) tuples.
 ///
 /// Indentation-agnostic: KiCad's own writers emit tab-indented, CRLF-terminated
@@ -1075,8 +1142,8 @@ pub(crate) fn footprint_lib_nickname_for_dir(dir: &Path) -> Option<String> {
 /// registers. The `.pretty` fallback covers a stock install whose global
 /// table is missing or unreadable.
 ///
-/// (`resolve_symbol_lib_path` still searches global-first for symbols; that
-/// asymmetry is pre-existing and noted on that function.)
+/// Symbols resolve the same way and in the same order, via
+/// [`KiCadSymbolSource`] and `resolve_symbol_lib_path`.
 pub(crate) fn resolve_footprint_path(
     reference: &str,
     project_dir: Option<&Path>,
@@ -1185,26 +1252,18 @@ async fn handle_register_footprint_library(
     let nickname = require_str(args, "nickname").map_err(|e| anyhow::anyhow!("{:?}", e))?;
     let scope = args["scope"].as_str().unwrap_or("project");
 
-    let table_path = if scope == "global" {
-        global_fp_lib_table()
-    } else if let Some(proj) = args["project"].as_str() {
-        PathBuf::from(proj)
-            .parent()
-            .unwrap_or(Path::new("."))
-            .join("fp-lib-table")
-    } else {
-        return Ok(CallToolResult::error(
-            "For project scope, provide 'project' path to .kicad_pro file",
-        ));
+    let (table_path, uri) = match lib_table_target(
+        scope,
+        args["project"].as_str(),
+        &lib_path,
+        global_fp_lib_table(),
+        "fp-lib-table",
+    ) {
+        Ok(target) => target,
+        Err(e) => return Ok(e),
     };
 
-    register_in_lib_table(
-        &table_path,
-        nickname,
-        lib_path.to_str().unwrap_or(""),
-        "KiCad",
-    )
-    .await?;
+    register_in_lib_table(&table_path, nickname, &uri, "KiCad").await?;
 
     Ok(CallToolResult::text(
         serde_json::to_string(&json!({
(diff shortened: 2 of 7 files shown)


anyn99:
Tested this branch against KiCad 10.0.4 on Windows, merged onto current `main`,
driving the MCP server directly over its HTTP transport. It fixes the blocker
for me: before this, no project-registered symbol could be placed at all.

| Case | Result |
|---|---|
| Project symbol, no `project_dir` passed | ✅ placed, 4/4 pins |
| Real design: `TXS0104E` from a project `.kicad_sym`, registered via `${KIPRJMOD}` | ✅ placed, 14/14 pins |
| Schematic with no `.kicad_pro` beside it, standard library | ✅ `Device:R` placed |
| Schematic with no `.kicad_pro` beside it, project library | ✅ honest error, no crash |
| `get_symbol_info` with explicit `project_dir` (regression check) | ✅ unchanged |
| Schematic in a **subdirectory** of the project | ❌ see below |

Deriving the project directory from the schematic path is what makes this work
without threading a `project_dir` argument through every tool — nice.

### One edge case: sheets in a subdirectory

```
add_schematic_component(schematic="<proj>/sheets/child.kicad_sch",
                        lib_id="MyLib:TESTCHIP", …)
→ Library 'MyLib' not found in the project or global sym-lib-table
```

The project's `sym-lib-table` sits at `<proj>/`, but the lookup only checks the
schematic's own directory:

```rust
/// For a `.kicad_sch` or `.kicad_pcb` — `sym-lib-table` sits beside it.
pub(crate) fn for_file(file: &Path) -> Self {
    Self::new(file.parent().map(Path::to_path_buf))
}
```

That assumption holds for a root schematic, but not for a hierarchical sheet:
KiCad anchors `KIPRJMOD` at the project, not at the sheet, so a sub-sheet in
`<proj>/sheets/` still resolves against `<proj>/sym-lib-table`. It's reachable
through Konnect itself — `add_hierarchical_sheet` accepts a `sheet_file` with a
subdirectory in it.

Suggested fix: from the schematic's directory, walk **upwards** looking for a
`.kicad_pro`, and use that directory; fall back to the immediate parent when
none is found, which keeps today's behaviour for a loose schematic (the
"no `.kicad_pro` beside it" rows above still passed, so that fallback matters).

Acceptance test: create a project, register a symbol library at project scope,
`create_schematic` at `<proj>/sheets/child.kicad_sch`, place the symbol there.

Not a blocker for merging as far as I'm concerned — the main path works and this
is a narrower case. Happy to test a follow-up if you push one.

pauliuszaleckas:
Thanks for the thorough pass — six cases with an exact repro made this trivial to locate.

Fixed in b46d090, now on the branch. `KiCadSymbolSource::for_file` walks up from the
schematic's directory to the nearest ancestor holding a `.kicad_pro` and uses that,
falling back to the immediate parent when there is none — your suggestion. It matches on
the extension rather than a project name, since a sheet's filename says nothing about
what the project is called. All five call sites inherit it.

Your acceptance case is now a test, and I checked it fails without the fix, falling
through to the install dirs exactly as you saw. The projectless-schematic fallback is
covered too, since those rows passing is what made it matter.

If you have time, a retest on 10.0.4/Windows would be welcome — especially the
subdirectory case end-to-end over HTTP, which a unit test doesn't reach.

anyn99:
Wow, thanks for the quick fix and answer! Here are all the test cases I ran, with the new commit `b46d090` included.

**Environment:** KiCad 10.0.4 on Windows, MCP server accessed directly over its HTTP transport.

---

### Against original PR #136 — *derive project directory from the schematic path*

| # | Scenario | Result |
|:--|:---------|:-------|
| 1 | Schematic sits directly in the project folder, symbol from the project library | ✅ **PASS** |
| 2 | Real project instead of a synthetic one: library registered via `${KIPRJMOD}`, multi-pin custom symbol | ✅ **PASS** — upward lookup and URI expansion work together |
| 3 | No project anywhere up the tree, standard symbol requested | ✅ **PASS** — global libraries stay reachable |
| 4 | No project anywhere up the tree, project symbol requested | ✅ **PASS** — clean "not found" error, no crash and no placement without a symbol definition |
| 5 | Legacy workaround path: `get_symbol_info` with explicit `project_dir` | ✅ **PASS** — unchanged, not superseded by the new path |
| 6 | Relative schematic path | ✅ **PASS** — comprehensible error (not a Konnect issue: server CWD differs from client CWD) |

---

### Against commit `b46d090` — *search upward for the project file*

| # | Scenario | Result |
|:--|:---------|:-------|
| 7 | Schematic one level down (`<proj>/sheets/child.kicad_sch`) | ✅ **PASS** |
| 8 | Schematic two levels down (`<proj>/sheets/sub/…`) | ✅ **PASS** — the search does not stop after one level |
| 9 | Real sub-sheet created via `add_hierarchical_sheet` (`sheet_file: "sheets/sub.kicad_sch"`), project symbol placed inside it | ✅ **PASS** — behaves like a manually created sheet |
| 10 | Neighbouring-project trap: a sheet under project A requests a symbol that exists only in sibling project B | ✅ **PASS** — correctly rejected, no reach into foreign library tables |
| 11 | Nested projects (`.kicad_pro` in both the parent and the sub folder), symbol of the inner project | ✅ **PASS** — the nearest project file wins |
| 12 | Nested projects, symbol of the parent project requested from the inner sheet | ✅ **PASS** — correctly invisible; the search stops at the first project boundary |
| 13 | Nickname collision (same library nickname in the global and the project table), symbol exists only in the project library | ✅ **PASS** — found |
| 14 | Nickname collision, same symbol name in both libraries, distinguishable by pin count | ✅ **PASS** — the project version is embedded; no silent substitution |
| 15 | Nickname collision, symbol exists only in the global library | ⚪️ Tables are merged (KiCad would shadow the global one) — deliberately classified as a non-issue, not pursued |

---

### Not covered

UNC / network paths, symlinks, differing path capitalisation.

anyn99 (approved):
This PR is **SO** ready to merge if you ask me 😉 More details in comments above.

mixelpixx:
Reviewed in depth and taking this over #172, which fixes the same issue. Merged main into your branch and pushed two grafts from the PRs being closed as superseded, so their authors' work isn't lost. Nothing of yours was changed.

**Why this one wins the sym-lib-table cluster.** #172 reads only the *global* table — it never threads a project dir — so #105's own reproduction, which registers at project scope, still fails there. It also lands a third lib-table parser in the tree rather than removing the duplication that caused #105 in the first place. This PR routes `resolve_symbol_lib_path` through the same `KiCadSymbolSource` the placement path uses, which makes `get_symbol_info` and `add_schematic_component` structurally incapable of disagreeing — that was the actual complaint in the issue thread. Deleting `konnect-core`'s drifted `resolve_lib_symbol` takes the count of resolvers from two to zero. The `SymbolLibrarySource` trait is the right answer to the crate-layering problem, and the blanket impl for `Fn(&str) -> Vec<PathBuf>` making the tests hermetic without `set_var` is a nice touch — as is `KICAD_ENV_LOCK`.

The `STUB_MARKER` change deserves particular credit: it proves the existing `sch_components` suite had been resolving against whatever `Device` library happened to be installed on the developer's machine, passing only because the real `Device:R` also has two pins. That's a fake-passing test suite, and nothing else in the queue caught it.

**Grafted in (see the commit for detail):**

1. **`kicad_user_path_vars()`, from @JYPochez's #172.** Preferences → Configure Paths variables live in `kicad_common.json` and are never process environment variables, so `expand_lib_uri`'s `var_os` lookup could not see them — a `${MY_PARTS}/house.kicad_sym` entry fell through the `kind` match and returned `None`. That's the recommended way to write a portable table, so it's a real gap. Reuses the existing `kicad_config_dir()` rather than adding another copy.
2. **URI normalisation and a lexical fallback, from @anyn99's #163.** `project_relative_uri`'s `absolute_uri()` returned `to_string_lossy()` with backslashes intact, and global scope passed the raw path through untouched — neither survives the project being opened on Linux or macOS. `portable_uri()` also strips the `\?\` verbatim prefix `canonicalize` returns on Windows. The lexical `strip_prefix` fallback is the substantive half: `canonicalize` fails for a path that doesn't exist yet, and registering a library before creating it is normal, so the URI was silently coming out absolute and the project quietly stopped being portable. Canonical comparison is still tried first.

Four new tests cover those; your `an_empty_table_dir_does_not_produce_a_rooted_kiprjmod_uri` now compares against `portable_uri(&file)` since the normalisation applies there too.

Gate is green on the merge: fmt, clippy, 544 tests.

**One open question I'd like your view on, non-blocking.** `candidates()` appends the install-dir candidates after the table entries unconditionally, so `Device:R` still resolves from the stock library even when a project table claims the `Device` nickname. KiCad maps a nickname to exactly one library. Your behaviour is friendlier, but it can mask a mis-registered project table. I'm inclined to leave it as-is and revisit if anyone hits it — say if you disagree.

Merging once CI is green.

Issue #117, opened 2026-07-31
add_via always fails with "could not unpack PCB_VIA from request (AS_BAD_REQUEST)" on v0.2.1 — regression from #99's fix


ncolomer:
## Summary

On v0.2.1, `add_via` fails on every call — including straightforward, valid input — with:

```
KiCAD IPC error: could not unpack PCB_VIA from request (AS_BAD_REQUEST)
```

Per the closing comments on #99, v0.2.1 shipped a rewrite of `add_via` to send a typed `kiapi.board.types.Via` with a proper per-layer `PadStack` through `CreateItems` (replacing the old `ParseAndCreateItemsFromString` stub that silently no-op'd), plus strict per-item status checking so a rejected item "can never report success again." That part is working as intended — it fails loudly instead of lying. But it now seems to fail loudly on *everything*, which suggests the typed `Via`/`PadStack` message being constructed client-side doesn't unpack correctly on KiCad's end, rather than only failing for genuinely-invalid vias.

## Environment

- Konnect v0.2.1 (installed via KiCad PCM)
- KiCad 10.0.5, macOS
- Live IPC session, PCB Editor open, board loaded (same session where `route_trace`, `route_pad_to_pad`, `delete_trace`, `get_nets_list`, `query_traces` all worked correctly against the same board)

## Repro

```
add_via(board="...", net_name="/CALIB", x=122.25, y=64.379999)
→ Error: KiCAD IPC error: could not unpack PCB_VIA from request (AS_BAD_REQUEST)
```

Retried with explicit `drill`/`pad_size` instead of relying on defaults, in case the omission was the issue:

```
add_via(board="...", net_name="/CALIB", x=122.25, y=64.379999, drill=0.4, pad_size=0.8)
→ Error: KiCAD IPC error: could not unpack PCB_VIA from request (AS_BAD_REQUEST)
```

Same error both times, on an otherwise fully responsive IPC session — this wasn't a stale connection or a board-not-loaded issue (confirmed via `open_project` returning `kicad_ui_running: true` immediately before and after).

## Expected

A via is created at the given position on the given net (the actual intent of #59's fix), or — if there's a genuine reason this specific request is invalid — an error that says what's wrong with it, not a generic unpack failure.

## Actual

[...]

Issue #99, opened 2026-07-28
add_via: reports success but no via is created (KiCad 10, live IPC session)


MrDix:
## Environment
- Konnect v0.2.0 (Windows x86_64 release binary, standalone MCP server)
- KiCad 10.0.5, Windows 11, board open in pcbnew, IPC socket configured via `--config`

## Bug
`add_via` returns a success payload, but no via is actually created on the live board. The via is missing from the running editor session and consequently also from the saved `.kicad_pcb`.

## Repro
1. KiCad 10 running with a board open in pcbnew, IPC connected (verified working: `get_nets_list`, `query_traces`, `route_trace` all behave correctly in the same session).
2. Call:

```json
{net_name: VCC_3V3, x: 163.8, y: 96.6, pad_size: 0.6, drill: 0.3}
```

3. Response (success):

```json
{drill: 0.3, net: VCC_3V3, pad_size: 0.6, x: 163.8, y: 96.6}
```

4. `save_project`, then inspect the saved board file: no `(via (at 163.8 96.6) ...)` exists; the total via count is unchanged. The ratsnest still shows the two clusters as unconnected. Repeated twice, same result.

## Suspicion
KiCad 10 changed vias to per-layer padstacks in both the SWIG and IPC APIs (e.g. size/width now needs a layer argument or a padstack object). A via constructed the KiCad-9 way may be rejected or dropped silently by KiCad 10 while the tool still reports success.

For comparison: creating the same via through KiCad 10's own `kipy` (`Via()` with `position`, `diameter`, `drill_diameter`, `net`, then `board.create_items([...])`) works and the via appears immediately in the editor - so the IPC channel itself is fine.

## Expected
Either the via is created, or the tool returns an error. A silent no-op with a success payload is the worst case for automation, since track segments routed on the assumption of a layer change then end up floating.

mixelpixx:
Confirmed, and good news: this is already fixed on `main` — just not in any release yet (you're on the v0.2.0 binary).

[...]

PR #137, opened 2026-08-08
fix(pcb): create the mounting hole in KiCAD, not just in the file

jctrouble:
## Problem

`add_mounting_hole` is a silent no-op against a running KiCAD. Called three times in one session with KiCAD open and IPC live, it returned a success JSON echoing the inputs each time — and **no mounting hole was ever created**: nothing on the canvas, and nothing in the saved `.kicad_pcb` afterwards (grepped: the reference `H1` absent, no ø3.45 drill anywhere in the file).

## Root cause

`handle_add_mounting_hole` had no IPC path at all. It read the `.kicad_pcb`, spliced a `(footprint …)` block in before the closing paren, and wrote the file — unconditionally, whether or not a live KiCAD was holding that board.

The emitted S-expression is fine; that is not the bug. Injecting exactly what `format_npth_footprint` produces into a board file and running `kicad-cli pcb export drill --excellon-separate-th` yields the hole in the `-NPTH.drl`, so KiCAD parses it happily. The bug is *where* it was written. With the board open in Pcbnew, KiCAD's in-memory copy never learns about the edit, and its next save writes that in-memory copy straight over the file — the footprint disappears with no error anywhere. Three calls, three successes reported, zero holes.

This is the same hazard #70 / cc5a995 fixed for `place_component`, and that commit's own comment already spells it out:

> A KiCad that answered — even with an error — may hold this board open, and a file edited behind a live editor is silently overwritten on its next save, so a rejection fails closed with no fallback.

`add_mounting_hole` was simply never brought onto that policy. It is the only footprint-creating tool left on the unconditional file path.

## Fix

A mounting hole is a footprint, so route it through the footprint path that already exists and is known to persist:

- **IPC first** — `KiCadIpcClient::place_footprint` with a single synthesized `np_thru_hole` pad, the same call `place_component` makes. `build_footprint_item` already maps `"np_thru_hole"` → `PadType::PtNpth` and `"*.Cu"`/`"*.Mask"` layer wildcards, so no new IPC plumbing is needed.
- **`IpcFailure::Rejected` → fail closed.** Error out and say the file was not modified, verbatim in the shape `place_component` uses.
- **`IpcFailure::Unreachable` → the old file edit**, unchanged, now reported as `"source": "file"` with the same "KiCAD will show this when it next loads the board" warning `place_component` returns.

The geometry constants (`mounting_hole_lib_id`, `mounting_hole_pad_size`, `mounting_hole_text_offset`, `mounting_hole_pad`) are shared by both paths so an IPC-placed hole and a file-written one cannot drift apart.

`with_ipc_classified` becomes `pub(crate)` so `pcb_board` can reuse it rather than growing a second copy.

## Compatibility

The success JSON keeps `reference` / `x` / `y` / `drill_diameter` and gains `footprint`, `source`, and (on the fallback) `warning` — additive, matching `place_component`'s response shape. The tool schema is unchanged.

One behaviour change worth calling out: with KiCAD reachable but this board *not* open, the tool now errors instead of editing the file. That is the deliberate fail-closed rule from cc5a995 — the same case already errors for `place_component`.

## Changes

- `crates/konnect-core/src/tools/pcb_board.rs` — IPC-first `handle_add_mounting_hole`; shared mounting-hole geometry helpers; new `mounting_hole_tests` module.
- `crates/konnect-core/src/tools/pcb_components.rs` — `with_ipc_classified` visibility only.

## Testing

New tests in `pcb_board::mounting_hole_tests`:

- `a_reachable_kicad_that_rejects_never_touches_the_file` — the regression test. A rep0 mock that answers every round-trip with `AS_BAD_REQUEST` (the helper `place_component`'s fallback test uses). **Before the fix: FAILS** — "a rejection must not be reported as success", and the board file was modified. After: the tool errors and the file is byte-identical.
- `an_unreachable_kicad_still_falls_back_to_the_board_file` — empty `ipc_address`, so the failure classifies `Unreachable`; asserts `source: "file"` and that the `np_thru_hole` pad, the `(drill 3.45)` and the reference land in the file. **Before the fix: FAILS** (no `source` key). After: passes.
- `mounting_hole_pad_is_an_unplated_hole_with_drill_and_annulus` — the pad definition: NPTH, circular, drill 3.45, annulus 3.95 matching the file path's `(size …)`, `*.Cu`/`*.Mask`.

Checklist commands, all clean:

- `cargo test --workspace --locked --lib --tests` — 236 lib tests pass (233 before + 3 new), all integration/conformance suites green
- `cargo test --workspace --locked --doc` — clean

Not run: a live-KiCAD end-to-end placement (no KiCAD session available in this environment). The IPC path is the byte-for-byte same `place_footprint` call `place_component` makes and is covered by its live tests.


mixelpixx:
Merged main into your branch and **live-verified against KiCAD 10.0.5** — for an IPC write that's the only check that means anything, so here's exactly what I ran.

Launched pcbnew on a copy of the `ecc83-pp` demo board, then drove the real server over stdio:

```
add_mounting_hole(board, x=30, y=30, diameter=3.2, reference="H9")
  -> {"source":"ipc", "footprint":"MountingHole:MountingHole_3.2mm", "reference":"H9", ...}

find_component(board, reference="H9")
  -> {"footprint":"MountingHole:MountingHole_3.2mm","layer":"F.Cu","x":30.0,"y":30.0,...}
```

`source: "ipc"` is the important half — it means the request went to a live KiCAD and was *accepted*, not that it quietly fell back to the file path. And the read-back matters independently: "no error" has not been proof that anything was created since #99, so I confirmed the hole is on the running board rather than trusting the return value.

I also re-ran the `#[ignore]`d live suite while KiCAD was up — `adding_a_via_actually_creates_it_on_the_board` and `moving_and_rotating_footprint_preserves_child_geometry` both pass, so nothing in this round regressed the IPC write path.

**On the padstack question**, which was my one real concern before testing: `layers: ["*.Cu", "*.Mask"]` is safe here. The shared pad builder pins `copper_layers` to exactly one `BlFCu` entry — KiCad's `ALL_LAYERS` sentinel — and the wildcard expands into the separate `layers` field, not into `copper_layers`. That's the rule that broke `add_via` in v0.2.1 (#117), and this PR doesn't introduce a new padstack constructor, it reuses `place_component`'s. Structurally safe, and now empirically confirmed.

The merge itself: #142/#153 added `net_count_tests` to `pcb_board.rs` at the same spot as your `mounting_hole_tests` and git interleaved the bodies, so I took main's file, re-applied your production diff, and separated the two modules — each keeps its own context helper, since yours needs `ctx_with_ipc(address)`.

On the review: mirroring `place_component`'s fail-closed discipline exactly — `Rejected` errors with no file write, only `Unreachable` falls back — is right, and `spawn_rejecting_kicad` is a genuine regression test for that gate rather than a smoke test. Extracting `mounting_hole_lib_id` / `_pad_size` / `_text_offset` so the IPC and file paths can't drift is the detail I'd have asked for if it weren't already there.

Merging once CI is green.

PR #141, opened 2026-08-08
fix(library): repair a lib-table whose root declares the wrong kind

jctrouble:
## Problem

A project's `sym-lib-table` can begin with `(fp_lib_table`, and once it does, nothing Konnect offers will fix it. KiCad rejects the whole file:

```
Library table 'sym-lib-table' has type FOOTPRINT but expected SYMBOL; skipping
```

So every symbol library registered in that project is invisible to KiCad, `register_symbol_library` keeps reporting `success: true`, and the only way out is to delete the file by hand.

## Root cause

#54 fixed the **creation** path: `table_root_element` picks the root from the filename, so a *new* `sym-lib-table` is scaffolded as `(sym_lib_table`. But that scaffold is only used when the file does not exist:

```rust
let content = if table_path.exists() {
    tokio::fs::read_to_string(table_path).await?   // whatever is there, root and all
} else {
    format!("({}\n  (version 7)\n)\n", table_root_element(table_path))
};
```

A table an older Konnect build wrote with the hardcoded `fp_lib_table` root is read back verbatim and appended into, so the bad root is carried forward by every subsequent registration. #54 never reaches a project that already tripped over the bug.

There is a second, sharper edge on the idempotent path:

```rust
if content.contains(&format!("(name \"{}\")", nickname)) {
    return Ok(()); // already registered, idempotent
}
```

Re-registering the same nickname — the obvious thing to try when a library will not resolve — returns success without writing anything at all, so even a repair placed above this line would never reach disk.

## Fix

`repair_table_root(content, expected_root)` rewrites a root element of the wrong kind and reports whether it did. `register_in_lib_table` calls it on the existing content, writes the repaired file even when the nickname is already registered, and returns the flag.

The rewrite is deliberately narrow: the wrong token must be the **first** thing in the file (`content[..pos].trim().is_empty()`), so `fp_lib_table` mentioned inside a `(descr …)` is left alone.

`register_symbol_library` and `register_footprint_library` add `"repaired_table_root": true` to their result and a `tracing::warn!` fires, because a project that was silently broken should say so rather than be mutely fixed.

## Compatibility

`register_in_lib_table` returns `anyhow::Result<bool>` instead of `Result<()>`; it is a private helper with two in-tree callers, both updated. The tool results gain one additive boolean field. Correct tables are byte-for-byte untouched — pinned by a test.

## Changes

- `crates/konnect-core/src/tools/library.rs` — `repair_table_root`; `register_in_lib_table` repairs and reports; both register handlers surface `repaired_table_root`; four tests.

## Testing

- `an_existing_sym_table_with_an_fp_root_is_repaired` — a pre-existing `(fp_lib_table`-rooted `sym-lib-table` with one entry; after registering a second library the file starts with `(sym_lib_table`, contains no `fp_lib_table`, and still has both entries. **Fails before the fix.**
- `a_repair_still_lands_when_the_nickname_is_already_registered` — re-registering the nickname that is already listed still repairs the root, and does not duplicate the entry. **Fails before the fix** (the early return wrote nothing).
- `a_correct_root_and_a_mention_inside_a_descr_are_left_alone` — a correct table is returned unchanged, and `fp_lib_table` inside a `(descr …)` of a `sym_lib_table` is not rewritten.
- `registering_a_symbol_library_scaffolds_a_sym_root` — extended to assert the new flag is `false` for a fresh scaffold.

Checklist commands, all clean:

- `cargo test --workspace --locked --lib --tests` — 236 lib tests (233 before + 3 new), all suites green
- `cargo test --workspace --locked --doc` — clean

## Aside, not fixed here

Project-scope registration writes the **absolute** path as the URI. `${KIPRJMOD}/…` would be the portable form and `expand_lib_uri` already resolves it (it is what `register_footprint_library`'s own test exercises), but choosing it unconditionally would break registering a library that lives outside the project directory, so it wants a rule rather than a one-liner. Happy to open a separate issue if that is worth doing.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -1198,7 +1198,7 @@ async fn handle_register_footprint_library(
         ));
     };
 
-    register_in_lib_table(
+    let repaired_root = register_in_lib_table(
         &table_path,
         nickname,
         lib_path.to_str().unwrap_or(""),
@@ -1211,7 +1211,8 @@ async fn handle_register_footprint_library(
             "success": true,
             "nickname": nickname,
             "scope": scope,
-            "table": table_path.to_str().unwrap_or("")
+            "table": table_path.to_str().unwrap_or(""),
+            "repaired_table_root": repaired_root
         }))
         .unwrap(),
     ))
@@ -1281,7 +1282,7 @@ async fn handle_register_symbol_library(
         ));
     };
 
-    register_in_lib_table(
+    let repaired_root = register_in_lib_table(
         &table_path,
         nickname,
         lib_path.to_str().unwrap_or(""),
@@ -1294,7 +1295,8 @@ async fn handle_register_symbol_library(
             "success": true,
             "nickname": nickname,
             "scope": scope,
-            "table": table_path.to_str().unwrap_or("")
+            "table": table_path.to_str().unwrap_or(""),
+            "repaired_table_root": repaired_root
         }))
         .unwrap(),
     ))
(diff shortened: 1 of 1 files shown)


mixelpixx:
Merged main into your branch — #136 landed first and rewrote the same two `register_in_lib_table` call sites to take their URI from `lib_table_target()`, which is what makes a project-scoped registration portable.

Resolved so both survive: the call passes #136's `&uri` **and** keeps your `repaired_root` return, and the symbol registrar now reports `"uri"` alongside `"repaired_table_root"`. The footprint registrar reports only the repair flag, matching its shape on main — it never reported a URI. Saved you the rebase you offered in the #105 thread; thanks for having it ready.

Gate green: fmt, clippy, 552 tests, including both of yours.

Worth recording why this is a real bug and not housekeeping: #54 fixed the *creation* side, so `table_root_element` already exists at `library.rs:1352` — but `register_in_lib_table` had no repair path, so a table an older build already wrote stays broken forever. KiCad rejects the whole file with "has type FOOTPRINT but expected SYMBOL; skipping", and every later registration just appends into the bad root.

The detail I liked most is the one your test pins: the idempotent "already registered" early return used to skip the write entirely, so re-registering the same nickname left the bad root in place — `a_repair_still_lands_when_the_nickname_is_already_registered` covers exactly that. And scoping `repair_table_root` to a *root* element so `(descr "copied from fp_lib_table")` is left alone is the right call, tested.

Issue #133, opened 2026-08-05
KiCad 10 net format: get_component_pads returns net:"" and get_board_info reports net_count 0 (pad net name is at index 1, not 2)


cvaldess:
## Summary

In KiCad 10 the `net` node inside a `pad` no longer carries a net number: it is written as `(net "GND")`, not `(net 1 "GND")`. There is also no top-level net table (`(net 0 "")`, `(net 1 "GND")`, …) in the file at all any more.

Konnect still reads the net name positionally at index 2, so on any board saved by KiCad 10 it silently returns an empty string instead of the net name. Because the fallback is `.unwrap_or("")` rather than an error, a fully connected pad is indistinguishable from an unconnected one.

**Environment:** Konnect 0.2.2 (latest release), KiCad 10.0.4, Windows 10 x64.

## Reproduction

Two minimal boards, identical except for the net syntax.

`repro_k10.kicad_pcb` (KiCad 10 format, `version 20260206`):

```
(kicad_pcb
	(version 20260206)
	(generator "pcbnew")
	(generator_version "10.0")
	(general (thickness 1.6))
	(paper "A4")
	(layers (0 "F.Cu" signal) (31 "B.Cu" signal))
	(footprint "TestPad"
		(layer "F.Cu")
		(at 100 100)
		(property "Reference" "R1" (at 0 0 0) (layer "F.SilkS"))
		(pad "1" smd rect (at -1 0) (size 1 1) (layers "F.Cu" "F.Mask") (net "GND"))
		(pad "2" smd rect (at 1 0) (size 1 1) (layers "F.Cu" "F.Mask") (net "VCC"))
	)
)
```

`repro_k9.kicad_pcb` is the same file with `version 20241229`, a net table, and `(net 1 "GND")` / `(net 2 "VCC")` on the pads.

### `get_component_pads(reference: "R1")`

| board | result |
| --- | --- |
| `repro_k9` | `[{"number":"1","net":"GND",…},{"number":"2","net":"VCC",…}]` ✅ |
| `repro_k10` | `[{"number":"1","net":"",…},{"number":"2","net":"",…}]` ❌ |

### `get_board_info`

| board | `net_count` |
| --- | --- |
| `repro_k9` | `2` ✅ |
| `repro_k10` | `0` ❌ |

Pad **coordinates** are correct in both cases, so this is purely the net lookup.

[...]

Issue #84, opened 2026-07-26
Remaining hardcoded-indentation scans: sch_export, integration, schematic_builder


mixelpixx:
Follow-up to #64/#69. The `rfind("\n  (tag")` / hardcoded two-space-indent class was eliminated from the deletion paths by #69, but the same pattern survives in non-delete paths. On eeschema-saved (tab-indented) schematics these silently no-op or pick wrong targets — lower severity than #64's data loss, but the same root cause:

- `crates/konnect-core/src/tools/sch_export.rs:396` — `rfind("\n  (wire")` in wire-endpoint nudging: silent no-op on tab-indented files.
- `crates/konnect-core/src/tools/integration.rs:631` — `rfind("\n  (symbol")`: silently skips datasheet backfill.
- `crates/konnect-core/src/tools/schematic_builder.rs:96-98, 162, 190` — `"\n  ("` top-level splitting drives insertion points; worth a careful look since it chooses where content lands.

The right fix is the same machinery #40/#69 established: `konnect_sexp::writer::find_block_starts`, `find_enclosing_block`, and (since #69) `find_direct_child_blocks` / `find_enclosing_direct_child_block` — indentation-agnostic, string-aware, and already unit-tested. No new primitives should be needed.

(The two `library.rs` occurrences of this class — `parse_lib_table` at :833 and its duplicate in `search_footprints` — are fixed by #65 and are not part of this issue.)

mixelpixx:
Another confirmed site for this class, found while verifying discussion #239 — `crates/konnect-core/src/tools/library.rs:3580`:

```rust
let pad_count = content.matches("\n  (pad ").count();
```

`get_footprint_info` therefore reports `pad_count: 0` for every real KiCad footprint, because stock library footprints are **tab-indented and CRLF**, not two-space LF. Verified against `C:\KiCad\10.0\share\kicad\footprints\Package_SO.pretty\SOIC-8_3.9x4.9mm_P1.27mm.kicad_mod`.

Filed separately as #248 because `pad_count: 0` is the symptom a user searches for, but it belongs to this issue's list — and it adds a fourth file (`library.rs`) to the three named here (`sch_export`, `integration`, `schematic_builder`).

It also adds a CRLF dimension this issue does not currently mention: `"\n  (pad "` fails on tab indentation *and* on `\r\n`, independently. Any sweep for this class should test both.

neusse:
[...]

Issue #58, opened 2026-07-23
add_via returns success but never creates the via (silently dropped)


jctrouble:
### Summary

The `add_via` MCP tool returns a **success** response echoing the drill/net/pad_size/x/y, but **no via is ever created** — nothing appears in the live KiCAD session and nothing persists after save.

### Environment

- KiCAD 10.0.4, macOS
- Konnect built from `main` (a782dd1), plugin installed via KiCAD PCM
- IPC at `ipc:///tmp/kicad/api.sock`
- Observed 2026-07-23

### Reproduction

1. Open a board in KiCAD with the IPC API enabled.
2. Call `add_via` (e.g. `net_name=VCC_BATT`, `x=146.268`, `y=89.194`, `pad_size=0.45`, `drill=0.2`).
3. Observe the tool returns success JSON: `{"net":"VCC_BATT","x":146.268,"y":89.194,"drill":0.2,"pad_size":0.45}`.
4. Save the board and inspect the `.kicad_pcb` (or the live canvas).

### Expected

A through via at (146.268, 89.194) on net `VCC_BATT` exists on the board and persists after save.

### Actual

Zero via objects on the board — nothing in the live session, and a regex over the saved `.kicad_pcb` finds no `(via …)`. The tool nonetheless reported success (the JSON is fabricated from the input args, not from the IPC result).

For contrast, `route_trace` called against the **same socket in the same session** persists its track segment correctly, so the connection and net resolution are fine.

### Root cause (server is healthy — the bug is client-side)

Creating the **same via** through the official `kicad-python` (kipy) client's `create_items()` against the very same socket persists correctly, which proves the IPC server side is healthy. The defect is in Konnect's `add_via` implementation:

`KiCadIpcClient::add_via` (`crates/konnect-ipc/src/client.rs`) does **not** use the protobuf `create_items` path that `add_track` uses. Instead it builds a bare `(via …)` S-expression string (`builders::via_sexp`) and sends it via `ParseAndCreateItemsFromString`. Two problems compound:

1. `ParseAndCreateItemsFromString` mimics an editor **paste** and returns a `CreateItemsResponse` whose overall status, per the proto comment, "may return IRS_OK even if no items were created." The paste of a bare, un-wrapped `(via …)` token (with a pad-style `(net code "name")` node that a via/segment does not take) parses to zero items.
[...]

PR #142, opened 2026-08-09
fix(pcb): read KiCad 10 net nodes by shape, not by fixed index

jctrouble:
Fixes the three sites in #133 (`get_component_pads`, `get_board_info`, `add_net`). Follows #140, which fixed a fourth site with the same root cause in `manufacturing.rs`.

## Problem

KiCad 10 changed the net node in two ways at once: a pad's net is written `(net "GND")` rather than `(net 1 "GND")`, and the top-level net table is gone entirely — a net exists only by being named on the items that carry it. Konnect read the name positionally at index 2, and the miss went through `.unwrap_or("")`, so on any board saved by KiCad 10:

| call | result on a KiCad 10 board |
| --- | --- |
| `get_component_pads` | `net: ""` for **every** pad — a connected pad is indistinguishable from an unconnected one |
| `get_board_info` | `net_count: 0` — `find_all("net")` sees only direct children of `(kicad_pcb …)`, and there is no table to see |
| `add_net` | next id = `content.matches("(net ").count()`, i.e. every reference on every pad, segment and zone; and it emits the legacy two-element node into a file whose format has no table for it |

@cvaldess's correlation datum from the issue is what makes this safe to fix by shape rather than by version: across ~40 `.kicad_pcb` files, `8.0/9.0/9.99 → two-space indent and (net N "x")`, `10.0 → tabs and (net "x")` — the two changes travel together, so a probe that assumes one is also wrong about the other. That is why nothing here sniffs indentation and nothing infers the net shape from the format version.

## Fix

New `konnect_sexp::net` reads by **shape**: whether child 1 is a quoted string or a bare atom is what distinguishes the formats, so both read correctly with no version check, and a file that mixes them still works.

- `net_name` / `net_id` — one node, either shape.
- `collect_net_keys` / `count_distinct_nets` — walks the tree (the nets are wherever the pads, segments, vias and zones are) and keys on **names when any are present, numeric ids otherwise**. That is the same rule #140 used, and it is load-bearing in both directions: keying on ids drops KiCad 10 entirely, and keying on names would drop a KiCad 9 file's bare `(net 1)` references. De-duplication is not a nicety — the 4-layer board in #133 has 872 `(net …)` occurrences for 65 distinct nets, and an 8-layer one measured here has 5423 for 776.

The pseudo-net rule also matches #140: the empty name and id `0` are excluded, so an empty board does not look wired.

## Verification

Against a real 8-layer KiCad 10 board (format `20260206`), through the MCP server over stdio:

- `get_board_info` → `net_count: 776`, cross-checked against an independent quote-aware scan of the same file that also reports 776 from 5423 occurrences. (A naive `\(net[^)]*\)` regex says 571 — KiCad's auto-generated names contain parentheses, e.g. `unconnected-(U59-PGOOD-Pad6)`, so the cross-check had to tokenize. Worth knowing for anyone else counting nets from these files.)
- `get_component_pads` on a 30-pad power part → real net names (`VBUS_SLOT_FUSED`, `GND`) where every pad previously read `""`; its four mechanical pads, which carry no `(net …)` node at all, still read `""`.
- `add_net` → refused, and the board file is byte-identical afterwards (sha256 unchanged).

Gate green: `cargo fmt --all --check`, `cargo clippy --workspace --locked -D warnings`, `cargo test --workspace --locked --lib --tests` and `--doc`. 20 new tests cover the KiCad 8/9 shape, the KiCad 10 shape, mixed files, a name that looks like a number, the pseudo-net, and the unreadable case.

## Two decisions I would like a ruling on

Both are judgement calls; I have implemented what I believe is right and will amend either without argument if you rule the other way.

**1. `get_component_pads` returns `null`, not `""`, for a net node it cannot read.**

Three outcomes are now distinguishable where there were two: no `(net …)` node at all is an unconnected pad and stays `""`; a readable node gives its name; a node that is *present but unreadable* gives `null`.

*Chosen because* `""` is a clean-looking answer that means nothing, which is the exact failure mode of this bug — a caller sanity-checking connectivity before fab ("is every pad on a net?") got a tidy, wrong answer on every KiCad 10 board. `null` makes the caller notice.

*The alternative* is to keep `""` for both and preserve the current contract exactly. This **is** a behaviour change for existing callers, and it is the reason I am flagging it: anyone treating `net` as a `string` will now see `null`. If you would rather not break that, say so and I will fold the unreadable case back into `""` — or, if you prefer a third way, emit `""` plus a separate boolean field.

**2. `add_net` fails closed on a KiCad 10 board rather than writing a matching-shape node.**

*Chosen because* KiCad 10 has no top-level net table, so there is nothing for a file-level insert to add. Writing `(net "NAME")` at top level would report success while KiCad discarded it on load — the same "reports success, does nothing" pattern as #58/#99 (`add_via`), which is what this class of bug keeps producing. The error names the format, says why, and points at the tools that create a net by naming it on copper (`route_trace` / `add_via` / `add_copper_pour`), plus `get_nets_list` for reading live nets over IPC.

*The alternative* — the one I started with — is to emit the shape matching the file, name-only in a KiCad 10 file and legacy in a legacy file, on a least-surprise argument for mixed-version installs. I moved off it once I confirmed the node has nowhere to live in KiCad 10. If you would rather `add_net` stay total and write the matching shape, that variant is a small change and I will send it.

Legacy boards are unaffected except for one repair: the next id is now one past the highest in use, instead of a count of `"(net "` occurrences, which collided with existing ids on any real board.

Format detection for that refusal is structural first — a single name-only node anywhere is conclusive, since no earlier format ever wrote that shape — and falls back to the format version only for a board that names no net at all. The threshold (`>= 20260101`) sits above every 9.x format including the 9.99 development builds that still wrote the legacy shape, per the correlation above; there is a test for that case.

## Notes

- **Sequencing with #140:** no textual overlap — that PR touches only `manufacturing.rs`, this one touches `konnect-sexp` and the three `pcb_*` tools — so they merge cleanly in either order. There is a semantic overlap: #140 carries a private `count_nets_and_tracks` with the same name/id rule. Whichever lands second, I will follow up with a commit pointing `manufacturing.rs` at `konnect_sexp::net` so there is one copy of the rule.
- **#84 (remaining hardcoded-indentation scans):** nothing here scans indentation, by design. Per @cvaldess's note on #133, `manufacturing.rs` is effectively a fourth file for that issue and #140 closes that slice; worth linking so the sweep does not re-land on it.
- **Adjacent, deliberately not touched:** `find_net_id` (in both `pcb_routing.rs` and `pcb_board.rs`) has the same legacy assumption on the *write* side, used by zone and pour creation; and `get_board_info` reports `layer_count: 0` on the 8-layer board above, from `find_all("")`, which is an unrelated silent zero of the same family. Both are outside the scope of #133 and I would rather not widen this PR without a nod — happy to take either as a follow-up.
- @cvaldess offered to keep testing against real KiCad 10 boards once a fix was up. This is up.

### DIFF
--- a/crates/konnect-core/src/tools/pcb_board.rs
+++ b/crates/konnect-core/src/tools/pcb_board.rs
@@ -161,7 +161,8 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "get_board_info",
-            "Return metadata about the PCB: title, revision, company, layer count, paper size.",
+            "Return metadata about the PCB: title, revision, company, layer count, paper size, \
+             and the number of distinct nets (excluding the unconnected pseudo-net).",
             json!({
                 "type": "object",
                 "properties": {
@@ -422,7 +423,11 @@ async fn handle_get_board_info(
         .unwrap_or("A4")
         .to_string();
 
-    let net_count = tree.find_all("net").len().saturating_sub(1); // exclude net 0
+    // Not find_all("net"): that counts only direct children of (kicad_pcb …),
+    // i.e. the top-level net table — which KiCad 10 does not write at all, so
+    // every KiCad 10 board reported 0. Collect from wherever the nets actually
+    // are and de-duplicate; see konnect_sexp::net.
+    let net_count = konnect_sexp::net::count_distinct_nets(&tree);
 
     Ok(CallToolResult::json(&json!({
         "file": board_path.display().to_string(),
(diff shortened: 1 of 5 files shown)


cvaldess:
Tested this on Windows against a real KiCad 10 board — the same one the numbers in #133 came from. It fixes both net-facing sites, and it surfaces that `layer_count` is a genuinely separate bug.

**Board under test:** `version 20260206`, `generator_version "10.0"`, tab-indented. 872 `(net "…")` occurrences, **0** in the legacy shape, **65** distinct nets. 27 declared layers, of which **2** are copper.

**Method:** same file, same call, the binary is the only variable — installed v0.2.2 vs a local `cargo build` of this PR (`stable-x86_64-pc-windows-msvc`).

### `get_component_pads`

`U3`, a DFN-5 whose net assignment I can state independently because I routed it:

| pad | v0.2.2 | this PR | file |
|---|---|---|---|
| 1 | `""` | `/SDA` | `/SDA` |
| 2 | `""` | `/SCL` | `/SCL` |
| 3 | `""` | `/+3.3V` | `/+3.3V` |
| 4 | `""` | `GND` | `GND` |
| 5 (EP) | `""` | `GND` | `GND` |

All five correct. Worth stating explicitly: **the coordinates are byte-identical between the two builds**. That is the shape of this bug — the geometry was always right, so every caller that checked "did the call succeed and are the pads where I expect?" got a clean-looking answer with the connectivity silently blanked. The loud `null` for an unreadable node is the right call for exactly that reason; `""` gave a wrong answer that looked like a fact.

### `get_board_info`

| | v0.2.2 | this PR | file |
|---|---|---|---|
| `net_count` | `0` | **`65`** | 65 |
| `layer_count` | `0` | `0` | 2 copper / 27 declared |

`net_count` lands on the exact figure, so the de-duplicating collector holds on a real board and not just on synthetic fixtures — 872 occurrences → 65 nets, and the unconnected pseudo-net is excluded consistently with #140.

**`layer_count` is still `0` with this PR applied.** That confirms it isn't a net-shape symptom: it fails the same way on KiCad 9 files, and it isn't in #133's list. Flagging it so it doesn't get assumed closed once this merges — happy to take it as a separate PR if nobody is on it.

### Test suite

`cargo test --workspace` on windows-msvc: **498 passed, 0 failed, 7 ignored** across 24 suites. The `net.rs` cases cover the part that actually needs covering — `a_name_that_looks_like_a_number_is_still_a_name` is the one I'd have gotten wrong, since a KiCad 10 net legitimately named `"5"` is indistinguishable from a legacy id by shape alone if you only check whether the token parses as a number.

### Building on Windows — three undocumented dependencies

Not a review point, but it cost me two failed builds and it will cost the next Windows contributor the same:

1. **MSVC toolset** — expected, and `stable-x86_64-pc-windows-msvc` is the default.
2. **CMake**, for `nng-sys`. Fails with `Could not find "cmake"`. Worth knowing: **VS Build Tools 2022 already ships one** at `Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe` (3.31.6 here), it just isn't on `PATH` — so nobody needs a separate install, only the hint.
3. **`protoc`**, for `konnect-ipc`. Fails with `Could not find 'protoc'`. Not bundled with anything; `winget install Google.Protobuf` works.

A short "Building from source" note covering 2 and 3 would remove the whole class of first-contribution friction. I can send that as a docs PR if it's wanted.

I have a couple more KiCad 10 boards here if you want another fixture for the de-dup path at a different scale.

Issue #62, opened 2026-07-24
get_symbol_info reports 0 pins for derived (extends) symbols


P4uLT:
## Problem

`get_symbol_info` returns `pin_count: 0` for any **derived** symbol — one declared with `(extends "BASE")`. Derived symbols inherit their pins from the base and define no `pin` nodes of their own, so walking only the symbol's own subtree finds nothing.

This was flagged as a known limitation in #51, which left this comment in `crates/konnect-core/src/tools/library.rs`:

> derived symbols (`(extends …)`) inherit pins from their base and may report 0 pins here — resolving the extends chain is not yet supported.

## Impact

The pinout of derived symbols can't be introspected via MCP — common for parts built on a shared base (opamps, connector/logic variants). The symbol is otherwise readable; only the pin list comes back empty.

## Expected

Follow the `extends` chain: collect pins from the symbol itself, then each base in turn, most-derived precedence on pin-number conflicts, with a cycle guard and graceful stop if a base is missing (mirrors `kicad-tools`' `resolve_extends`).

## Context

- Konnect [`main` @ `88ccb9d`](https://github.com/mixelpixx/Konnect/commit/88ccb9d9c2c2823fecd91b1a7116000053c123b4)
- [`crates/konnect-core/src/tools/library.rs` — `handle_get_symbol_info`, the `extends` NOTE (~L1781)](https://github.com/mixelpixx/Konnect/blob/88ccb9d9c2c2823fecd91b1a7116000053c123b4/crates/konnect-core/src/tools/library.rs#L1781)

PR #122, opened 2026-08-01
feat(schematic): add conflict-safe transaction layer

perara:
## Summary

This is PR 1 of the replacement roadmap for #85. It reconstructs only the schematic command/transaction foundation on the latest `upstream/main`; the three maintenance tools and Schematic Studio viewer feature/source are intentionally excluded. The existing standalone viewer lockfile is regenerated only so its current path dependency on `konnect-sexp` remains buildable with `--locked`.

## Maintainer-blocker mapping

- [x] Split the original 22k-line contribution into an independently reviewable foundation PR
- [x] Reconciled by hand with current `sch_components.rs`, `sch_wiring.rs`, `sch_analysis.rs`, `library.rs`, and `sch_batch.rs`
- [x] Preserved #35 multi-unit validation, replacement, pin extraction, wiring, and flattened library inheritance
- [x] Preserved #104 pin-mid-wire/junction semantics and current analysis call sites
- [x] Preserved #111 batch placement/connect behavior and junction deduplication
- [x] Added UUID-targeted commands, exact revision preconditions, inverses, and atomic multi-file transactions
- [x] Replaced `fs2` completely with `fs4`
- [x] Moved cooperative locks out of projects to `KONNECT_STATE_DIR/locks` or the platform local-data `konnect/locks`
- [x] Lock identities use canonical parents plus length-prefixed native filename data and support new/non-Unicode targets
- [x] Rejects symlinked lock artifacts/state directories and keeps project files untouched
- [x] Added redacted `transaction status`, safe `transaction recover`, and force-gated `transaction abandon`
- [x] Validates journal version, ID/filename agreement, containment, duplicates, malformed paths/content, and symlinks
- [x] Active/abandoned journals are ignored and documented as sensitive full before/after images
- [x] `Schematic::save(new_path)` is create-only; loaded-path saves and `overwrite()` are revision-checked
- [x] Existing schematic mutation callers were migrated without weakening replacement semantics
- [x] No maintenance tools, viewer feature/source, screenshots, registry count, or tool-directory changes

## Validation

- [x] `cargo fmt --all -- --check`
- [x] `cargo check --workspace --locked`
- [x] `cargo test --workspace --locked --lib --tests`
- [x] `cargo test --workspace --locked --doc`
- [x] `cargo clippy --workspace --locked -- -D warnings`
- [x] Standalone viewer `cargo check` and `cargo test` with its own lockfile
- [x] Repository metadata validation
- [x] Linux PCM assembly/structure validation
[...]

PR #144, opened 2026-08-10
fix(schematic): preserve (lib_name ...) and resolve lib_symbols like KiCad

### LINKED ISSUE #143: Symbol round-trip drops (lib_name ...), and lib_symbols resolution ignores it — schematics with sheet-local derived symbols get silently rewired
## Summary

`(lib_name ...)` is absent from the entire codebase — `grep -rn lib_name crates/ --include='*.rs'` returns **zero hits**. It is never parsed, never preserved, never emitted, and never consulted when resolving `lib_symbols`. That produces two defects that compound into silent netlist corruption:

1. **Write side — data loss.** Any tool that goes through the typed model (`Schematic::load` → mutate → `save()`/`to_source()`) re-serializes **every** symbol in the file and strips `(lib_name ...)` and `(exclude_from_sim ...)` from each one. A single `add_schematic_component` is enough to damage symbols it never touched.
2. **Read side — wrong or missing resolution.** All `lib_symbols` lookups match the instance's `lib_id` verbatim. KiCad's rule is `lib_name` when present, otherwise `lib_id`. So on an **untouched, eeschema-authored** file Konnect either returns silently wrong pin geometry, or falsely reports the symbol as having no embedded definition.

Neither KiCad nor Konnect flags the resulting file. The damage only appears in the netlist.

`konnect-schematic-editor`'s own package description is *"Typed KiCAD schematic parser and manipulator with lossless round-trips"* — this is the gap in that claim.

**Versions:** Konnect `main` @ `5cd6454` (v0.2.2), KiCad 9.0 / file format `(version 20250114)`, macOS arm64.

---

## Background: what `lib_name` means

When a symbol's library definition is edited locally, eeschema stores the edited copy in the sheet's `lib_symbols` under a *derived* name and points the instance at it:

```lisp
(lib_symbols
  (symbol "Device:R" ...)   ; stock definition, pins at ±3.81
  (symbol "R_1"      ...)   ; sheet-local edit,  pins at ±6.35
)
...
(symbol
  (lib_name "R_1")          ; <- the lib_symbols entry that actually applies
  (lib_id "Device:R")       ; <- provenance only
  ...
```

`lib_name` is written **before** `lib_id`. This is not exotic: a single real project of ours has **25** such instances across 5 sheets, and on 2 of those sheets the base `Device:R` / `Device:C` entry is not present in `lib_symbols` at all — only the derived one.

---

## Reproducer

`libname_repro.kicad_sch` (148 lines, full source in the collapsed block at the bottom). Three symbols:

| ref | `lib_id` | `lib_name` | base in `lib_symbols`? | pin pitch |
|---|---|---|---|---|
| R1 | `Device:R` | — | yes | ±3.81 |
| R2 | `Device:R` | `R_1` | yes | ±6.35 |
| C1 | `Device:C` | `C_1` | **no** | ±3.81 |

Wires and labels are placed on the *correct* (`lib_name`-resolved) pin coordinates. R1's `NET_TOP`/`NET_BOT` wires end exactly where R2's pins would land **if** `lib_name` were ignored.

Baseline from KiCad itself — six nets, one pin each:

```console
$ kicad-cli sch export netlist --format kicadsexpr -o before.net libname_repro.kicad_sch
  /NET_BOT                     ['R1.2']
  /NET_CAP_BOT                 ['C1.2']
  /NET_CAP_TOP                 ['C1.1']
  /NET_DERIVED_BOT             ['R2.2']
  /NET_DERIVED_TOP             ['R2.1']
  /NET_TOP                     ['R1.1']
```

### Defect 1 — the round-trip corrupts the file

Minimal harness, the exact code path every write tool takes:

```rust
let sch = Schematic::load(&args[1])?;
std::fs::write(&args[2], sch.to_source())?;
```

```console
$ grep -c lib_name libname_repro.kicad_sch after.kicad_sch
libname_repro.kicad_sch:2
after.kicad_sch:0
```

R2's block afterwards — `lib_name` and `exclude_from_sim` are gone:

```lisp
(symbol
  (lib_id "Device:R")
  (at 88.9 63.5 0)
  (unit 1)
  (in_bom yes)
  (on_board yes)
  (dnp no)
  (uuid "44444444-0002-4111-8111-111111111111")
  ...
```

The netlist KiCad derives from the rewritten file:

```console
$ kicad-cli sch export netlist --format kicadsexpr -o after.net after.kicad_sch
  /NET_BOT                     ['R1.2', 'R2.2']     <- R2 shorted to R1
  /NET_TOP                     ['R1.1', 'R2.1']     <- R2 shorted to R1
  Net-(C1-Pad??)               ['C1.1', 'C1.2']     <- C1's two pins shorted together
```

Six nets became three. R2 moved off `NET_DERIVED_TOP`/`NET_DERIVED_BOT` and shorted onto R1's nets, because its pins are now computed from `Device:R`'s ±3.81 pitch instead of `R_1`'s ±6.35. C1 lost its definition entirely — `Device:C` is not in `lib_symbols` — so KiCad collapses both pads into one unnamed net. KiCad opens the file without a single warning.

### Defect 2 — resolution is wrong on files Konnect never wrote

Against the **original, untouched** reproducer:

```console
$ batch_get_schematic_pin_locations(schematic=libname_repro.kicad_sch, references=[R1,R2,C1])
R1 -> [('1', 63.5, 59.69), ('2', 63.5, 67.31)]
R2 -> [('1', 88.9, 59.69), ('2', 88.9, 67.31)]
C1 -> ERROR: no embedded definition for 'Device:C' in lib_symbols — likely added with a nonexistent lib_id
```

- **R2 is silently wrong.** Correct answer is `57.15` / `69.85` (from `R_1`); Konnect returns `Device:R`'s geometry. No error, no warning. Every downstream consumer — wiring, net graphs, `design_review` — inherits the wrong coordinates.
- **C1 produces a false error.** The definition is present, under `C_1`. The advice in the message ("likely added with a nonexistent lib_id … re-add the component") is the opposite of what a user should do, and following it would destroy the local edit.

This is the same class as #62 (derived-symbol pins) but a different mechanism — #62 is about the `(extends ...)` chain inside a library; this is about the sheet-local `lib_name` indirection.

---

### Write side

`crates/konnect-schematic-editor/src/schematic/symbol.rs:80` — `Symbol::from_sexp` keeps an allow-list of child nodes and discards everything else:

```rust
const PRESERVE: &[&str] = &["pin", "instances"];
let raw_sub_nodes = node
    .args()
    .iter()
    .filter(|n| n.tag().map(|t| PRESERVE.contains(&t)).unwrap_or(false))
    .cloned()
    .collect();
```

`Symbol::to_sexp` (`symbol.rs:103`) then rebuilds the block from the modelled fields plus `raw_sub_nodes`. Anything neither modelled nor allow-listed is dropped: `lib_name`, `exclude_from_sim`, `convert`/`body_style`, `default_instance`.

Note this is per-symbol only — unmodelled *top-level* nodes are correctly preserved via `Schematic::raw_other` (`schematic/mod.rs:435`), which is why `lib_symbols` itself survives and the corruption is so easy to miss.

`Sheet::from_sexp` (`schematic/sheet.rs:251`, `PRESERVE = ["stroke", "fill"]`) has the same shape. `label.rs` / `wire.rs` / `misc.rs` have no raw preservation at all.

The lossy path reaches production through `Schematic::save()` / `to_source()` (`schematic/mod.rs:106`, `:133`), used throughout `crates/konnect-core/src/tools/sch_hierarchy.rs` and by `add_schematic_component` (`sch_components.rs:411` → `sch.add_symbol(...)`).

### Read side

`SymbolInstance` (`crates/konnect-sexp/src/schematic.rs:205`) has no `lib_name` field; `extract_symbol_instances` (`:235`) reads only `lib_id`. Every call site then duplicates the same lookup:

```rust
lib_syms.iter().find(|n| n.get(1).and_then(|c| c.as_str()) == Some(&inst.lib_id))
```

- `crates/konnect-core/src/tools/sch_components.rs:807`, `:895`
- `crates/konnect-core/src/tools/sch_analysis.rs:440`, `:485`, `:517`, `:689`
- `crates/konnect-core/src/tools/design_review.rs:145`, `:251`, `:349`
- `crates/konnect-core/src/tools/sch_export.rs:200`, `:330`
- `crates/konnect-core/src/tools/sch_batch.rs:346`, `:503`, `:1043`, `:1142`
- `crates/konnect-core/src/tools/mod.rs:342`, `:381`

The `design_review` and `sch_batch` sites skip silently on a miss, so audits pass while reporting on nothing.

Related: `ensure_lib_symbol`'s `already_present` check (`konnect-schematic-editor/src/library.rs:250-262`) tests `format!("\"{}\"", lib_id)` against a `{:?}` debug rendering of the node — a substring match on debug output, also `lib_name`-blind.

---

## Suggested fix

1. Model `lib_name` explicitly on both `Symbol` and `SymbolInstance`, and emit it **before** `lib_id` to match eeschema.
2. Replace the per-symbol `PRESERVE` allow-list with a deny-list of the tags `to_sexp` already reconstructs, so unmodelled tokens round-trip verbatim (same for `Sheet`).
3. Add one shared resolver — `lib_name` when present, else `lib_id` — and use it at all ~17 sites above. Make the "no embedded definition" errors name the identifier actually looked up.

## Why CI didn't catch it

`crates/konnect-core/tests/conformance_test.rs` is the golden-file suite that would have caught this, but it **skips silently when KiCad isn't installed**, and the bundled demo corpus it runs against contains no `lib_name`. Neither fixture in `crates/konnect-core/tests/fixtures/` has the derived shape either. A fixture with a `lib_name` instance would make this a permanently guarded regression.

Related but distinct: #55 (other `.kicad_sch` output bugs), #62 (derived `(extends ...)` pin resolution), #84 (remaining hardcoded-indentation scans).

Minor aside: `CONTRIBUTING.md` lists `protoc` as the only native prerequisite, but `cargo build --workspace` also needs `cmake` (for `nng-sys`, via `konnect-ipc`).

I have a patch for all three points and will open a PR against this issue.

<details>
<summary><code>libname_repro.kicad_sch</code> (148 lines)</summary>

```lisp
(kicad_sch
	(version 20250114)
	(generator "eeschema")
	(generator_version "9.0")
	(uuid "11111111-1111-4111-8111-111111111111")
	(paper "A4")
	(lib_symbols
		(symbol "Device:R"
			(pin_numbers (hide yes))
			(pin_names (offset 0))
			(exclude_from_sim no)
			(in_bom yes)
			(on_board yes)
			(property "Reference" "R" (at 2.032 0 90))
			(property "Value" "R" (at 0 0 90))
			(symbol "R_0_1"
				(rectangle (start -1.016 -2.54) (end 1.016 2.54)
					(stroke (width 0.254) (type default))
					(fill (type none))
				)
			)
			(symbol "R_1_1"
				(pin passive line (at 0 3.81 270) (length 1.27) (name "~") (number "1"))
				(pin passive line (at 0 -3.81 90) (length 1.27) (name "~") (number "2"))
			)
			(embedded_fonts no)
		)
		(symbol "R_1"
			(pin_numbers (hide yes))
			(pin_names (offset 0))
			(exclude_from_sim no)
			(in_bom yes)
			(on_board yes)
			(property "Reference" "R" (at 2.032 0 90))
			(property "Value" "R" (at 0 0 90))
			(symbol "R_1_0_1"
				(rectangle (start -1.016 -2.54) (end 1.016 2.54)
					(stroke (width 0.254) (type default))
					(fill (type none))
				)
			)
			(symbol "R_1_1_1"
				(pin passive line (at 0 6.35 270) (length 1.27) (name "~") (number "1"))
				(pin passive line (at 0 -6.35 90) (length 1.27) (name "~") (number "2"))
			)
			(embedded_fonts no)
		)
		(symbol "C_1"
			(pin_numbers (hide yes))
			(pin_names (offset 0.254))
			(exclude_from_sim no)
			(in_bom yes)
			(on_board yes)
			(property "Reference" "C" (at 0.635 2.54 0))
			(property "Value" "C" (at 0.635 -2.54 0))
			(symbol "C_1_0_1"
				(polyline (pts (xy -2.032 -0.762) (xy 2.032 -0.762))
					(stroke (width 0.508) (type default))
					(fill (type none))
				)
				(polyline (pts (xy -2.032 0.762) (xy 2.032 0.762))
					(stroke (width 0.508) (type default))
					(fill (type none))
				)
			)
			(symbol "C_1_1_1"
				(pin passive line (at 0 3.81 270) (length 3.048) (name "~") (number "1"))
				(pin passive line (at 0 -3.81 90) (length 3.048) (name "~") (number "2"))
			)
			(embedded_fonts no)
		)
	)
	(wire (pts (xy 63.5 59.69) (xy 88.9 59.69)) (stroke (width 0) (type default)) (uuid "22222222-0001-4111-8111-111111111111"))
	(wire (pts (xy 63.5 67.31) (xy 88.9 67.31)) (stroke (width 0) (type default)) (uuid "22222222-0002-4111-8111-111111111111"))
	(wire (pts (xy 88.9 57.15) (xy 114.3 57.15)) (stroke (width 0) (type default)) (uuid "22222222-0003-4111-8111-111111111111"))
	(wire (pts (xy 88.9 69.85) (xy 114.3 69.85)) (stroke (width 0) (type default)) (uuid "22222222-0004-4111-8111-111111111111"))
	(wire (pts (xy 139.7 59.69) (xy 152.4 59.69)) (stroke (width 0) (type default)) (uuid "22222222-0005-4111-8111-111111111111"))
	(wire (pts (xy 139.7 67.31) (xy 152.4 67.31)) (stroke (width 0) (type default)) (uuid "22222222-0006-4111-8111-111111111111"))
	(label "NET_TOP" (at 76.2 59.69 0) (uuid "33333333-0001-4111-8111-111111111111"))
	(label "NET_BOT" (at 76.2 67.31 0) (uuid "33333333-0002-4111-8111-111111111111"))
	(label "NET_DERIVED_TOP" (at 114.3 57.15 0) (uuid "33333333-0003-4111-8111-111111111111"))
	(label "NET_DERIVED_BOT" (at 114.3 69.85 0) (uuid "33333333-0004-4111-8111-111111111111"))
	(label "NET_CAP_TOP" (at 152.4 59.69 0) (uuid "33333333-0005-4111-8111-111111111111"))
	(label "NET_CAP_BOT" (at 152.4 67.31 0) (uuid "33333333-0006-4111-8111-111111111111"))
	(symbol
		(lib_id "Device:R")
		(at 63.5 63.5 0)
		(unit 1)
		(exclude_from_sim no)
		(in_bom yes)
		(on_board yes)
		(dnp no)
		(uuid "44444444-0001-4111-8111-111111111111")
		(property "Reference" "R1" (at 66.04 62.23 0))
		(property "Value" "10k" (at 66.04 64.77 0))
		(pin "1" (uuid "55555555-0001-4111-8111-111111111111"))
		(pin "2" (uuid "55555555-0002-4111-8111-111111111111"))
		(instances
			(project "min"
				(path "/11111111-1111-4111-8111-111111111111" (reference "R1") (unit 1))
			)
		)
	)
	(symbol
		(lib_name "R_1")
		(lib_id "Device:R")
		(at 88.9 63.5 0)
		(unit 1)
		(exclude_from_sim no)
		(in_bom yes)
		(on_board yes)
		(dnp no)
		(uuid "44444444-0002-4111-8111-111111111111")
		(property "Reference" "R2" (at 91.44 62.23 0))
		(property "Value" "22k" (at 91.44 64.77 0))
		(pin "1" (uuid "55555555-0003-4111-8111-111111111111"))
		(pin "2" (uuid "55555555-0004-4111-8111-111111111111"))
		(instances
			(project "min"
				(path "/11111111-1111-4111-8111-111111111111" (reference "R2") (unit 1))
			)
		)
	)
	(symbol
		(lib_name "C_1")
		(lib_id "Device:C")
		(at 139.7 63.5 0)
		(unit 1)
		(exclude_from_sim no)
		(in_bom yes)
		(on_board yes)
		(dnp no)
		(uuid "44444444-0003-4111-8111-111111111111")
		(property "Reference" "C1" (at 142.24 62.23 0))
		(property "Value" "100n" (at 142.24 64.77 0))
		(pin "1" (uuid "55555555-0005-4111-8111-111111111111"))
		(pin "2" (uuid "55555555-0006-4111-8111-111111111111"))
		(instances
			(project "min"
				(path "/11111111-1111-4111-8111-111111111111" (reference "C1") (unit 1))
			)
		)
	)
	(sheet_instances
		(path "/" (page "1"))
	)
	(embedded_fonts no)
)
```

</details>

jtperry:
DISCLOSURE: Claude AI used heavily in the generation of this PR.

Fixes #143.

REVIEW: this branch was independently reviewed by two separate models — kimi-k3
and gpt-5.6-luna — and both approved it as ready to submit. Their findings were
non-blocking: two wording corrections (now applied to this description) and two
suggested tests (now added as the second commit). Details in §5 and §7.

Two commits. The first is the fix; the second adds the test coverage those
reviews asked for, and carries no production code changes.

## 1. The problem

`(lib_name ...)` did not exist anywhere in the codebase — `grep -rn lib_name crates/ --include='*.rs'` returned zero hits. It was never parsed, preserved, emitted, or consulted when resolving `lib_symbols`. Two defects followed, and together they rewire schematics silently.

KiCad writes `lib_name` whenever a symbol's library definition has been edited inside the sheet. The edited copy is stored in `lib_symbols` under a derived name, and the instance points at it:

```lisp
(lib_symbols
  (symbol "Device:R" ...)   ; stock definition,  pins at ±3.81
  (symbol "R_1"      ...)   ; sheet-local edit,  pins at ±6.35
)
...
(symbol
  (lib_name "R_1")          ; <- the entry that actually applies
  (lib_id "Device:R")       ; <- provenance only
```

`lib_name` is emitted **before** `lib_id`, and the base entry is not always present — if the only instance of `Device:R` on a sheet is the edited one, `lib_symbols` contains `R_1` and nothing else.

**Write side — data loss.** `Symbol::from_sexp` kept an allow-list of child nodes and dropped the rest:

```rust
const PRESERVE: &[&str] = &["pin", "instances"];
```

`to_sexp` then rebuilt the block from the modelled fields plus that list, so every full-file write deleted `lib_name`, `exclude_from_sim`, `convert` and `default_instance` from **every symbol in the file**, including ones the caller never touched. `Sheet::from_sexp` had the same shape with `["stroke", "fill"]`.

**Read side — wrong resolution.** All 20 `lib_symbols` lookups matched `lib_id` verbatim:

```rust
lib_syms.iter().find(|n| n.get(1).and_then(|c| c.as_str()) == Some(&inst.lib_id))
```

KiCad's rule is `lib_name` when present, otherwise `lib_id`. So on untouched, eeschema-authored files Konnect either returned silently wrong pin coordinates, or falsely reported a symbol as having no embedded definition and advised re-adding it — which would destroy the local edit.

## 2. Reproducing it

Everything below uses `crates/konnect-core/tests/fixtures/derived_lib_name.kicad_sch`, added by this PR. It is 148 lines, hand-written in eeschema's format, and needs no external project. Three symbols:

| ref | `lib_id` | `lib_name` | base in `lib_symbols`? | pin pitch |
|---|---|---|---|---|
| R1 | `Device:R` | — | yes | ±3.81 |
| R2 | `Device:R` | `R_1` | yes | ±6.35 |
| C1 | `Device:C` | `C_1` | **no** | ±3.81 |

Wires and labels sit on the correct, `lib_name`-resolved pin coordinates. R1's `NET_TOP`/`NET_BOT` wires end exactly where R2's pins land *if* `lib_name` is ignored — so ignoring it produces a short rather than an open, which is the harder failure to notice.

### 2a. Write side

```bash
git checkout <this-branch>
cp crates/konnect-core/tests/fixtures/derived_lib_name.kicad_sch /tmp/repro.kicad_sch
kicad-cli sch export netlist --format kicadsexpr -o /tmp/before.net /tmp/repro.kicad_sch

git checkout main && PROTOC=$(which protoc) cargo build --release -p konnect
# then, over MCP stdio: load_toolset("sch_components")
#   add_schematic_component(schematic="/tmp/repro.kicad_sch",
#                           lib_id="power:GND", x=40, y=110, reference="#PWR099")

kicad-cli sch export netlist --format kicadsexpr -o /tmp/after.net /tmp/repro.kicad_sch
diff /tmp/before.net /tmp/after.net
```

One unrelated component added. Nets, before:

```
/NET_BOT          ['R1.2']
/NET_CAP_BOT      ['C1.2']
/NET_CAP_TOP      ['C1.1']
/NET_DERIVED_BOT  ['R2.2']
/NET_DERIVED_TOP  ['R2.1']
/NET_TOP          ['R1.1']
```

on `main` (5cd6454), after:

```
/NET_BOT          ['R1.2', 'R2.2']     <- R2 shorted to R1
/NET_TOP          ['R1.1', 'R2.1']     <- R2 shorted to R1
Net-(C1-Pad??)    ['C1.1', 'C1.2']     <- C1's two pads shorted together
```

Six nets became three. `grep -c '(lib_name'` goes 2 → 0. KiCad opens the damaged file without a single warning; only a netlist diff catches it.

With this PR the same call leaves all six nets byte-identical and both `lib_name` tokens intact.

**One trap worth knowing when reproducing.** `add_schematic_component` calls `ensure_lib_symbol` first, so adding a component whose `lib_id` happens to be the *missing base* (`Device:R` here) re-embeds that base and can mask the damage in the netlist — even though `lib_name` is still gone and the local symbol edit is still lost. Use an unrelated `lib_id`, as above. `edit_schematic_component` is unaffected either way: it writes through `konnect-sexp`'s targeted text edits rather than the typed model, so it was never lossy.

### 2b. Read side, no write required

`batch_get_schematic_pin_locations` on the **untouched** fixture.

`main` (5cd6454):

```
R1  [('1', 63.5, 59.69), ('2', 63.5, 67.31)]
R2  [('1', 88.9, 59.69), ('2', 88.9, 67.31)]
C1  ERROR: no embedded definition for 'Device:C' in lib_symbols — likely added with a nonexistent lib_id
```

this PR:

```
R1  [('1', 63.5, 59.69), ('2', 63.5, 67.31)]
R2  [('1', 88.9, 57.15), ('2', 88.9, 69.85)]
C1  [('1', 139.7, 59.69), ('2', 139.7, 67.31)]
```

R2's correct pins are at 57.15 / 69.85 — `main` silently returns `Device:R`'s geometry instead, with no error at all, and every consumer downstream (wiring, net graphs, `design_review`) inherits it. C1's definition is present under `C_1`; the error is wrong, and its advice would discard the local edit.

## 3. The fix

**Deny-list instead of allow-list.** Added `unmodelled_children(node, modelled)` in `schematic/mod.rs`; `Symbol::from_sexp` and `Sheet::from_sexp` now list the tags `to_sexp` reconstructs and keep everything else verbatim. Bare atoms are kept too. This closes the whole class rather than just `lib_name`.

**`lib_name` and `exclude_from_sim` modelled explicitly.** Raw pass-through nodes are appended after the properties, but eeschema writes both in fixed early positions, so leaving them to the generic path would reorder every file. `lib_name` is emitted ahead of `lib_id`; `exclude_from_sim` sits between `unit` and `in_bom`. It is `Option<bool>` so files that omit it don't silently gain one.

**One resolver.** Added `konnect_sexp::schematic::find_lib_symbol` — `lib_name` when present, else `lib_id`, mirroring `SCH_SYMBOL::GetSchSymbolLibraryName()` — plus `lib_symbol_name()` on both `Symbol` and `SymbolInstance`. Replaced all 20 duplicated lookup closures across `sch_components`, `sch_analysis`, `sch_batch`, `sch_export`, `sch_wiring`, `design_review` and `tools::mod`. The "no embedded definition" errors now name the identifier that was actually looked up.

**`ensure_lib_symbol`'s already-present check** was substring-searching a `{:?}` debug rendering of the whole `lib_symbols` node, which matches nested occurrences — a property value, a unit sub-symbol, or an unrelated entry containing the name. It is now a structural check on the direct children.

## 4. Compatibility

`Symbol` gains `lib_name: Option<String>` and `exclude_from_sim: Option<bool>`; `SymbolInstance` gains `lib_name: Option<String>`. Additive, but struct literals constructing either need updating — all in-tree ones are. No MCP tool, schema field, CLI flag, environment variable, config key or documented path changed, so `tool_count` in `router/registry.rs`, `tool-directory.md`, DEV.md and the README are untouched by design.

On round-trip fidelity, precisely: instances carrying no `lib_name` resolve exactly as before, and semantic behaviour is preserved throughout. But a file carrying a previously *unmodelled* token (`convert`, `default_instance`, …) now **keeps** that token rather than losing it, emitted after the typed fields — so its byte ordering can differ from the input, where before it differed by omission. KiCad's parser is order-insensitive for these tokens. `raw_and_typed_symbol_children_survive_interleaving` pins this deliberately, so restoring original positions later is a conscious change rather than a silent one; see "Out of scope".

A schematic that omits `exclude_from_sim` does not acquire one on save — there is a test for that.

## 5. Tests

The four commands from the PR checklist, all green locally:

- `cargo test --workspace --locked --lib --tests` — 488 pass, 0 fail

Also `cd crates/schematic-viewer && cargo test` (excluded from the workspace, depends on `konnect-schematic-editor`): 20 pass.

**Environment-dependent checks that did not run.** The golden-file conformance suite, `crates/konnect-core/tests/conformance_test.rs`, **skipped** on my machine. This KiCad install ships no `demos/` corpus, so `demo_dirs()` returns `None` and all three tests return early — while still reporting `ok`:

```
SKIP: no KiCAD demos found
SKIP: no KiCAD demos found
SKIP: no KiCAD demos found (set KICAD_DEMOS to enable)
test result: ok. 3 passed; 0 failed
```

I have not run that suite against this change; a maintainer with the demo corpus installed, or the scheduled `e2e-kicad.yml` workflow, would be the first to. It would not have caught this defect regardless — its corpus contains no `lib_name`, which is the point of the subsection below. Every regression test this branch adds is a plain fixture test that needs no KiCad install and runs unconditionally.

New coverage:

- `crates/konnect-core/tests/fixtures/derived_lib_name.kicad_sch` — the first fixture in the repo with the real derived shape.
- 7 tests in `konnect-schematic-editor/tests/integration.rs`: parsing, `lib_symbol_name` precedence, round-trip survival, `lib_name`-before-`lib_id` ordering, unmodelled-token survival, `exclude_from_sim` positioning and non-invention, and "editing one symbol does not corrupt the others".
- `pin_locations_resolve_through_lib_name_not_lib_id` in `sch_components`, covering both the wrong-geometry and false-error paths.
- 2 fixture-driven tests in `konnect-core/tests/integration_test.rs`, one asserting every pin position is unchanged after an edit.

Added in the second commit, from the reviews:

- `ensure_lib_symbol_presence_check_is_exact_and_structural` — exact direct child (present, not embedded twice), the name occurring only nested as an unrelated entry's property value, a longer near-miss sibling name, and a match in a second `lib_symbols` block. The negative cases use `lib_id`s that resolve nowhere, so a `false` return proves the check declined to claim presence rather than merely failing to embed.
- `raw_and_typed_symbol_children_survive_interleaving` — a symbol whose unmodelled tokens sit *between* the modelled ones; asserts each survives exactly once (not dropped, not duplicated by both the typed path and `raw_sub_nodes`) and that the field values survive a re-parse.

I confirmed each new test **fails** when the change it guards is reverted, so they are real regression guards rather than tautologies.

Beyond the fixture, I ran the same `add_schematic_component(power:GND)` through the real MCP server against copies of a 6-sheet KiCad project carrying 26 `lib_name` instances, diffing `kicad-cli` netlist exports against the untouched originals. On `main`, all 26 tokens are dropped and 38 net lines change across two sheets — on one, two named nets disappear and four series resistors end up shorted pin-to-pin, bypassed. With this PR, all 26 survive and every net on all six sheets exports identically. That project isn't mine to share, which is why the fixture above reproduces the same mechanism standalone.

### Why CI never caught this

`crates/konnect-core/tests/conformance_test.rs` is the golden-file suite that would have, but it **skips silently when KiCad isn't installed**, and the bundled demo corpus it runs against contains no `lib_name`. Neither existing fixture had the derived shape. The new fixture makes this a permanently guarded regression on every run.

## 6. Risk and rollback

The file-format change is confined to symbol and sheet serialization and is additive in content: tokens that were being deleted are now kept. `lib_name` and `exclude_from_sim` land in the positions eeschema uses; other previously unmodelled tokens are kept but grouped after the typed fields, as §4 describes. The deny-list means future unknown tokens are preserved rather than dropped, so the worst-case failure mode shifts from silent data loss to a reordered-but-valid token.

Reverting the second commit removes test coverage only. Reverting both restores the previous behaviour entirely; no migration, no on-disk state, no IPC or packaging surface involved.

## 7. Out of scope

- Restoring unmodelled children to their **original interleaved positions** rather than emitting them after the typed fields. Semantically irrelevant to KiCad's parser, and pinned by a test so it can't drift silently; a byte-faithful round-trip would be a nice follow-up.
- `label.rs`, `wire.rs` and `misc.rs` have no raw preservation at all — same class of defect, deliberately left for a separate PR to keep this one reviewable.
- Related but distinct: #55 (other `.kicad_sch` output bugs), #62 (derived `(extends ...)` pin resolution), #84 (remaining hardcoded-indentation scans).
- Minor aside, not changed here: `CONTRIBUTING.md` lists `protoc` as the only native prerequisite, but `cargo build --workspace` also needs `cmake` for `nng-sys` via `konnect-ipc`.

I agree to the CLA terms in CONTRIBUTING.md.


mixelpixx:
Merged main into your branch and resolved the one hunk that collided with #136 (sym-lib-table resolution), which landed just ahead of this.

`ensure_lib_symbol` now takes #136's `src: &dyn SymbolLibrarySource` parameter **and** uses your structural presence check. #136's `check_name` binding is gone — it belonged to the substring search your PR replaced. Your four presence-check cases pass a source that resolves nothing, so a `false` return still proves the check *declined* rather than a real library happening to be absent on whatever machine runs the test. I checked the other five shared files and the hunks are disjoint.

Gate green on the merge: fmt, clippy, doctests, 562 tests.

On the review itself — this is a data-corruption fix, not a round-trip nicety, and I want that on the record. `PRESERVE` at `konnect-schematic-editor/src/schematic/symbol.rs:80` lists only `pin` and `instances`, and `git grep lib_name -- '*.rs'` returned **zero hits** across the whole repo before this PR. So every full-file write dropped `(lib_name …)` from every symbol in the file, and all 20 `lib_symbols` lookups matched `lib_id` verbatim instead of KiCad's `lib_name`-else-`lib_id` rule. The netlist diff in your description — six nets collapsing to three, two resistors shorted — is the consequence, and it happens silently.

I also verified the "all 20 sites" claim rather than taking it: there are exactly 20 lookup closures on main, in exactly the 7 files you touch, and #122's new `command.rs`/`transaction.rs` add none.

The secondary fix is worth noting too, since it's easy to miss in the diff: the old `ensure_lib_symbol` presence check substring-searched a `{:?}` debug rendering of the entire `lib_symbols` node, so a property value or a unit sub-symbol containing the name read as "already embedded" and the real definition was never written.

Merging once CI is green. This closes #143.

Issue #101, opened 2026-07-29
Schematic property placement: use library anchors instead of hardcoded ±3.81, account for rotation


mixelpixx:
Follow-up from PR #95's review and post-merge verification.

`positioned_property` (crates/konnect-core/src/tools/mod.rs) is now the single shared path for symbol-instance property placement (`add_schematic_component` + `add_power_symbol`), but it hardcodes Reference at `y − 3.81` and Value at `y + 3.81`, ignoring both the symbol's library anchors and the instance rotation.

Verified against KiCAD 10's power library (`power.kicad_symdir`):
- **GND**: library Value anchor (0, −3.81) → sheet `y + 3.81`. Our hardcoded offset **matches eeschema exactly** — pure luck of the common case.
- **VCC**: library Value anchor (0, +3.556) → sheet `y − 3.556` (label *above*, since the graphic points up). We place it *below* at `y + 3.81` — no overlap, but not where eeschema would put it, and the same will hold for every up-pointing power flavor (+5V, +3V3, …).
- **Rotation**: a symbol placed at 90° keeps axis-aligned property offsets, so labels sit beside the wrong edge (pre-existing, shared by both call paths).

Suggested fix, mirroring the PCB side's `IpcFieldPlacement` precedent from #70: extract the Reference/Value `(property … (at x y rot))` anchors from the resolved library symbol, negate Y for sheet coords, rotate by the instance rotation, and fall back to today's ±3.81 only when the library provides no anchor.

mixelpixx:
Shipped in **v0.3.2** — https://github.com/mixelpixx/Konnect/releases/tag/v0.3.2

PR #148, opened 2026-08-10
fix(sch): orient net labels away from the symbol body

### LINKED ISSUE #147: Net labels render across the symbol body: batch_connect_to_net hardcodes rotation 0, ignoring which way the pin faces
## Summary

Labels placed on component pins render on top of the pin names KiCad draws inside the symbol body, making both unreadable.

`handle_batch_connect_to_net` writes every label at rotation `0.0` (`sch_batch.rs:391`). `label_justify(0)` is `left`, so the text runs east from the anchor. A pin on the **left** edge of an IC has its body to the east, so the label runs straight across it.

Nothing on any label path consults the pin's orientation, even though `LibPin` has carried `rotation` and `length` since the start.

## Repro

Place any IC with left-edge pins, then label one:

```
batch_connect_to_net(schematic, net_name="SWDIO", pins=[{reference:"U1", pin_number:"5"}])
```

The label is written as `(at <tip> 0)` with `(justify left bottom)` and renders over the pin name. Flipping it by hand to `(at <tip> 180)` / `justify right bottom` fixes it — the coordinates were already correct.

## What eeschema does

Across the 115 demo schematics shipped with KiCad 10, matching every label anchored on exactly one unit-aware pin endpoint:

- **249 of 249** labels on horizontal pins use the pin's *outward* direction (library pin angle + 180, through the instance rotation/mirror). No counter-examples.
- All 10 labels on vertical pins keep the text horizontal. No pin-anchored label in the corpus is ever rotated 90 or 270.

## Related defects with the same cause

- `connect_to_net` sets `at.rotation` directly instead of calling `set_rotation()`, so it emits labels with **no `(effects)` block at all** and KiCad centres the text on the anchor. This is #43 surviving in a second handler after PR #44 fixed `add_schematic_net_label`.
- `direction` on `connect_to_net` / `connect_passthrough` is a caller hint defaulting to `"right"`, never checked against geometry. `SKILL.md` documents `connect_to_net(component_reference, pin_number, net_name)` — a signature the tool does not have — so the model never passes `direction`.
- `batch_connect_to_net` is the only label path still using non-unit-aware `extract_lib_pins`, so another unit's pin resolves to a superimposed phantom endpoint instead of erroring (#35).
- Symbols that stack pins on one endpoint (a USB-C receptacle has four GND pins there) get one label each, rendering as a smeared bold string.

Found while building a real board with Konnect; every left-edge pin of the main IC was affected.

pauliuszaleckas:
Fixes #147.

## Problem

`batch_connect_to_net` wrote every label at rotation `0.0`, so `label_justify` returned `left` and the text ran east from the anchor. On a **left-edge pin the body is east**, so the label rendered across it, over the pin names KiCad draws inside. No label path consulted the pin's orientation, though `LibPin` has carried `rotation` and `length` all along.

## What eeschema actually does

Measured across all 115 KiCad 10 demo schematics, matching every label anchored on exactly one unit-aware pin endpoint:

| population | result |
|---|---|
| labels on horizontal pins | **249 / 249** use the pin's outward direction. No counter-examples. |
| labels on vertical pins | 10 / 10 keep the text horizontal (6 chose 0, 4 chose 180) |
| labels on mirrored instances | 33 matched, 32 agree (the miss is a vertical pin) |

The stronger invariant: **no pin-anchored label in the corpus is rotated 90 or 270**, corroborated by 2562 wire-endpoint labels (~1% vertical, and horizontal wins 48:15 on vertical stubs). So the `"up" => 90` / `"down" => 270` rows in both direction tables are gone too.

`pin_label_rotation` encodes this. The vertical branch is our tie-break, not eeschema's, and the doc comment says so.

## Changes

`transform_direction` (konnect-sexp/geometry.rs) maps a library-space direction through a placement by pushing a unit vector through `transform_pin` itself — rotation *and* mirror come for free. `pin_outward_direction` / `pin_label_rotation` sit next to `pin_endpoint`; `label_justify` needed no change, and the two agree at every angle.

Same root cause, also fixed:

- **`connect_to_net` emitted no `(effects)` block at all** — it set `at.rotation` directly instead of calling `set_rotation()`, so KiCad centred the text on the anchor, half of it over the body whatever the rotation. #43 surviving in a second handler after PR #44 fixed `add_schematic_net_label`.
- **`direction` gains `"auto"`, now the default** on `connect_to_net` and `connect_passthrough`: it resolves from pin geometry, falls back to `"right"` on a bare point, and the response echoes the resolved value. `"right"` still works verbatim. `connect_passthrough`'s schema had no `enum` at all; it does now.
- **`connect_to_net` accepts `reference` + `pin_number`** — the signature `SKILL.md` already documented but the tool never had. `pin_x`/`pin_y` keep working.
- **`batch_connect_to_net` was the last label path on non-unit-aware `extract_lib_pins`**, so another unit's pin resolved to a superimposed phantom endpoint instead of erroring (#35).
- **Stacked pins got a label each** — four GND pins sharing one endpoint rendered as four superimposed strings. One label serves them now; the rest are reported in `added[]` marked `deduplicated`, never `errors[]`, since they *are* connected by the shared endpoint.
- `get_schematic_pin_locations` reports `orientation_degrees` and `length_mm` (additive keys) so callers on the manual `add_schematic_net_label` path can choose correctly. Both tool descriptions, `tool-directory.md`, and the schematic skill document them.
- **Resolving a pin by name stopped at the first instance sharing the reference.** A multi-unit part places one `(symbol …)` node per unit under one reference — 26 of the 115 demo schematics do — so every unit past the first was unreachable, an op-amp's power pins among them. Every placed unit is searched now, and the not-found error names all of them.

## Review notes

**Behaviour changes:** the unit-aware switch turns a silent wrong-coordinate label into an `errors[]` entry — correct per #35, but it looks like a new failure to a script that relied on the accident. And `up`/`down` stubs now get horizontal labels, changing the `label_rotation` those two tools return.

No tools added or removed, so no `tool_count` movement.

Not in scope: `check_schematic_overlaps` compares anchor points only and structurally cannot see label-over-body collisions (that needs text extents + body rects); `fix_connectivity` has the same non-unit-aware pin bug with a different outcome; #101 is the same class of bug for property anchors.

## Testing

Gate green — `cargo fmt --check`, `clippy --workspace -D warnings`, 500 tests, 4 doctests.

- **`conformance_test.rs` is the real oracle**: it replays the measurement above against the installed KiCad, asserting zero disagreements on horizontal pins, zero vertical pin-anchored labels, and a floor of 200 matches so a matcher that stops matching cannot pass vacuously. It landed *before* any handler changed, so it validates the helper independently.
- `transform_direction` ground-truth table (4 angles × 4 rotations × both mirrors) plus a guard that a mirrored pin does not report the unmirrored answer — the mirror is easy to "simplify" away, and 33 corpus cases exercise it.
- First tests for `batch_connect_to_net` and `connect_to_net`, which had none. Both verified to fail against the old behaviour: stubbing `pin_label_rotation` to `0.0` fails the left-edge test, dropping the outward flip fails the conformance test.
- **e2e against real KiCad**: `full_design_loop_with_real_kicad` places a label, then asserts R1 pin 1 appears as a *node of* the VIN net in eeschema's own netlist — a substring match on the net name would pass with the label detached, which is the only thing this test exists to catch. Verified both ways against KiCad 10.0.4: green as written, and red when the assertion looks for the wrong pin.
- Naming a pin now outranks matching one by coordinate, tested on two parts butted pin-to-pin: the named pin resolves its own direction where the coordinate lookup has to give up and fall back to `right`.


mixelpixx:
Merged main into your branch. This was the most involved resolve of the round — eight conflicts across four files — because this PR, #144 (`lib_name` resolution) and #179 (multi-unit pins) each rewrote the same schematic pin-resolution path, and all three landed in the same session. Taking any one side whole would have silently undone another, so the detail is worth stating:

- **`tools/mod.rs`** — your `placed_pins()` / `pin_outward_at()` refactor is kept, with `find_lib_symbol()` inside it. That means `all_pin_endpoints` — and therefore junction insertion — picks up the `lib_name` rule for free.
- **`sch_wiring.rs`** — `resolve_placed_pin()` is already unit-aware and **subsumes #179's inline fan-out**, so it stays as the single resolution path; its lookup now goes through `find_lib_symbol`.
- **`sch_batch.rs`** — `handle_batch_connect_to_net` routes through `resolve_placed_pin` rather than #179's inline candidate loop. Strictly more capable: same per-unit resolution, plus your label rotation and duplicate-label suppression. Both test modules kept.
- **`sch_components.rs`** — import union.

Full gate green: fmt, clippy, doctests, 596 tests — all four `connect_to_net_orientation_tests` and all seven multi-unit tests passing together.

On the review: deriving the rule from all 115 KiCad 10 demo schematics (249/249 horizontal-pin labels follow the pin's outward direction, no pin-anchored label anywhere rotated 90/270) is the right way to settle a question like this, and it's why I merged it without hesitation. Pushing a unit vector through the existing `transform_pin` so rotation *and* mirror fall out rather than being re-derived is the other thing I'd call out.

Two disclosed behaviour changes I'm accepting knowingly: unit-aware pin resolution converts a silent wrong-coordinate label into an `errors[]` entry, and `up`/`down` stubs now return a different `label_rotation`. Both are correct-but-visible.

**One correction to the PR description:** this does not subsume #101. The `±3.81` constants on main are all in `library.rs` and are glyph *body* geometry from #130 — `gh pr diff 148 | grep 3.81` returns nothing. #101 is about Reference/Value **field** anchors; this is net **label** orientation. Different code, different concern, so #101 stays open.

Closes #147. Merging once CI is green.

PR #150, opened 2026-08-10
feat(sch): add update_symbols_from_library

pauliuszaleckas:
## Problem

A schematic carries its own copy of every symbol it places, and nothing in Konnect refreshed it after the library changed — `ensure_lib_symbol_in_schematic` returns early when the lib_id is already embedded. Edit a symbol in its library and the sheet keeps drawing the old shape, which is exactly what KiCAD's ERC reports as:

```
Symbol 'U1' doesn't match copy in library 'my-project'
```

There was no way to resolve that through the tools. `delete_schematic_component` leaves `lib_symbols` untouched, so deleting and re-placing reuses the stale copy; `replace_component` goes through the same early-returning embed.

## Fix

`update_symbols_from_library`, mirroring eeschema's command of the same name. Takes an optional `references` list (default: every symbol) and `dry_run` to preview. Reports `updated`, `unchanged`, and per-lib_id `errors` — a lib_id that no longer resolves leaves its embedded copy alone rather than dropping it.

## Review notes

Two details worth a look:

- **The lookup is structural, not a string search.** The two embed paths in the codebase lay a definition out differently — `add_schematic_component` writes the name inline, the resolver's writer puts it on the next line — so a fixed `(symbol "Nick:Name"` pattern finds the definition before an update and misses it after. `konnect_sexp::writer::find_direct_child_blocks(content, "lib_symbols")` sidesteps that: it is scoped to the `lib_symbols` node, string-aware, and indentation-agnostic, so it cannot match a placed instance's own `(lib_id …)` or a property value that happens to contain the name.
- **"Unchanged" compares parse trees, not bytes.** For the same reason, a byte comparison reports every definition as changed the first time and rewrites the file to no purpose. Comparing `parse_sexp` output rather than normalized text also avoids conflating layout whitespace with whitespace *inside* a quoted property value — a library edit that only changed the latter would otherwise be reported as unchanged and silently skipped.

The batch shape (`reembed_lib_symbols` takes the whole list) is deliberate: `lib_symbols` is located once, and the edits are applied back to front so earlier offsets stay valid.

Deliberately explicit rather than automatic: KiCAD keeps the schematic's copy until you ask it to update, and silently overwriting on placement would clobber a deliberately-edited embedded symbol.

Per CLAUDE.md, four counts moved together — `tool!(…)` + handler, `registry.rs` `tool_count` (17 → 18), `tool-directory.md` section and Overview totals, and `DEV.md`/`README.md` (187 → 188).

## Testing

Gate green — `cargo fmt --check`, `clippy --workspace -D warnings`, full `--lib --tests`, doctests.

New coverage: a symbol edited out from under a schematic is detected, reported by `dry_run` without writing, refreshed on the real run, and reported `unchanged` on a second run (idempotent); an unknown reference is an error rather than a silent no-op. `protocol_stdio`'s batch-load count moves 36 → 37.

Verified against a real board: after the update the embedded copy parses identical to the library definition.

**Not fixed here:**

- KiCAD still emits the "doesn't match copy in library" warning for that board even with the copies identical, so something else in its comparison differs — most likely our `(pin_numbers)` empty node, which KiCAD 10 omits entirely when numbers are shown. Separate investigation; ERC errors are unaffected (0 before and after).
- The two embed paths still write different layouts. `ensure_lib_symbol_in_schematic`'s presence check is `content.contains("(symbol \"<lib_id>\"")`, and `schematic_builder.rs` keys on the same literal, so neither sees a definition written with the name on the next line. Unifying them needs a depth-aware serializer in `konnect-schematic-editor`'s writer — out of scope here, worth a follow-up.


mixelpixx:
Adjudicated in your favour over #177 (@JYPochez), which implements the same tool. Recording why, and what's left to do.

**Why this one wins:**

- **Doc/registry discipline.** You update all five required sites — `registry.rs`, `tool-directory.md` (header count, toolset count, and the table row), `README.md`, DEV.md in four places, plus the `protocol_stdio.rs` counter. #177 misses DEV.md entirely and the README. Worth noting `registry_tool_counts_match_reality` would have passed for both — it only checks `registry.rs` against `tools_for()`. Main now has a real guard for the rest (`crates/konnect/tests/doc_tool_counts.rs`).
- **Reuse over reinvention.** You use `konnect_sexp::writer::find_direct_child_blocks`, which is string-aware and indentation-agnostic. #177 hand-rolls a paren counter that doesn't skip quoted strings, so a stock symbol carrying `(property "Description" "Dual Op Amp (DIP-8)")` unbalances it — and parenthesised descriptions are common in KiCad's shipped libraries.
- **Idempotence.** You compare parsed trees and re-indent through a shared helper, with a test asserting a second run reports `updated_count: 0`. #177 compares `fresh.trim() == embedded.trim()`, but the embedded copy is tab-indented while the resolver returns unindented text — so every symbol always compares as changed, is always rewritten, and is written back de-indented.

**One thing I want grafted from #177 before merge, and it's a safety issue.** It guards against pin movement: it compares pin anchors before and after and refuses any symbol whose pins moved unless `allow_pin_moves` is passed. Yours will silently move pins out from under existing wires and labels, with only a sentence in the description as warning — and orphaned connections are exactly the kind of damage that shows up later as a netlist you can't explain.

I'd reimplement it structurally over the parsed trees rather than porting #177's version: it compares coordinate *tokens* as strings, so `1.27` vs `1.270` reads as "moved", and it sorts anchors without keeping pin numbers attached, so two pins swapping positions compares equal — missing the one case it most needs to catch. You already have both trees parsed at the comparison point, so it's cheap. Default `allow_pin_moves: false`, and report skipped symbols with before/after positions.

**Also needs a rebase onto #136**, which landed today: `resolve_lib_symbol_flattened` is 2-arg now and `resolve_lib_symbol` is gone. Thread a `KiCadSymbolSource::for_file(&sch_path)` through `reembed_lib_symbols`. That's not just repair — it's the point, since post-#136 the tool refreshes from the user's *registered* libraries, which is what "Update Symbols from Library" should mean.

Deferring to the next round with the rest of the new-tool queue. Your #148, #149, #151, #155 and #136 all landed today — thanks, that was a lot of good work in one batch.

mixelpixx:
Landing this as the winner of the #150/#177 pair — the structural lookup and the parse-tree comparison were the deciding differences, exactly as your review notes framed them. I pushed a merge of current `main` to your branch (thanks for `maintainerCanModify`) with two changes beyond conflict resolution:

**Rebase onto #136's source-based resolution.** `resolve_lib_symbol_flattened` now takes a `SymbolLibrarySource` and the raw local fallback is gone, so `resolve_lib_symbol_for_embed` no longer had a reason to exist — both the embed path and `reembed_lib_symbols` call the flattened resolver directly, with the source built from `KiCadSymbolSource::for_file`. Your `indent_lib_symbol` factoring is kept and now serves both paths.

**Grafted #177's pin-anchor guard, credited to @JYPochez.** The one thing #177 had that this PR didn't: wires and labels attach at pin coordinates, so a library edit that moved or removed a pin silently orphans them. `reembed_lib_symbols` now compares pin anchors between the embedded and fresh definitions — over the parse trees it already builds for the unchanged check, not a string compare — and refuses such symbols with per-pin detail in a `pins_moved` response field, unless `allow_pin_moves` is set. New pins don't trigger it; nothing attaches to a pin that didn't exist. Body/property-only edits (your original test) update exactly as before.

Verification, the full four steps:
- Unit: moved pin refused with `pin 2 moved from (0, -3.81) to (0, -5.08)` detail and an untouched file; removed pin refused as `removed`; `allow_pin_moves: true` applies; your value-only and idempotence tests still pass as the safe case.
- Neutered the guard and confirmed both new tests fail — they aren't vacuous.
- Full gate green: fmt, clippy `--all-targets -D warnings`, 654 workspace tests, doctests. `doc_tool_counts.rs` (new since you filed) now *enforces* the 188/194 totals your PR updates.
- End-to-end through the real binary over stdio against KiCAD 10.0.5: project-local library, place, edit lib, dry-run reports without writing, update lands, pin move refused then applied with the flag — and `kicad-cli sch erc` reads the updated file with pins at the refreshed anchors, no new violation class vs. the pre-edit baseline.

On your "not fixed here" note about the residual "doesn't match copy in library" warning: still worth its own investigation, agreed on the `(pin_numbers)` suspicion — tracked implicitly under the embed-path unification follow-up.

PR #153, opened 2026-08-11
fix(pcb): read layers by shape, and stop add_layer writing unopenable boards

cvaldess:
A layer entry is `(0 "F.Cu" signal)` — its head is the ordinal, so unlike every other table in the board file there is no tag to match on. All three sites that read the stackup used `find_all("")`, which filters children by head and so never matched a layer.

This is **not** a KiCad 10 format change: layers have always been written this way, and these have always returned nothing.

| | before | after | board |
|---|---|---|---|
| `get_board_info.layer_count` | `0` | `27` | 27 declared |
| `get_board_info.copper_layer_count` | — | `2` | F.Cu, B.Cu |
| `get_layer_list.count` | `0` | `27` | 27 declared |

`get_layer_list` was doubly broken: its accessors read `get_f64(1)`/`get(2)`/`get(3)`, one place late, because the ordinal occupies the slot a tag would. Even with the filter fixed it would have returned an empty list.

## `add_layer` wrote boards that do not open

Three separate faults, all of which returned success.

**It never saw the ids in use.** With the entries unreadable, the "first free id in 1..30" search always returned `1` — duplicating `In1.Cu` on any board that has one.

**It wrote the new layer inside the first entry.** The end of the block was located with a literal `"\n  )"` probe, falling back to the first `')'` in the block. On a tab-indented KiCad 10 file the probe never matches and the fallback is the close of the *first layer entry*. Actual output of v0.2.2 on a real board:

```
(layers
		(0 "F.Cu" signal "Top Layer"
    (1 "TestLayer" user))
		(2 "B.Cu" signal "Bottom Layer")
```

Now located by paren balance, which is indentation-agnostic, and the entry is indented to match whatever the file already uses.

**It accepted any `layer_name`.** KiCad's layer set is closed. A board carrying a name outside it does not degrade — it does not open at all. Measured with `kicad-cli pcb export svg` on a KiCad 10 board:

| inserted | result |
|---|---|
| `(53 "User.8" user)` | loads |
| `(4 "User.8" user)` | loads |
| `(53 "TestLayer" user)` | **`Failed to load board`** |

So the id was never what KiCad validated — the name was, and the tool invited the caller to invent one. `add_layer` now refuses a non-canonical name and leaves the file untouched, pointing the caller at the real mechanism for a custom label: add the canonical layer and give it a user name, `(53 "User.8" user "My Label")`.

## Notes on the implementation

Stackup reading moves to `konnect-sexp::layers`, next to `geometry`, as file-format knowledge rather than tool-level code.

**Copper is decided by the `.Cu` suffix, not by an allow-list of kinds.** KiCad marks copper as `signal`, `power`, `mixed` *or* `jumper`, so a board using `power` for a plane would be undercounted by a kind check.

**`layer_count` stays the count of declared entries** and `copper_layer_count` is added alongside, rather than redefining `layer_count` to mean copper. It keeps `get_board_info` and `get_layer_list` reporting the same number for the same board, and "how many layers is this board" — the fab-house question — still gets a direct answer.

**The name check is backed by KiCad's own `BoardLayer` enum** from the API protos already vendored in `konnect-ipc`, mapping `BL_F_Cu` → `F.Cu` and `BL_User_1` → `User.1`. A test walks every variant of that enum and asserts the checker accepts it, so the list cannot drift from KiCad as the protos are updated.

## Verification

`cargo clippy --workspace --locked -- -D warnings`, `cargo fmt --all -- --check` and `cargo test --workspace --locked --lib --tests` all clean on `x86_64-pc-windows-msvc`; **488 passed, 0 failed**, of which 22 are new.

End-to-end against a real KiCad 10 board (`version 20260206`, tab-indented, 27 layers / 2 copper), driving the built binary over MCP:

- the read figures in the table above;
- `add_layer "TestLayer"` → refused, file **byte-identical** to the original;
- `add_layer "User.8"` → id **4**, the first genuinely free ordinal on that board (it uses 0,1,2,3,5), written after the last entry, and **the result loads in KiCad** where the previous output did not.

## Relationship to #133 / #142

Independent of both. #133 is about the *shape* of `(net …)` nodes under KiCad 10, and #142 fixes it; this is a table that was never readable on any version. They touch neighbouring lines of `handle_get_board_info`, so whichever lands second may want a trivial rebase — happy to do that on this one either way.

`layer_count` is also the one item I flagged as still `0` after applying #142, so this closes it.

### DIFF
--- a/crates/konnect-core/src/tools/pcb_board.rs
+++ b/crates/konnect-core/src/tools/pcb_board.rs
@@ -139,6 +139,53 @@ fn find_net_id(content: &str, net_name: &str) -> Option<i32> {
     before[num_start..num_start + num_end].parse().ok()
 }
 
+/// Byte offset of the `)` that closes the block opening at `open_pos`.
+///
+/// Balances parens while skipping quoted strings, so it is independent of how
+/// the file is indented — KiCad 9 writes two spaces, KiCad 10 writes tabs, and
+/// a probe for either is wrong on the other.
+fn close_of_block(content: &str, open_pos: usize) -> Option<usize> {
+    let mut depth = 0usize;
+    let mut in_str = false;
+    let mut escaped = false;
+    for (i, c) in content[open_pos..].char_indices() {
+        if in_str {
+            match c {
+                _ if escaped => escaped = false,
+                '\\' => escaped = true,
+                '"' => in_str = false,
+                _ => {}
+            }
+            continue;
+        }
+        match c {
+            '"' => in_str = true,
+            '(' => depth += 1,
+            ')' => {
+                depth -= 1;
+                if depth == 0 {
+                    return Some(open_pos + i);
+                }
+            }
+            _ => {}
+        }
+    }
+    None
+}
+
+/// The leading whitespace of the first entry inside the block at `open_pos`,
+/// so an inserted sibling matches the file it is written into.
+fn entry_indent(content: &str, open_pos: usize) -> Option<String> {
+    let after = &content[open_pos..];
+    let nl = after.find('\n')?;
+    let line = &after[nl + 1..];
+    let indent: String = line
+        .chars()
+        .take_while(|c| *c == ' ' || *c == '\t')
+        .collect();
+    (!indent.is_empty() && line[indent.len()..].starts_with('(')).then_some(indent)
+}
+
 // ─── Tool definitions ─────────────────────────────────────────────────────────
 
 pub fn tools() -> Vec<ToolDef> {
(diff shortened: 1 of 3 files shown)


mixelpixx:
Merged main into this branch and resolved the two collisions with #142 (KiCad 10 net nodes by shape), which landed just ahead of it. Pushed to your branch — nothing about your change was altered.

- `crates/konnect-sexp/src/lib.rs` — add/add: you added `pub mod layers;`, #142 added `pub mod net;`. Kept both.
- `crates/konnect-core/src/tools/pcb_board.rs` — `handle_get_board_info` auto-merged, and I checked it rather than trusting it: the result carries `layer_count` and `copper_layer_count` from here alongside `net_count` from #142's `count_distinct_nets`. No field was dropped.

Gate is green on the merge: fmt, clippy, and 530 tests passing.

For the record on the review: `close_of_block` being string-aware — skipping quoted strings and tracking escapes — is better than the paren scanners already in this repo, and `entry_indent` matching the file's own indentation rather than hardcoding spaces is the right instinct. The `is_canonical_name` fail-closed behaviour on `add_layer` is a hard behaviour change for callers, but it is the correct one: the old code wrote boards KiCad refuses to open and reported success, which is the same class of silent failure as the `add_via` bug we shipped in v0.2.1.

One thing I'm adding on top after this lands: `tool-directory.md` rows for `get_board_info`, `get_layer_list` and `add_layer`, since all three change contract here and the directory still describes the old behaviour.

PR #159, opened 2026-08-13
feat(library): inspect and edit footprint graphics by UUID

Pytonballoon810:
## Problem and scope

The `library` toolset can create a whole footprint and edit its pads, but it cannot touch a single line of an existing footprint's graphics. Any change to an outline that already exists — a courtyard, a fabrication-layer rectangle, or a pin-1 marker that turns out to sit exactly where a panel's mouse-bite holes will open the solder mask — has no route through Konnect at all.

The only workaround is `create_footprint` with `overwrite`, which rebuilds the part from a generic pad description. For a footprint with custom pad shapes, a long `descr`, and 3D models, that loses more than it fixes.

This adds the missing read/write pair for footprint graphics. Pads keep their existing tool; nothing changes for them.

## Design

Two tools in `library`:

- **`list_footprint_graphics`** — every `fp_line`, `fp_rect`, `fp_circle`, `fp_arc` and `fp_poly` in a `.kicad_mod`, with UUID, layer, stroke width and vertices. Optional `layer` filter. Items come back in file order.
- **`edit_footprint_graphic`** — updates one item's geometry, `layer`, or `width_mm`, selected by UUID.

Four decisions worth reviewing:

1. **Addressed by UUID, not index.** A footprint's graphics are reordered whenever KiCad rewrites the file, so an index would silently address a different outline after any unrelated edit.

2. **One `points` array for every shape**, interpreted per kind in the order KiCad stores the vertices: 2 for `fp_line`/`fp_rect` (start, end), 2 for `fp_circle` (center, circumference point), 3 for `fp_arc` (start, mid, end), 3+ for `fp_poly`. Arity is checked per kind and the error names the vertices that kind wants — a 2-point `fp_poly` is the easy mistake, and writing one silently produces a footprint KiCad will not open.

3. **Block surgery, not a rewrite of the item.** Only the geometry sub-blocks, `(layer …)`, or the width inside `(stroke …)` are replaced, so `fill`, stroke `type`, `locked` and `uuid` all survive a geometry-only edit. `width_mm` is resolved *inside* the item's `(stroke …)` block rather than by a bare search, so it stays correct if a future format revision puts another `(width …)` elsewhere in the item.

4. **`fp_text` is out of scope.** Its position is a single `(at x y angle)` carrying a rotation this schema cannot express, and its size lives in `(effects (font …))` rather than in a stroke. It deserves its own tool.

`width_mm` carries its unit per `docs/NAMING_CONVENTIONS.md`. The older sibling `edit_footprint_pad` spells its pad dimensions `width`/`height` without units; those mean pad *size* rather than *stroke* width, so the distinct name is deliberate rather than an inconsistency — happy to rename if you would rather the two tools match.

## Compatibility

Additive only. `library` goes 14 → 16 tools and the registered total 187 → 189. `registry.rs`, `tool-directory.md` and `DEV.md` are updated to match.

One cosmetic note: like `edit_footprint_pad`, the rewritten block is emitted inline, so a multi-line `(pts …)` comes back as one line. KiCad reads both and re-expands on its next save.

## Tests

8 new unit tests in `tools::library::tests`, driving the real handlers against footprints written in KiCad 10's tab-indented style:

- the read path reports UUID, layer, width and points, in file order;
- the `layer` filter;
- a geometry-only edit takes the new vertices and leaves `fill`, `stroke`, `uuid` and the sibling items untouched;
- `width_mm` lands inside `(stroke …)` and does not disturb the other items' widths;
- both arity guards (`fp_poly` under 3 points, `fp_arc` not exactly 3), including that a rejected edit leaves the file byte-identical;
- an unknown UUID, whose error points at `list_footprint_graphics`;
- an edit with nothing to change.

```
cargo test --workspace --lib --tests   # all green
cargo fmt --all -- --check             # clean
cargo clippy -p konnect-core --lib     # no new warnings
```

Also validated end to end against the release binary over stdio on a real KiCad 10 footprint: `load_toolset("library")` reports 16 tools, the polygon is located by UUID, moved, and re-read at its new vertices, and the resulting file differs from the original in the `(pts …)` block and nothing else.

## Risk and rollback

File-format risk is confined to `.kicad_mod` graphic items and only on the sub-blocks named in the request; a request that fails validation writes nothing. No IPC, packaging, or protocol surface is touched. Reverting the commit removes both tools and restores the counts.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -206,7 +206,7 @@ if !path.exists() {
 
 Adding a new kind: edit `mcp/error.rs`, add the variant, add the match arm in `short_code()`, use it from the handler. The `short_code_matches_serialized_kind_field` test will fail loudly if they drift.
 
-The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 187 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
+The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 189 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
 
 ## Observability
 
@@ -227,7 +227,7 @@ Source: [`crates/konnect-core/src/observability.rs`](crates/konnect-core/src/obs
 
 ## Tool Routing (Starter Kit + On-Demand Loading)
 
-The server does NOT expose all 187 tools (193 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
+The server does NOT expose all 189 tools (195 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
 
 - **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 6 meta-tools, baseline `tools/list` is ~19 tools ≈ 2K tokens.
 - **On demand**: the LLM reads `list_toolboxes` → calls `load_toolset(name)` to expose a toolset's tools in subsequent `tools/list` responses. `unload_toolset(name)` prunes them when the task shifts.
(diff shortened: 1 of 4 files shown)


mixelpixx:
Holding this one because it is contained in #160.

Both PRs add the same `crates/konnect-core/src/tools/library.rs` change — **+578/-0, byte-identical**. #160 is this plus the IPC layer (`client.rs +353`, `pcb_components.rs +105`, `types.rs +13`), which is why its `registry.rs` diff moves two toolsets where this one moves one.

GitHub reports both as mergeable against main independently, but that doesn't mean they compose: merging one and then the other conflicts on all 578 lines. So it's one or the other, not both.

My preference is to merge **#160 alone**, since it's the complete feature — unless you'd rather land this half first, in which case say so and I'll take this and have #160 rebased on top.

One thing to fix in whichever lands: both bump `registry.rs`, `tool-directory.md` and DEV.md but leave `README.md` at the old count. Main now has a test that catches exactly this (`crates/konnect/tests/doc_tool_counts.rs`) — it will fail on the current diff, so it's worth updating before you rebase.

Deferring to the next round, when the new-tool queue gets worked through.

Pytonballoon810:
Agreed — closing this in favour of #160. It carries this commit byte-for-byte, the split was only ever an artifact of a fork branch not being able to base an upstream PR, and there is nothing here that is better reviewed on its own.

You're right about `README.md`, and it's a miss rather than an ambiguity — line 16 still reads **187 tools across 18 on-demand toolsets**, which is 191 with both commits. On the rebase I'll move `README.md`, `registry.rs`, `tool-directory.md` and `DEV.md` together and check the result against `doc_tool_counts.rs` before pushing.

The Copilot points against `library.rs` in this thread — the `require_str` downgrade, `fmt_mm`, and the empty-string UUID — I've answered on #160, since that is where the commit now lands. Short version: taking all three, with a correction on the reasoning for `fmt_mm`.

No objection to the round scoping. Everything continues on #160.

PR #160, opened 2026-08-13
feat(pcb_components): inspect and edit a placed footprint's graphics over IPC

Pytonballoon810:
> **Stacked on #159.** This branch contains that PR's commit as well, because a fork branch cannot be a base for an upstream PR. Review/merge #159 first and this diff reduces to its own commit, `feat(pcb_components): …`.

## Problem and scope

Konnect can move, rotate, swap and delete a footprint on the board, but it cannot touch the artwork *inside* one. So when a library footprint changes, there is no way to bring a single placed instance in line with it — the board carries its own copy of every footprint, and the only routes are KiCad's own **Update Footprints from Library** in the GUI, or re-placing the part and rebuilding its net assignments.

The case this came from: a pin-1 marker sat 0.1 mm from its board edge, a panel perforated that exact edge, and every panel built from the part carried 16 `silk_over_copper` warnings. Fixing the library (#159) did nothing for the boards already using it.

Two IPC tools in `pcb_components`:

- **`list_board_footprint_graphics`** — every graphic item inside a placed footprint with UUID, kind, layer and vertices; optional `layer` filter.
- **`edit_board_footprint_graphic`** — replaces one polygon's vertices, selected by UUID.

## Design

**Coordinates are the part worth reviewing.** KiCad carries a footprint's children in *absolute board* coordinates over IPC — the same fact `move_footprint` already deals with when it shifts them — while the `.kicad_mod` a user reads shows them *footprint-local*. Both tools take and report local coordinates and do the conversion themselves, rotating by the instance's orientation and translating onto its anchor. The point is that someone comparing board artwork against its library source is comparing like with like; the alternative is every caller redoing that arithmetic against the wrong frame.

The reverse transform reuses `transform::Xform` rather than open-coding a second rotation, so the two directions cannot drift. `Xform::point` becomes `pub(crate)` for it.

**Only polygons are editable for now.** Segments, arcs, circles, rectangles and beziers are reported by the list tool but rejected *by name* on edit — each wants its own vertex contract, and this is the one the motivating case needed. The error says which kind it found.

## Tests

6 new unit tests in `konnect-ipc`:

- the frame helper with and without a position/orientation;
- shape-kind naming;
- vertex extraction for a polygon outline and for a segment, in KiCad's storage order;
- a **local → board → local round trip at 90°**, where a sign error in the rotation would be visible and silent otherwise.

```
cargo test --workspace --lib --tests   # all green
cargo fmt --all -- --check             # clean
```

Also validated end to end against KiCad 10 over its live IPC socket: the polygon was read back at its `.kicad_mod` coordinates, moved, re-read at the new ones, and after `save_project` the board on disk carried exactly that change and nothing else beyond KiCad's own `(type default)` → `(type solid)` normalisation of the footprint it rewrote.

## Compatibility and risk

Additive. `pcb_components` goes 13 → 15 tools and the registered total 189 → 191; `registry.rs`, `tool-directory.md` and `DEV.md` updated. Both tools need KiCAD running with the board open, like the rest of the toolset. The write path touches one graphic item inside one footprint; a request that fails validation sends no `UpdateItems`. Reverting the commit removes both tools and restores the counts.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -234,7 +234,7 @@ if !path.exists() {
 
 Adding a new kind: edit `mcp/error.rs`, add the variant, add the match arm in `short_code()`, use it from the handler. The `short_code_matches_serialized_kind_field` test will fail loudly if they drift.
 
-The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 200 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
+The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 202 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
 
 ## Observability
 
@@ -255,7 +255,7 @@ Source: [`crates/konnect-core/src/observability.rs`](crates/konnect-core/src/obs
 
 ## Tool Routing (Starter Kit + On-Demand Loading)
 
-The server does NOT expose all 200 tools (206 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
+The server does NOT expose all 202 tools (208 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
 
 - **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 6 meta-tools, baseline `tools/list` is ~19 tools ≈ 2K tokens.
 - **On demand**: the LLM reads `list_toolboxes` → calls `load_toolset(name)` to expose a toolset's tools in subsequent `tools/list` responses. `unload_toolset(name)` prunes them when the task shifts.
@@ -327,9 +327,9 @@ convention for other `kicad-cli`-calling code.
 
 ## Current Stats
 
-- **19 toolsets, 200 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
+- **19 toolsets, 202 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
 - Baseline `tools/list`: ~19 tools / ~2K tokens (starter kit + meta-tools)
-- Full-catalog `tools/list` (all loaded): 206 tools (200 registered + 6 meta) / ~25K tokens
+- Full-catalog `tools/list` (all loaded): 208 tools (202 registered + 6 meta) / ~25K tokens
 - **0 IPC stubs** (all protobuf methods implemented)
 - **0 unimplemented tools**
 - **3 CLI commands removed in KiCAD v10** (specctra DSN/SES, pcb sync — return clear errors)
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 Rust binary — that lets Claude and other AI assistants design schematics and PCBs
 through the [Model Context Protocol](https://modelcontextprotocol.io) (MCP).
 
-**200 tools across 19 on-demand toolsets.** Schematic capture, PCB layout and
+**202 tools across 19 on-demand toolsets.** Schematic capture, PCB layout and
 routing, ERC/DRC, design-review audits, JLCPCB part search, Freerouting, reference
 circuits, and a full manufacturing export pipeline — with bundled skills and agents
 that teach Claude KiCAD conventions out of the box.
@@ -69,7 +69,7 @@ through its own S-expression engine with atomic writes (write, fsync, rename), U
 preservation, and round-trip tests — no third-party schematic library with known
 gaps, no text-manipulation workarounds.
 
-**Context economy is a feature.** Exposing all 200 tools to an LLM costs roughly 23K
+**Context economy is a feature.** Exposing all 202 tools to an LLM costs roughly 23K
 tokens of context on every listing. Konnect's router loads a starter kit (~2K
 tokens) and lets the model pull in toolsets on demand — plus built-in observability
 (`get_recent_calls`, `server_stats`, JSONL call logs) so the model can diagnose its
--- a/crates/konnect-core/src/router/registry.rs
+++ b/crates/konnect-core/src/router/registry.rs
@@ -76,7 +76,7 @@ pub static ALL_TOOLSETS: &[ToolsetMeta] = &[
         name: "pcb_components",
         description: "Place, move, rotate, align, and duplicate PCB footprints",
         category: "pcb",
-        tool_count: 13,
+        tool_count: 15,
     },
     ToolsetMeta {
         name: "pcb_routing",
--- a/crates/konnect-core/src/tools/pcb_components.rs
+++ b/crates/konnect-core/src/tools/pcb_components.rs
@@ -99,6 +99,39 @@ pub(crate) fn resolve_footprint_source(lib_id: &str, board: &Path) -> anyhow::Re
 /// Placing on the back is not a layer rename: KiCAD's flip mirrors the
 /// footprint's geometry (pad X positions negate, every front layer swaps with
 /// its back counterpart per item). Until Konnect implements that mirror,
+/// Read a `[{x, y}, …]` argument into footprint-local millimetre pairs.
+///
+/// Each rejection names the offending index, because "each point needs a
+/// numeric 'x'" on a twelve-vertex courtyard tells the caller nothing about
+/// which vertex to fix.
+fn parse_points(value: &serde_json::Value) -> Result<Vec<(f64, f64)>, CallToolResult> {
+    let invalid = |field: &str, reason: String| {
+        CallToolResult::error_kind(
+            crate::mcp::error::ToolErrorKind::InvalidArgument {
+                field: field.to_string(),
+                reason: reason.clone(),
+            },
+            format!("Argument '{field}' is invalid: {reason}"),
+        )
+    };
+
+    let array = value
+        .as_array()
+        .ok_or_else(|| invalid("points", "missing or not an array of {x, y}".to_string()))?;
+
+    let mut points = Vec::with_capacity(array.len());
+    for (i, p) in array.iter().enumerate() {
+        let x = p["x"]
+            .as_f64()
+            .ok_or_else(|| invalid("points", format!("point {i} has no numeric 'x'")))?;
+        let y = p["y"]
+            .as_f64()
+            .ok_or_else(|| invalid("points", format!("point {i} has no numeric 'y'")))?;
+        points.push((x, y));
+    }
+    Ok(points)
+}
+
 /// pretending to support `B.Cu` silently produces wrong copper, so the layer
 /// is refused up front — before anything is resolved, sent, or written.
 fn back_side_layer_error(layer: &str) -> Option<CallToolResult> {
@@ -1354,20 +1387,14 @@ async fn handle_edit_board_footprint_graphic(
         Ok(v) => v.to_string(),
         Err(e) => return Ok(e),
     };
-    let points: Vec<(f64, f64)> = args["points"]
-        .as_array()
-        .ok_or_else(|| anyhow::anyhow!("'points' must be an array of {{x, y}}"))?
-        .iter()
-        .map(|p| {
-            let x = p["x"]
-                .as_f64()
-                .ok_or_else(|| anyhow::anyhow!("each point needs a numeric 'x'"))?;
-            let y = p["y"]
-                .as_f64()
-                .ok_or_else(|| anyhow::anyhow!("each point needs a numeric 'y'"))?;
-            Ok::<_, anyhow::Error>((x, y))
-        })
-        .collect::<Result<_, _>>()?;
+    // A malformed `points` is the caller's mistake, so it gets the structured
+    // `invalid_argument` the rest of this file returns — bubbling `anyhow`
+    // here would surface it as a handler error with the offending field
+    // flattened into prose, which is the class #194 is open against.
+    let points = match parse_points(&args["points"]) {
+        Ok(p) => p,
+        Err(e) => return Ok(e),
+    };
 
     let (ref_ipc, uuid_ipc, pts) = (reference.clone(), uuid.clone(), points.clone());
     let kind: String = ipc!(ctx, args, |c| c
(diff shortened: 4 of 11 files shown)


mixelpixx:
Reviewed and queued for the next round. Summary of where it stands so you're not waiting blind.

**This supersedes #159** — its entire `library.rs` change (+578/-0) is contained here byte-for-byte, and this adds the IPC layer on top. I'd rather merge this one alone than land both, since merging one then the other conflicts on all 578 lines.

**Two things before it can land:**

1. **It needs live verification against real KiCAD**, not just green CI. This adds 353 lines of outbound IPC message construction in `client.rs`, and no offline test can catch a malformed message there — it is schema-valid, so prost encodes and decodes it happily and only KiCad's own `Deserialize` refuses it. That is exactly how a 100%-broken `add_via` shipped in v0.2.1 (#117). I did this for #137 earlier today: launch pcbnew on a scratch board, drive the tool through the real server, then *read the item back* to confirm it exists — "no error" has not been proof of anything since #99. Happy to run it; I'd just rather it were a stated step than an assumption.

2. **`README.md` is missing from the count update.** You bump `registry.rs`, `tool-directory.md` and DEV.md, but the README still quotes the old total. Main now has a test for precisely this (`crates/konnect/tests/doc_tool_counts.rs`) — CONTRIBUTING has asked for those four to move together for a while and nothing enforced it, which is how the counts drifted before. It will fail against the current diff, so worth fixing when you rebase.

Also note main has moved a fair amount today — #136 (sym-lib-table resolution), #142/#153 (KiCad 10 read-by-shape), #144, #148, #179 and #137 all landed — so expect a rebase. `pcb_components.rs` and `library.rs` in particular have both changed.

This round was deliberately scoped to bug fixes and de-duplication; new-tool features go next. Nothing wrong with the PR, it's just not this round's shape.

Pytonballoon810:
Agreed on both counts. I've closed #159 — this contains it byte-for-byte, and the split was only ever an artifact of a fork branch not being able to base an upstream PR, so there's nothing there worth keeping open as a separate route. Everything continues here.

**On live verification** — that was done, and it's in the description under Tests: driven against KiCad 10 over its live IPC socket, with the polygon read back at its new coordinates after the write and the saved board checked on disk, where the only change beyond the edit was KiCad's own `(type default)` → `(type solid)` normalisation of the footprint it rewrote. I agree entirely that it should be a stated step rather than something a reviewer takes on trust — #117 is a fair thing to hold over 353 lines of outbound message construction, and "no error" proving nothing is the whole reason I read the item back rather than stopping at the response. What's missing is the evidence, not the run. I'll re-run after the rebase and post the transcript here: the calls, the read-back, and the diff of the board file. Please run it yourself as well if you'd rather not take mine.

The rebase makes that re-run matter more than it otherwise would. #142/#153 changed KiCad 10 read-by-shape, which is precisely the shape decoding both tools sit on, so this is not a mechanical rebase and green tests afterwards won't tell me much on their own.

**`README.md`** — a miss, not an ambiguity. Line 16 still says 187, which is 191 with both commits. The four move together on the rebase, checked against `doc_tool_counts.rs` before I push.

The Copilot passes across the two PRs caught five things worth taking — three in `library.rs`, two in the IPC half:

1. **`require_str` downgraded to `anyhow`** (`library.rs:1002`) — correct, and the kind of thing that only shows up at the client. Returning the `CallToolResult` directly.

2. **`fmt_mm`** — the scientific-notation part of that comment is wrong: Rust's `Display` for `f64` never uses exponent form, it prints the shortest decimal that round-trips, so `1e-300` comes out as a 300-digit number rather than as `1e-300`. The conclusion still holds, and for a worse reason than the one given — round-tripping the full `f64` is exactly how `0.30000000000000004` ends up written into a `.kicad_mod`. Switching to `format_at`'s `{:.6}` + trim, which has the side benefit of making the two agree.

3. **Empty-string UUID for items with no `(uuid …)`** (`library.rs:976`) — real, and reachable with a hand-written or pre-6.0 footprint. I'd rather not skip them silently, though: the point of the list tool is to be a faithful picture of the file, and an item that vanishes from it is harder to debug than one that says it cannot be edited. Proposing `uuid: null` plus an explicit `editable: false`, with `edit_footprint_graphic` rejecting an unknown UUID as it already does. Say if you'd prefer the plain skip and I'll do that instead.

4. **`geometry_points` fabricating `(0, 0)`** for a missing `Vector2` (`client.rs:82`) — agreed. It only bites if KiCad hands back a shape with a vertex absent, which should not happen, but an origin point that looks valid is the worst available way to report it. Dropping the missing vertices instead, so the arity is visibly wrong.

5. **Polygons with holes** — the sharpest of the five, and a data-loss path rather than a reporting one: the write rebuilds the shape as a single outline with `holes: vec![]`, and the read side carries the same assumption at `.polygons.first()`. A footprint with a cutout would read back looking correct and be silently flattened on the next edit. Rejecting multi-polygon and holed shapes by name on edit — same shape as the existing kind check — and fixing the read side to report what it is ignoring rather than dropping it quietly.

Next round is fine. I'll push the rebase with these folded in and the verification transcript alongside, so it arrives ready rather than needing another pass.

Pytonballoon810:
Rebased onto `main` at v0.3.0 and pushed, with the review points folded into the two commits rather than stacked on top — so each commit is independently green under `doc_tool_counts`, which I checked by building `fd76145` on its own. README is 189 at the library commit and 191 at this one. DEV.md and `tool-directory.md` were already right; README was the only one behind, as you said. One conflict on the rebase, in `tool-directory.md`, where main had rewritten `get_component_pads`' description under the two rows this adds.

`cargo fmt --check` clean, clippy clean on both crates, and the suite green **except two failures that are already on `main`** — see the last section.

## Live verification

Driven against KiCad 10 over the real IPC socket, through the release binary over stdio. Scratch board, one footprint placed at **(120, 90) rotated 90°**, carrying an `fp_poly` pin-1 marker and an `fp_line`.

**Read.** The polygon comes back at its footprint-local coordinates, identical to the `.kicad_mod` source, with the instance rotated 90° — which is the case where a sign error in the transform would be visible:

```json
{ "kind": "polygon", "layer": "F.SilkS", "editable": true, "outlines": 1, "holes": 0,
  "points": [ {"x":13.9,"y":-0.5}, {"x":13.9,"y":0.5}, {"x":13,"y":0} ],
  "uuid": "aaaaaaaa-0000-4000-8000-000000000004" }
```

The sibling `fp_line` reports `"kind":"segment", "editable":false, "outlines":0` — the new fields behave for non-polygons.

**Write, then read back.** Moved to `(11.4,-0.75) (11.4,0.75) (10.1,0)`; the subsequent list returns exactly those. **Refusals**, both by name and without writing:

```
graphic 'aaaa…0005' on 'U1' is a segment; only polygons take a vertex list
graphic 'deadbeef-…' not found on footprint 'U1' — call list_board_footprint_graphics for the UUIDs it does have
```

**On disk.** After `save_project`, diffing the board across a second edit — normalised file to normalised file, so KiCad's own rewrite is not in the picture — is a single hunk:

```diff
 		(fp_poly
 			(pts
-				(xy 12.9 -0.5) (xy 12.9 0.5) (xy 12 0)
+				(xy 11.4 -0.75) (xy 11.4 0.75) (xy 10.1 0)
 			)
```

**Library half**, against a real KiCad-saved `.kicad_mod`. Geometry and stroke width change; `fill`, stroke `type` and `uuid` survive, which is the block-surgery claim:

```diff
 	(fp_rect
-		(start -1.05 -1.05)
-		(end 1.05 1.05)
+		(start -1.2 -1.2)
+		(end 1.2 1.2)
 		(stroke
-			(width 0.05)
+			(width 0.06)
 			(type default)
 		)
 		(fill no)
```

A rejected edit left the target file byte-identical.

## What the review caught

1. **`require_str` downgrade** — fixed, and visible in the transcript: a missing `uuid` now returns `{"kind":"invalid_argument","field":"uuid"}` where it used to be a generic handler error, while an unknown UUID correctly stays a `handler_error`. Note this pattern is not unique to my line — `library.rs` has ~8 other `require_str(...).map_err(|e| anyhow!("{:?}", e))` sites that still downgrade. I left them alone to keep this diff honest; happy to send a separate PR if you want them swept.

2. **`fmt_mm`** — now `{:.6}` + trim, matching `format_at`. For the record the comment's stated reason is wrong: Rust's `Display` for `f64` never uses exponent form — `1e-300` prints as a 300-digit decimal, not `1e-300`. The real failure is duller and more likely: `{}` round-trips the full `f64`, so an accumulated coordinate lands in a `.kicad_mod` as `0.30000000000000004`.

3. **Graphics with no UUID** — now `uuid: null` + `editable: false` rather than `""`. Live checking turned this from an edge case into something worth your attention: **KiCad's own stock library is full of it.** 41 of the 60 `Package_SO.pretty` footprints I sampled carry no `(uuid …)` on any graphic — including `SOIC-8_3.9x4.9mm_P1.27mm`, whose seven silkscreen items all listed as `uuid: ""` before this change, every one looking addressable and none of them being so. Footprints saved by KiCad's own footprint editor do carry them (6 of 6 of mine). So UUID addressing is right for the footprints a user edits and simply cannot reach a large part of the stock library. That is a real limit of the design rather than a bug in it, but it should be stated somewhere the caller sees, which is what `editable` now does.

4. **`geometry_points` fabricating `(0,0)`** — fixed; absent vertices are dropped so the arity is visibly wrong.

5. **Polygons with holes** — the write path refuses multi-outline and holed shapes by name instead of flattening them, and the read path reports `outlines`/`holes` so the caller can see when `points` is not the whole shape. **Caveat:** this one is unit-tested only. I could not construct a holed polygon to drive live, because the `.kicad_mod`/`.kicad_pcb` `fp_poly` form has no syntax for holes — such a shape has to come from KiCad internally. If you know a way to produce one I will drive it.

## Two failures already on main

Both reproduce on a clean checkout of `40f2469` (v0.3.0) with no changes of mine:

- `tools::library::symbol_source_tests::a_schematic_with_no_project_falls_back_to_its_own_directory`
- `tools::sch_components::tests::add_component_writes_eeschema_style_instance_path`

The first resolves `${KIPRJMOD}` against the KiCad **install** directory, so the candidates come back as `C:\Program Files\KiCad\10.0\share\kicad\symbols\Loose.kicad_sym` instead of the table beside the schematic. Both look environment-dependent in the same direction: they pass on CI, which has no KiCad installed, and fail on a machine that does. Worth its own issue rather than this thread — say the word and I will open one with the detail.

mixelpixx:
Status update, and what this needs to land. I have closed #188 (its two library
fixes are on `main`; the footprint-graphics commits it also carried are the ones
here), so this is now the single home for the on-board work.

**This is still wanted, and it is not a duplicate of #205.** #205 landed
yesterday and edits footprint graphics in **library files**, with layer and mode
selectors. This PR edits a **placed footprint on the board, over IPC**. Those
are different operations on different files and both are worth having — the
overlap is in the count bookkeeping, not the feature.

Two things to do:

1. **Drop `fd76145`.** It is #159's library commit, still here byte-for-byte,
   and #159 is closed. Its territory is now covered by #205. Keeping it makes
   this PR conflict with `main` for no gain.

2. **Rebase the IPC-only half onto current `main`.** A fair amount moved
   yesterday: 12 PRs merged, counts went 188 -> 200 registered and 18 -> 19
   toolsets. `doc_tool_counts.rs` enforces those totals across README.md,
   DEV.md and tool-directory.md, so read the new numbers by summing `tool_count`
   in `router/registry.rs` rather than adjusting the old ones — hand-derived
   counts have been wrong every time this round.

Two things to check while you are in there, both from changes that landed after
you opened this:

- Board mutations now go through `attempt_ipc_write` (`pcb_board.rs:84`), which
  calls `ensure_board_is_active` before the operation and only falls back to a
  file edit on `IpcFailure::Unreachable`. That is #193's rule: a write must not
  silently land on whichever board KiCad happens to have in front. Please route
  the new writes through it rather than calling the client directly.
- `MAX_TOOLS_PER_TOOLSET` is 20 and several toolsets are at or near it now, so
  check where the new tools land before assuming `pcb_components` has room.

If the rebase is more than you want to take on, say so and I will do it — the
feature is worth having and I do not want it stalled on mechanical work.

mixelpixx:
Rebased this myself, as offered. Your commit is unchanged and still yours;
I added two on top. Force-pushed to your branch — `git fetch && git reset
--hard origin/feat/board-footprint-graphics` if you have local work.

**First: I owe you a retraction.** I asked you to route the writes through
`attempt_ipc_write` per #193. I checked properly this time, and I was wrong.
`attempt_ipc_write` appears **zero times** in `pcb_components.rs` — every
mutator in that file uses the local `ipc!` macro or a bare `with_ipc`, and
`ipc!` already calls `ensure_board_is_active` before the body runs. So your
code already satisfies both halves of #193: board identity checked before
mutating, and no file fallback that could race a live editor.
`attempt_ipc_write` exists for tools that *do* have an S-expression fallback to
gate, which these don't. Doing what I asked would have made this the only tool
in its own module using it, for no observable difference. Sorry for the
detour — that was me pattern-matching on a rule instead of reading the file.

**Dropped `fd76145`** as discussed. Before dropping I checked whether #205
really covers it, and it does not, entirely: `set_footprint_graphics` selects
by **layer** only, so `replace` replaces every graphic on that layer. Your
`edit_footprint_graphic` selected one graphic **by UUID**, which nothing on
main can do for library files. That capability is real and worth having — I've
filed it separately rather than resurrect 668 lines against a `library.rs` that
#205 rewrote, because git *would* have auto-merged it and quietly left two
parallel implementations in the tree.

**Counts.** The PR had 189 → 191; main had moved to 200. It's **202 registered
/ 208 total** now. Four more files quote the total than when you opened this
(`docs/TROUBLESHOOTING.md`, `packaging/metadata.json`, `plugin/plugin.json`,
`README.md:72`) — a guard added yesterday sweeps every `.md` and `.json`, so
they all move together now.

**What I added, and why.**

Your 8 tests are good — `polygon_extent_counts_outlines_and_holes` and the
rotation round-trip are the load-bearing ones, and the hand-injected hole case
is exactly right since `board_polygon` can't build one. But all 8 exercise pure
read helpers. The **write path had none**: the kind check, the `< 3 points`
guard, and the multi-outline/hole refusal all sat inside the loop that walks a
live footprint's items, so reaching any of them needed a running KiCad. Nothing
checked the property that actually matters — that a refused request leaves the
shape untouched, since the caller goes on to pack and send whatever it holds.

I extracted that into `replace_polygon_outline` and covered each refusal with
the shape asserted unchanged afterwards, plus the rebuild itself: closed
outline, anchored translation, and local +X rotating to board −Y at 90°.
Getting that last one backwards mirrors silkscreen into something that still
looks like artwork, which is the kind of wrong nobody spots in a diff.

Also filled in `geometry_kind`'s other three names. They aren't decoration —
your refusal message interpolates the name, so a wrong one sends the reader
looking for a shape they didn't pick.

One small fix in your handler: a malformed `points` bubbled `anyhow` via `?`
while every other argument beside it returned a structured `invalid_argument`.
It now goes through `parse_points`, which names the offending index — "each
point needs a numeric 'x'" doesn't help on a twelve-vertex courtyard.

**And one thing your PR surfaced that wasn't your bug.** Running the refusals
live, they came back as:

```
KiCAD must be running with the board loaded
(IPC error: a polygon needs at least 3 points, got 1)
```

KiCAD *was* running and *did* hold the board. The `ipc!` macro asserted a cause
it doesn't know for every failure, so your carefully-worded refusals arrived
underneath a false statement. `with_ipc_classified` already draws the
distinction properly, so the macro now uses it. They now read:

```
a polygon needs at least 3 points, got 1
graphic 'a37eee…' on 'U2' is a segment; only polygons take a vertex list
```

That's a behaviour change for five other tools too, noted in the commit.

**Verification.** Full gate green at **805 tests**, clippy and fmt clean. Each
of the six new guards neutered individually — all six caught, no vacuous test.

Live against KiCAD 10.0.5 on the `video` demo, U2 (18 graphics, one polygon):
listed the items and got the UUID, layer filter returned 9 of 18, replaced the
polygon's four vertices, then had **KiCad save the board and read the polygon
back out of the file** — `(-3,-3) (3,-3) (3,3) (-3,3)`, exactly what was sent.
All three refusals held live: 1-point, non-polygon (naming "segment"), and a
board KiCad isn't holding.

Merging once CI is green. Thank you for this — the `editable` flag and refusing
a multi-outline polygon by name instead of flattening it is the right call, and
it's the part I'd have been most tempted to skip.

PR #161, opened 2026-08-13
fix(export): write drill files into a directory, not a path named "drill.drl"

anyn99:
## Problem

`kicad-cli pcb export drill --output` takes a **directory** and writes the 
file named after the board (`MyBoard.drl`) inside it. Both callers passed
`<output_dir>/drill.drl` as if it were a filename, so KiCad created a *directory*
literally named `drill.drl` containing the real `.drl` (plus a `debug.log`).

The reported path therefore pointed at a directory. Anything that opened the
"drill file" — a fab-package consumer, a script, a user — got a directory-read
error instead of Excellon data, while the export itself reported success.

Repro:

```
export_manufacturing_package { board: …, output_dir: "out" }
# → out/drill.drl/  (a directory)
#   out/drill.drl/MyBoard.drl   ← the actual file, one level down
```

kicad-cli's own help documents the argument as a directory:

```
  -o, --output    Output directory [nargs=0..1] [default: ""]
```

## Changes

- `export_drill` now takes the output **directory**, creates it if missing, and
  passes it with a trailing separator so KiCad treats it as a directory.
- It returns the `.drl` file(s) actually produced, so callers report real paths.
  Collecting every `.drl` in the directory also covers `--excellon-separate-th`,
  where KiCad writes two files instead of one.
- `export_manufacturing_package` and `export_gerber` updated accordingly; the
  fab package now lists one entry per drill file instead of one bogus path.

## Verification

Against KiCad 10.0.4 on Windows, board with a via, a through-hole pad and an
NPTH hole:

```
kicad-cli pcb export drill --output <dir>/ mini.kicad_pcb
→ "Datei ... mini.drl wurde erstellt"          (mini.drl, inside the directory)

kicad-cli pcb export drill --excellon-separate-th --output <dir>/ mini2.kicad_pcb
→ mini2-PTH.drl, mini2-NPTH.drl                (both picked up by the new return value)
```

`cargo test -p konnect-core` — 233 passed.

### DIFF
--- a/crates/konnect-core/src/tools/cli.rs
+++ b/crates/konnect-core/src/tools/cli.rs
@@ -374,18 +374,45 @@ pub async fn export_gerber(cli: &str, pcb: &Path, output_dir: &Path) -> Result<(
     Ok(())
 }
 
-/// KiCAD 10: `pcb export drill --output <dir> <input>`
-pub async fn export_drill(cli: &str, pcb: &Path, output: &Path) -> Result<()> {
+/// KiCAD 10: `pcb export drill --output <dir>` writes Excellon file(s) *into*
+/// `output_dir`, named after the board (e.g. `MyBoard.drl`, or `-PTH`/`-NPTH`
+/// when separated). `--output` is therefore a DIRECTORY, not a file — passing
+/// a filename like `drill.drl` makes KiCAD create a directory by that name and
+/// hide the real `.drl` inside it. `output_dir` is created if missing and is
+/// passed with a trailing separator so KiCAD treats it as a directory. Returns
+/// the `.drl` file(s) produced.
+pub async fn export_drill(cli: &str, pcb: &Path, output_dir: &Path) -> Result<Vec<PathBuf>> {
+    tokio::fs::create_dir_all(output_dir)
+        .await
+        .with_context(|| format!("Cannot create drill output dir {}", output_dir.display()))?;
+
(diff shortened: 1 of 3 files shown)


mixelpixx:
Thanks — the bug is real and you diagnosed it correctly: `export_drill` passes `--output` straight through, and both call sites (`manufacturing.rs:144`, `pcb_export.rs:320`) hand it `output_dir.join("drill.drl")`, so kicad-cli creates a *directory* literally named `drill.drl` and buries the Excellon files inside it.

I'm taking #138 (@jctrouble) for the same bug, mainly on testability:

- #138 factors out `drill_args` specifically so the argument vector can be asserted without kicad-cli installed. For a bug whose entire symptom is "the wrong string was passed to a subprocess", a test that pins the argv is the whole point — and this PR has no tests.
- It writes into `gerber_dir` rather than the package root, which is where a fab expects the drill files to sit relative to the layers they belong to.
- It handles the empty-result case, warning instead of silently reporting an empty file list.

**What I'm taking from yours:** the trailing-separator handling. You're right to be defensive there — kicad-cli keys directory-vs-file off the trailing separator in some versions, and #138 doesn't do it. I'm grafting

```rust
if !dir_arg.ends_with(['/', '\']) { dir_arg.push(std::path::MAIN_SEPARATOR); }
```

into #138 with credit to you in the commit, and extending its `drill_args` test to cover it.

Closing this one as superseded — but the fix is landing, with your contribution in it.

PR #163, opened 2026-08-13
fix(library): write portable lib-table URIs (${KIPRJMOD} + forward slashes)

anyn99:
## Problem

`register_footprint_library` and `register_symbol_library` wrote the raw library
path into the lib-table. On Windows that means an absolute back-slash URI:

```
(lib (name "MyParts") (type "KiCad") (uri "E:\KiCAD\proj\MyParts.pretty") …)
```

KiCad does not reliably pick such entries up — in my case neither the project nor
the global entry loaded until I opened the library manager and re-added the
library by hand. An absolute path also breaks the moment the project is moved or
shared.

Notably, `expand_lib_uri` already *reads* `${KIPRJMOD}` entries — that is exactly
what KiCad itself writes for project-scoped libraries — but nothing in the
codebase ever *wrote* one. The read and write sides disagreed.

## Changes

- New `format_lib_uri`, used by both register handlers:
  - always writes forward slashes (both scopes), and
  - for project scope, expresses a library that lives inside the project
    directory relative to `${KIPRJMOD}`, matching KiCad's own writer.
- The resulting URI is returned in the tool response, so the caller can see what
  was registered.

Libraries outside the project directory stay absolute (just normalized to
forward slashes); global scope is never relativized.

## Tests

Four new tests, including a round-trip that asserts what `format_lib_uri` writes
resolves back through the existing `expand_lib_uri`:

```rust
let uri = format_lib_uri(&lib, Some(proj));          // ${KIPRJMOD}/MyParts.pretty
let resolved = expand_lib_uri(&uri, Some(proj));     // → the same directory
```

`cargo test -p konnect-core` — 237 passed.

## Note

Touches `library.rs`, so this may want rebasing after #141 lands. Happy to do
that if you'd like it in a different order.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -1198,19 +1198,20 @@ async fn handle_register_footprint_library(
         ));
     };
 
-    register_in_lib_table(
-        &table_path,
-        nickname,
-        lib_path.to_str().unwrap_or(""),
-        "KiCad",
-    )
-    .await?;
+    let table_dir = if scope == "global" {
+        None
+    } else {
+        table_path.parent()
+    };
+    let uri = format_lib_uri(&lib_path, table_dir);
+    register_in_lib_table(&table_path, nickname, &uri, "KiCad").await?;
 
     Ok(CallToolResult::text(
         serde_json::to_string(&json!({
             "success": true,
             "nickname": nickname,
             "scope": scope,
+            "uri": uri,
             "table": table_path.to_str().unwrap_or("")
         }))
         .unwrap(),
@@ -1281,19 +1282,20 @@ async fn handle_register_symbol_library(
         ));
     };
 
-    register_in_lib_table(
-        &table_path,
-        nickname,
-        lib_path.to_str().unwrap_or(""),
-        "KiCad",
-    )
-    .await?;
+    let table_dir = if scope == "global" {
+        None
+    } else {
+        table_path.parent()
+    };
+    let uri = format_lib_uri(&lib_path, table_dir);
+    register_in_lib_table(&table_path, nickname, &uri, "KiCad").await?;
 
     Ok(CallToolResult::text(
         serde_json::to_string(&json!({
             "success": true,
             "nickname": nickname,
             "scope": scope,
+            "uri": uri,
             "table": table_path.to_str().unwrap_or("")
         }))
         .unwrap(),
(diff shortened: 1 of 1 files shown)


anyn99:
Heads-up on overlap with #136, which I've been testing locally.

#136 refactors both register handlers onto new `lib_table_target` /
`project_relative_uri` helpers, and `project_relative_uri` already emits
`${KIPRJMOD}/…` with forward slashes for project-scoped libraries — the main
thing this PR set out to fix. That part of this PR is therefore superseded, and
#136's version is the better home for it (it sits next to the fp-lib-table
machinery it mirrors).

What #136 does *not* cover, and what this PR would still add: it returns the
library path unchanged in three cases, so on Windows those keep backslashes.

```rust
// lib_table_target
if scope == "global" {
    return Ok((global_table, lib_path.to_string_lossy().into_owned()));
}

// project_relative_uri
let absolute_uri = || lib_path.to_string_lossy().into_owned();
let (Ok(lib), Ok(dir)) = (canonicalize(lib_path), canonicalize(table_dir)) else {
    return absolute_uri();        // 1. either side not canonicalisable
};
match lib.strip_prefix(dir) {
    Ok(rel)  => format!("${{KIPRJMOD}}/{}", rel.to_string_lossy().replace('\', "/")),
    Err(_)   => absolute_uri(),   // 2. library outside the project directory
}
// 3. global scope, above
```

Case 1 is not just theoretical: `canonicalize` fails on a path that doesn't
exist yet, so registering a library before creating it — a normal ordering —
falls back to an absolute back-slash URI even though the library ends up inside
the project directory.

That matters because KiCad writes forward slashes on Windows too, and
`expand_lib_uri` on the read side is already slash-agnostic — so the written
form is the only place the two conventions can drift apart. In my case the
registered library did not load until I re-added it through the library manager;
normalising the separators is the cheap half of making the written URI match
what KiCad itself would have written.

Happy to do whichever you prefer:

- **Rebase this PR on top of #136** and shrink it to just the separator
  normalisation in those three fallbacks (a few lines plus tests), or
- **close this** and let #136 own it entirely, in which case the three cases
  above are worth a follow-up.

Either way #136 should land first — I'll rebase, not the other way round.

mixelpixx:
Thanks — this is a correct fix and the tests are the valuable part.

I'm closing it as superseded by #136 (@pauliuszaleckas), which rewrites the same two registration paths (`handle_register_footprint_library` and `handle_register_symbol_library`) as part of the larger sym-lib-table work for #105. Its `lib_table_target` + `project_relative_uri` do the same job — `${KIPRJMOD}/…` for project scope, absolute otherwise — so merging both would be redundant *and* conflicting, since they rewrite the same statements.

#136 additionally carries the empty-parent guard that this one lacks: `Path::new("board.kicad_pro").parent()` returns `Some("")` rather than `None`, which is how you end up with a corrupt `${KIPRJMOD}//abs/path`.

**Two things from here that #136 does *not* have, which I'm grafting in with credit to you:**

1. **Unconditional forward-slash normalisation of the non-relativizable path.** #136's `absolute_uri()` returns `lib_path.to_string_lossy()` with Windows backslashes intact — so a library outside the project still gets a non-portable URI. Your `uri_outside_the_project_stays_absolute_with_forward_slashes` and `global_scope_is_never_relativized` tests are coming along.
2. **The lexical `strip_prefix` fallback.** #136 falls back to absolute whenever `canonicalize` fails on either side; your lexical strip is what makes registration work for a library that doesn't exist on disk yet, which is a normal case. Your `strip_unc` handling of the Windows `\?\` verbatim prefix comes with it — #136 happens to be safe there only because it canonicalises both sides so the prefixes cancel, which won't hold once the lexical path is added.

Closing as superseded, with the substance of it landing inside #136.

PR #166, opened 2026-08-13
Fix JLCPCB database download and rebuild

### LINKED ISSUE #97: JLCPCB parts DB URL is gone (HTTP 404)
The hardcoded (https://github.com/mixelpixx/Konnect/blob/main/crates/konnect-core/src/tools/integration.rs#L262C16-L262C74) JLCPCB parts DB URL is gone, it throws HTTP 404: `https://bouni.github.io/kicad-jlcpcb-tools/jlcpcb_parts.db`
Are there any alternatives?

neusse:
## Summary

Fixes the broken `download_jlcpcb_database` tool by consuming the maintained chunked JLCPCB catalog feed and rebuilding it as a Konnect-owned SQLite cache.

Closes #97

## Approach

The previous hard-coded `jlcpcb_parts.db` URL now returns 404. A URL-only replacement is insufficient because the maintained feed is split across ZIP chunks and its FTS5 `parts` schema does not match Konnect's existing `components` queries.

This change:

- reads the feed's chunk-count sentinel and streams each archive chunk to disk;
- safely extracts the single upstream SQLite database;
- validates the required upstream columns before importing anything;
- builds a fresh `components` database with Konnect's existing stable query schema;
- converts the upstream tiered price string to its first/single-unit numeric price;
- validates the generated database with a nonzero row count and SQLite `quick_check`;
- replaces the configured/default cache only after generation succeeds.

Keeping the conversion at the download boundary avoids coupling every query tool to the upstream project's packaging and column names. KiCad's plugin settings API can provide plugin-owned storage, but it cannot discover this third-party feed.

## Compatibility and safety

No public API or configuration impact. Existing `output_path`, `jlcpcb_db_path`, `force`, search, lookup, alternatives, and statistics behavior remains available.

Downloads and conversion occur in a temporary sibling directory. If an existing cache is present, it is moved aside only during the final replacement; an installation failure restores it, and a successful installation removes the temporary backup. Invalid manifests, unsafe archives, schema drift, empty imports, and failed SQLite validation leave the active cache unchanged.

## Validation

- [x] `cargo test --workspace --locked --lib --tests` (238 konnect-core tests plus all workspace test targets passed; environment-dependent tests remained ignored as designed)
- [x] Live maintained-feed conversion: 3 chunks / 175,301,304 compressed bytes / 771,456 parts generated successfully
- [x] Viewer, plugin, packaging, and real-KiCad checks are not applicable; this changes only the JLCPCB database downloader/importer

## Review checklist

- [x] New names follow `docs/NAMING_CONVENTIONS.md`; there are no public renames.
- [x] The database mutation replaces the target only after validation and restores it if installation fails.
- [x] No IPC mutations are involved.
- [x] No tools were added or removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
+name = "arbitrary"
+version = "1.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
+dependencies = [
+ "derive_arbitrary",
+]
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -327,6 +336,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "be1e0bca6c3637f992fc1cc7cbc52a78c1ef6db076dbf1059c4323d6a2048376"
 
 [[package]]
+name = "derive_arbitrary"
+version = "1.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e567bd82dcff979e4b03460c307b3cdc9e96fde3d73bed1496d2bc75d9dd62a"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "diff"
 version = "0.1.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


mixelpixx:
Verified the feed against the live upstream before merging, since a hardcoded URL going stale is exactly what got us here (#97):

```
chunk_num_current_parts_fts5.txt        -> HTTP 200, body "3"
current-parts-fts5.db.zip.001           -> HTTP 200, 80,000,000 bytes
current-parts-fts5.db.zip.002           -> HTTP 200
current-parts-fts5.db.zip.003           -> HTTP 200
current-parts-fts5.db.zip.004           -> HTTP 404
```

Chunk count matches the sentinel exactly, and `last-modified` on chunk 001 is today — the feed is current, not an artifact someone left behind. I also pulled the first four bytes of chunk 001 and got `50 4b 03 04`, so the concatenate-then-`ZipArchive` approach is right: the chunks are a split of one zip, not separate archives.

The old single-file URL this replaces is confirmed gone, which is #97.

Two things I liked: the size cap (`JLCPCB_MAX_ARCHIVE_BYTES`) and the chunk-count range check mean a redirect to something unexpected fails fast rather than filling a disk. And the five tests are fully offline — no `await`, no `reqwest`, working from a locally seeded SQLite fixture — so this doesn't put a network dependency into CI. `download_jlcpcb_archive` itself stays untested, which is inherent and fine.

Merging. Closes #97.

PR #170, opened 2026-08-14
feat(sch): add the sch_bus toolset — buses, bus entries, pin fan-out

JYPochez:
Konnect could not draw a bus. SchematicBuilder already round-tripped bus and bus_entry nodes and knew their slot in KiCad's required element order, so a bus drawn in eeschema survived a Konnect edit, but nothing could create one. Any repeated multi-signal link had to be one wire per signal or bare labels.

sch_wiring is at 19 tools and the registry invariant caps a toolset at 20, so this lands as a new toolset rather than crowding that limit.

  - add_bus / batch_add_bus: bus segments. Geometrically identical to wires; KiCad distinguishes them by node name and the attached label.
  - add_bus_entry: the 45-degree tick bridging wire and bus. Not decoration — a wire and bus that merely touch are not connected without one.
  - connect_pins_to_bus: per pin, the stub, the entry, AND the member label, because bus membership is by name and a stub with no label joins nothing. Pin resolution goes through the owning unit, so multi-unit parts work.

Also adds format_bus/format_bus_entry to konnect-sexp.

Verified against KiCad rather than only unit-tested: two 4-pin connectors fanned onto one bus export a netlist with D0..D3 each carrying exactly its two pins, ERC 0 violations. Unit tests cover the node name, the signed size offset on an entry (wrong sign puts the tick past the bus, connecting nothing), and registration. Registry counts and tool-directory.md updated: 18 -> 19 toolsets, 187 -> 191 tools. 447 -> 450.


mixelpixx:

mixelpixx:
Merged main in and pushed to your branch. This is the **first new toolset since the registry was written**, so the count surface is wider than usual — 18 → 19 toolsets, 199 registered + 6 meta = 205 — and the toolset count appears in three places the tool count doesn't (README's headline, DEV.md's crate comment, tool-directory's overview). All moved together; `doc_tool_counts.rs` checks the tool half.

Verified the toolset actually wires up, through the real server rather than from the diff: `load_toolset("sch_bus")` exposes `add_bus`, `batch_add_bus`, `add_bus_entry` and `connect_pins_to_bus`, `add_bus` writes a `(bus …)` node, and `kicad-cli` loads the resulting sheet.

Gate green: fmt, clippy `--all-targets -D warnings`, 757 tests.

Issue #156, opened 2026-08-12
add_schematic_text writes a raw newline into the quoted string, leaving the schematic unloadable by KiCad


jhaubrich:
**Konnect version:** v0.2.2 (source build, commit 5cd6454)
**KiCad version:** 10.0.5 (Linux AppImage)
**File format:** `(version 20260306)`
**OS:** Debian 13 (trixie), x86_64

## Summary

`add_schematic_text` with a newline in the `text` argument emits the newline
raw inside the quoted S-expression atom. KiCad then refuses to load the file:
`Failed to load schematic`. The tool reports success, including the text and a
UUID, so nothing signals that the schematic has just been broken.

Related, but not duplicates: #55 is the same family (invalid S-expression
output) from `add_schematic_component`, and #42 produced the same symptom — a
schematic KiCAD refuses to load — from `batch_connect_to_net` writing a
`net_label` tag. This one is a quoting bug in a third tool: the text is written
verbatim into a quoted atom without escaping.

## Reproduction

Against a schematic created by `create_project` (two `Device:R` symbols placed
with `add_schematic_component`; no other edits):

```jsonc
// 1. single-line text — fine, file still loads
{"name": "add_schematic_text",
 "arguments": {"schematic": "repro.kicad_sch", "text": "single line note",
               "x": 100, "y": 150}}

// 2. same call with an embedded newline — file is unloadable afterwards
{"name": "add_schematic_text",
 "arguments": {"schematic": "repro.kicad_sch", "text": "line one\nline two",
               "x": 100, "y": 160}}
```

Reported for call 2:

```json
{"added": "line one\nline two", "rotation": 0.0, "size": 1.27,
 "uuid": "4536065c-9902-4eda-8651-d77ea382475c", "x": 100.0, "y": 160.0}
```

On disk:

```
(text "line one
line two" (at 100 160 0) ...)
```

Then:

```
$ kicad-cli sch export netlist --output out.net repro.kicad_sch
Failed to load schematic
```

## Bisect

Applying one call at a time to a pristine file and asking `kicad-cli` to load
it after each:

| after | loads? |
|---|---|
| baseline (untouched) | yes |
| `add_schematic_text` single line | yes |
| `add_schematic_text` embedded newline | **no** |

The single-line call is harmless; the newline is what breaks it.

## Expected

Either escape the newline when writing the atom, or reject multi-line input
with an error. Silently writing an unparseable file is the worst of the three
[...]

PR #171, opened 2026-08-14
fix(sch): add_schematic_text produced schematics KiCad could not open

JYPochez:
Two independent defects, either fatal at the file level:

  - The node was inserted at content.rfind(')'), placing it after the symbol instances and sheet_instances. KiCad 10 requires instances last. sch_wiring::insert_before_close already existed for this, with the ordering rule documented above it; this handler never used it.
  - Only backslash and quote were escaped, so a multi-line annotation was written with literal newline bytes inside the quoted string. KiCad's reader wants the two-character escape.

Either one yields "Failed to load schematic" with no indication of which element is at fault, while the tool reports success. Worse, once the file will not load, kicad-cli erc leaves any previous report in place, so a stale ERC result is what gets read next.

Make insert_before_close pub(crate) and use it; escape \r (dropped), \n and \t alongside the existing handling.

Tests: text lands before the first symbol instance and after lib_symbols; a multi-line string is written with \n escapes not raw bytes; quotes, backslashes and tabs are escaped. The first two fail against the old code. 444 -> 447.


mixelpixx:
Merged main into your branch. The conflict was interleaved across three test modules added in the same session (#148, #179 and this one), so rather than splice it by hand I took main's `sch_batch.rs` and re-applied your two production changes and your test module deliberately — safer than trusting a hand-resolved hunk in a file this contended.

Both defects verified on main before merging, and both are fatal at the file level exactly as you describe:
- `sch_batch.rs` `handle_add_schematic_text` spliced at `content.rfind(')')`, putting `(text …)` after the symbol instances and `sheet_instances`.
- The escape handled only backslash and quote, so a multi-line annotation went in with literal newline bytes inside the quoted string.

Using the existing `insert_before_close` — with the ordering rule already documented above it — rather than writing a new splice is the right instinct, and widening it to `pub(crate)` turned out to be more useful than you knew.

**I extended the fix to the two siblings you left behind.** `handle_batch_connect_to_net` and `handle_connect_passthrough` had the identical `rfind(')')` splice, so a net label or a stub wire landed after the instances and broke the file the same way. Both now go through `insert_before_close`.

Your point about the downstream effect is the reason I chased those rather than filing them: once the file will not load, `kicad-cli erc` leaves the previous report in place, so the next ERC read returns a stale result. A silent write failure that also silently poisons the verification step is worth closing out completely.

Added `insert_order_tests` to pin the ordering, with a companion test asserting the *old* splice point really does fall after the instances — otherwise the guard could quietly become vacuous if the fixture changed.

Full gate green: fmt, clippy, doctests, 601 tests, including all three of yours.

Closes #156. Merging once CI is green.

PR #172, opened 2026-08-14
fix(sch-editor): resolve symbol libraries through sym-lib-table

JYPochez:
resolve_lib_symbol never read sym-lib-table. It scanned a hardcoded list of install directories for a file named after the library nickname, which fails two ways, both total rather than partial.

Every user library was invisible: registered in the table but living anywhere else on disk, so add_schematic_component and replace_component failed for any part not shipped with KiCad. And KiCad's own libraries were invisible on any non-standard install — the macOS candidates are /Applications/KiCad, /usr/local/share/kicad and ~/Applications, so a bundle kept anywhere else makes find_symbol_dirs() return empty and *no* symbol resolves at all.

The scan also assumed nickname == filename, which KiCad never requires: 

Footprints already resolved through the lib-table (#61). Symbols did not.

sym-lib-table is now consulted first, with the directory scan kept as a fallback for setups with no table. The reader follows (type "Table") indirection depth-bounded, expands ${KIPRJMOD} against the table's own directory, expands exported environment variables, and expands user path variables from kicad_common.json — those are set in Preferences > Configure Paths, are not process environment variables, and are the normal way to write a portable table. ${KICAD*_SYMBOL_DIR} falls back to the install root recovered from the table's own location.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1109,6 +1109,7 @@ name = "konnect-schematic-editor"
 version = "0.2.2"
 dependencies = [
  "konnect-sexp",
+ "serde_json",
  "tempfile",
  "thiserror 1.0.69",
  "uuid",
--- a/crates/konnect-schematic-editor/Cargo.toml
+++ b/crates/konnect-schematic-editor/Cargo.toml
@@ -11,6 +11,7 @@ description = "Typed KiCAD schematic parser and manipulator with lossless round-
 thiserror.workspace = true
 uuid.workspace = true
 konnect-sexp.workspace = true  # For geometry (pin transforms, snap, T-junction detection)
+serde_json.workspace = true    # Reading KiCad's kicad_common.json path variables
 
 [dev-dependencies]
 tempfile = "3"
--- a/crates/konnect-schematic-editor/src/library.rs
+++ b/crates/konnect-schematic-editor/src/library.rs
@@ -11,7 +11,34 @@
 
 use crate::sexp::{parser, SexpNode};
 use crate::Schematic;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+
+/// Prefix a symbol block's outer name with its library, as eeschema expects
+/// in an embedded `lib_symbols` entry, and prefix any `(extends "PARENT")`
+/// to match.
+///
+/// Unit sub-symbols ("Name_0_1", "Name_1_1") deliberately stay UNPREFIXED:
+/// eeschema names only the outer symbol with the library prefix and refuses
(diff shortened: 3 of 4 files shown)


mixelpixx:
Thank you for this — the diagnosis is right and the `kicad_common.json` insight is genuinely valuable. I'm closing it in favour of #136, which covers the same bug more completely, but I want to be specific about why rather than just picking one.

**Where this one stops short of #105:**

`symbol_lib_path(nickname)` reads `kicad_config_dir()?.join("sym-lib-table")` — the **global** table only. No `project_dir` is threaded in, and `resolve_lib_symbol(lib_id)` keeps its one-argument signature, so `<project_dir>/sym-lib-table` is never opened. Issue #105's own reproduction registers at project scope, so **that case still fails after this patch**. The `${KIPRJMOD}` branch in `expand_lib_uri` can only ever expand against a table that isn't read, so it's dead code today.

Global-scope registration genuinely is fixed here, and `ensure_lib_symbol`/`symbol_unit_count` do inherit that through `resolve_lib_symbol_flattened` — so this is a real partial fix, not a no-op.

**The structural concern:** `crates/konnect-core/src/tools/mod.rs:633` holds a second, drifted `resolve_lib_symbol` that this doesn't touch, and the new `read_sym_lib_table`/`kicad_config_dir`/`expand_lib_uri` are third implementations of helpers that already exist in `konnect-core/src/tools/library.rs` (`read_lib_table_checked`, `global_sym_lib_table`, `resolve_symbol_lib_path` — the last already checks global *and* project). I understand this was partly forced by crate layering, since `konnect-schematic-editor` sits below `konnect-core` and can't depend on it. #136 solves that layering problem with a `SymbolLibrarySource` trait in the lower crate and the implementation injected from above, which lets it **delete** the duplicate rather than add to it.

One more thing worth flagging for whatever you do next: `read_sym_lib_table`'s `field()` is line-oriented, so a lib-table written across multiple lines — legal s-expression, and what a hand-edited table often looks like — is silently skipped.

**What I want from this PR, please:** `kicad_user_path_vars()` and the reasoning behind it. You're right that Preferences → Configure Paths variables live in `kicad_common.json` and are never process environment variables, and #136 does **not** handle them — `expand_lib_uri` falls through and returns `None` for a `${MY_LIB}/parts.kicad_sym` entry. That's a real gap in the PR I'm taking, and portable tables are exactly where it bites.

Would you re-submit just that piece as a small PR against `expand_lib_uri` in `crates/konnect-core/src/tools/library.rs` once #136 lands? I'd merge it. If you'd rather not, tell me and I'll port it with attribution to you in the commit — but I'd rather it carried your name on it.

PR #173, opened 2026-08-14
feat(sch): add batch_add_no_connect

JYPochez:
batch_delete_no_connect existed with no batch add — an asymmetry that only shows up when it bites. Marking one MCU's unused pins is routinely 15-20 flags, each a separate add_no_connect round trip; this design needed 23 for the CH32V203, the programmer header and the three USB sockets.

Mirrors the delete tool. A malformed entry is reported and skipped rather than aborting, so placing twenty flags fails as "nineteen landed and one is named" instead of losing the batch.

Tests: all flags land in one pass, one bad entry does not cost the good ones, empty list is a no-op. sch_wiring 19 -> 20 tools (its cap), 192 -> 193 total. 455 -> 458.


mixelpixx:

mixelpixx:
Merged main in and pushed to your branch. Two mechanical things: the test `ServerConfig` fixture needed `eager_toolsets` (a field main gained after you branched), and the counts land at 195 registered + 6 meta = 201.

Worth flagging: this takes `sch_wiring` to **exactly 20 tools, which is `MAX_TOOLS_PER_TOOLSET`**. The toolset is now full — the next tool that belongs there will need either a split or a cap change, and the registry test will catch it rather than letting it slide. Not a problem with this PR, just the thing to know before proposing the next wiring tool.

Gate green: fmt, clippy `--all-targets -D warnings`, 753 tests. Merging once CI confirms.

PR #175, opened 2026-08-14
feat(sch): add set_schematic_page

JYPochez:
Nothing could set the sheet size. create_schematic emits a blank sheet with no (paper ...) node at all, and there was no way to change one, so a design that outgrew A4 either kept content hanging off the frame or forced a hand edit of the .kicad_sch — the thing the tooling exists to prevent.

An undersized page is silent: content outside the frame still exports and still nets up, and ERC says nothing. The tool returns the chosen size in mm so a caller can check it against the layout extents.

Handles A0-A5, A-E, US Letter/Legal/Ledger; replaces an existing paper node or inserts one into the header ahead of any element; an unknown size is a structured InvalidArgument that leaves the file untouched.

Tests: replace without leaving the old size, insert into a blank sheet before the first element, portrait flag, unknown size writes nothing, and a table invariant that every entry is stored landscape since portrait swaps them. Registry and tool-directory updated: sch_components 18 -> 19, 191 -> 192 tools. 450 -> 455.


mixelpixx:

PR #177, opened 2026-08-14
feat(sch): add update_symbols_from_library

JYPochez:
Editing a symbol in its library had no effect on schematics already using it. add_schematic_component and replace_component both go through ensure_lib_symbol_in_schematic, which short-circuits when a definition with that lib_id is already embedded, so the schematic keeps its stale copy. Hit for real: widening the TM16xx body in the library left every placed instance rendering at the old width, and replace_component with the same lib_id did not refresh it. eeschema has Tools > Update Symbols from Library for this; Konnect had nothing, leaving a manual GUI step as the only route.

The new tool re-resolves each embedded lib_symbols entry from disk and replaces it, with an optional lib_id filter and dry_run.

Wires and labels attach at pin coordinates, so a library edit that moved a pin would silently orphan them. The tool therefore compares pin anchors before and after and refuses any symbol whose pins moved, reporting why, unless allow_pin_moves is passed. Body size and pin length changes leave anchors untouched — the safe case, and the one that motivated this.

### DIFF
--- a/crates/konnect-core/src/router/registry.rs
+++ b/crates/konnect-core/src/router/registry.rs
@@ -28,7 +28,7 @@ pub static ALL_TOOLSETS: &[ToolsetMeta] = &[
         name: "sch_components",
         description: "Add, edit, move, rotate, and delete schematic symbols",
         category: "schematic",
-        tool_count: 17,
+        tool_count: 18,
     },
     ToolsetMeta {
         name: "sch_wiring",
--- a/crates/konnect-core/src/tools/sch_components.rs
+++ b/crates/konnect-core/src/tools/sch_components.rs
@@ -274,6 +274,27 @@ pub fn tools() -> Vec<ToolDef> {
             |args, ctx| async move { handle_replace_component(args, ctx).await }
         ),
         tool!(
+            "update_symbols_from_library",
+            "Refresh the schematic's embedded lib_symbols definitions from the on-disk \
+             libraries — the equivalent of eeschema's Tools > Update Symbols from Library. \
+             Use after editing a symbol in its library: placing or replacing a component \
+             reuses the copy already embedded in the schematic, so library edits are \
+             otherwise invisible. Refuses any symbol whose pins moved, because existing \
+             wires and labels sit at the old pin coordinates and would be silently \
+             orphaned; pass allow_pin_moves to override and re-check connectivity after.",
+            json!({
+                "type": "object",
+                "properties": {
+                    "schematic": { "type": "string", "description": "Path to .kicad_sch file" },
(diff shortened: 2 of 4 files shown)


pauliuszaleckas:
Same issue already fixed by my PR https://github.com/mixelpixx/Konnect/pull/150

mixelpixx:
Thanks for this — the motivating case is real and the safety instinct behind it is the best idea in either version of this tool.

#150 (@pauliuszaleckas) implements the same tool and arrived first, so I'm taking that one and grafting your pin guard into it. Concretely, why:

**1. The block scanner isn't string-aware.** `lib_symbol_blocks` counts parens without skipping quoted strings, so a stock KiCad symbol carrying `(property "Description" "Dual Op Amp (DIP-8)")` unbalances it. Parenthesised descriptions are common in the shipped libraries. The repo already owns the correct helpers for this — `find_block_starts` / `find_balanced_block` / `find_enclosing_block` — and #150 reuses `konnect_sexp::writer::find_direct_child_blocks`, which is indentation-agnostic and string-aware.

**2. `regex_lite_pin_at` duplicates `extract_lib_pins` / `extract_lib_pins_for_unit`**, which already exist and are unit-aware.

**3. The idempotence check never reports "unchanged".** `fresh.trim() == embedded.trim()` compares a tab-indented embedded copy against the unindented text `resolve_lib_symbol_flattened` returns, so every symbol always compares as changed, is always rewritten, and is written back **de-indented**. The tool churns the file on every call and `unchanged` is effectively unreachable. The tests don't exercise a second run, so this is invisible to the suite.

**4. The pin-anchor comparison is textual.** `pin_anchors` builds `Vec<(String, String, String)>` from the coordinate *tokens*, so `1.27` vs `1.270`, or `-0` vs `0`, compare unequal and the "pins moved, refusing" guard fires on formatting-only differences. It also sorts anchors without keeping pin numbers attached, so two pins swapping positions compares equal — the guard misses the one case it most needs to catch.

Credit where it's due: `read_consistent` / `write_atomic_if_unchanged` / `apply_edits` and `dry_run` are all used correctly here, and reusing `resolve_lib_symbol_flattened` is right.

**What I'm taking from you:** the pin-anchor guard and `allow_pin_moves`. #150 will happily move pins out from under existing wires and labels with only a sentence in the description as warning — that's a real safety hole and your PR is the one that spotted it. I'll reimplement the comparison structurally over the parsed trees (which #150 already has at hand) rather than porting the string version, default `allow_pin_moves` to false, and report skipped symbols with before/after pin positions the way you do. You'll be credited in the commit.

This tool is deferred to the next round along with the other new-tool work, so it won't land immediately.

PR #178, opened 2026-08-14
feat(project): add rename_project

JYPochez:
Renaming a KiCad project by hand is a four-file rename plus an edit that is easy to miss: every symbol instance stores (project "name"), and the sheet instance path is keyed on it. Rename only the files and KiCad loads the design as unannotated — every reference designator is gone, and re-annotating renumbers parts that a PCB is already laid out against.

rename_project renames the .kicad_pro/.kicad_sch/.kicad_pcb/.kicad_prl set and rewrites the internal references together, the equivalent of eeschema's File > Save As. dry_run reports the planned moves and rewrites without touching anything.


mixelpixx:
Reviewed. The problem is real and worth solving — renaming a project by hand genuinely is a four-file rename plus an edit that is easy to miss, and getting it wrong de-annotates the design. But I can't merge this as written, and the reason is the same class of bug the rest of your series fixes.

**The blocker** is `project.rs`:

```rust
let updated = text.replace(&old_name, &new_name);
```

That is an unbounded, non-word-bounded, case-sensitive global replace over the entire body of the `.kicad_pro`, `.kicad_sch`, `.kicad_pcb` and `.kicad_prl`. Your own description names the actual target — `(project "NAME")` in the symbol instances — but the code doesn't scope to it. A project named `R`, `C`, `led`, `usb`, `power`, or anything hex-ish rewrites `lib_id`s, net names, footprint names, property values and UUID substrings. That is worse than the corruption #179 fixes, because it is silent and total.

Three more, each independently blocking for a tool that renames files:

1. **No rollback.** The files are `fs::rename`d first; a `?` on the subsequent `read_to_string`/`write_atomic` leaves the project half-renamed, with no way back.
2. **Hierarchical designs break.** Only `<name>.kicad_sch` is processed. Sub-sheet files keep `(project "OLD")` in their symbol instances — which is exactly the de-annotation failure this tool exists to prevent, just moved one level down.
3. **Zero tests.** `git diff` shows 167 added lines and no `#[test]`. For a destructive, multi-file, non-atomic operation that is the one place tests are non-negotiable.

**What I'd merge:** scope the rewrite to the nodes that actually carry the name — parse and target `(project "…")` specifically, or at minimum match the full quoted token rather than a bare substring. Walk the sheet hierarchy so every reachable `.kicad_sch` is updated. Do the content rewrites *before* the renames, or stage into a transaction so a failure leaves the project untouched — note `konnect-sexp`'s transaction layer (#122) now exists on main and `commit_file_transaction` handles exactly this multi-file all-or-nothing case; #168 uses it for the same reason. And tests covering a short name that appears inside other tokens, a hierarchical project, and a mid-operation failure.

Happy to look again when you've had a go. The rest of your series landed — #174, #171 and #179 are all on main.

mixelpixx:
Re-reviewed after your rework, and this is now good. Three of the four things I raised are properly fixed, and I finished the fourth on your branch so it can land.

**The blocker is genuinely gone.** `rewrite_project_references` matching only quoted, structurally-meaningful occurrences — and splitting `.kicad_sch`/`.kicad_pcb` (only `(project "NAME"`) from `.kicad_pro`/`.kicad_prl` (whole-string and filenames) — is exactly the right shape. Your `led` test is the one I'd have written. **Rollback on partial rename** and the **five tests** cover the other two.

**What I added: child sheets.** The rewrite only visited the four files named after the project, but a hierarchical design keeps each sheet in its own file and every one stores `(project "NAME"` on its instances. KiCad's own `complex_hierarchy` demo has **46 of them in `ampli_ht.kicad_sch`** — a sheet that isn't named after the project and so is never renamed. Rewriting only the root left those pointing at the old name, which is the exact de-annotation this tool exists to prevent, just relocated from the root sheet to the children.

The rewrite pass now covers the renamed set plus every other `.kicad_sch` in the project directory. Child sheets are rewritten **in place and not renamed** — their names are referenced by `(property "Sheetfile" …)` entries. `dry_run` previews them the same way. Two tests, both failing with the extension neutered.

**Verified end to end against that demo**, which turned out to make the case for your structural rewrite better than any synthetic fixture could:

```
child sheet before: 46 project refs, 58 lib_id refs
content_rewrites: [(renamed_amp.kicad_pro, 1), (renamed_amp.kicad_sch, 29), (ampli_ht.kicad_sch, 46)]
child: 46 project refs renamed, all 58 lib_id refs untouched, filename kept
KiCad loaded the renamed hierarchy; annotation violations: 0
```

Note the second column. **That project's symbol library is also called `complex_hierarchy`**, so the child sheet carries 58 `complex_hierarchy:SYMBOL` `lib_id` references alongside its 46 project references. The old blind `text.replace` would have rewritten all 104 and broken every symbol reference in the design. Yours rewrites exactly the 46 that are project keys. That is the bug in miniature, on a real KiCad file.

Counts moved together: `project` 6 → 7, 190 registered + 6 meta = 196. Gate green — fmt, clippy `--all-targets -D warnings`, 689 tests.

Merging once CI confirms. Thanks for turning this around quickly and for taking the criticism structurally rather than patching the symptom — the `.kicad_sch` vs `.kicad_pro` split in particular is a distinction the original didn't have at all.

PR #179, opened 2026-08-14
fix(sch): operate on every unit of a multi-unit symbol, not just the first

JYPochez:
The description leads with the bug that actually costs you data — batch_connect_to_net resolving every pin against instance #1, so two nets land on one coordinate and are shorted with no error, no warning, and a success return. It names the real case (X_CLK_IN and X_DATA_IN both on U6 pin 1) and notes that ERC can't catch it either, since as far as the file is concerned the labels genuinely are on the same point. Then the three visible-but-wrong tools, the approach, and the seven tests.


mixelpixx:
Merged main into your branch. This is the fix I most wanted from the ten, and it needed a careful resolve — worth explaining, because a mechanical one would have quietly broken it.

**The bug is confirmed, and it is data corruption.** Verified on main before the merge: `sch_batch.rs:370` took the *first* instance carrying the reference, and `:383` used unit-blind `extract_lib_pins`. A multi-unit part is placed once per unit, all sharing the designator, so "first" is always unit 1 — and a pin owned by unit N was then transformed by unit 1's placement. On a 74HC14 where unit 1 pin 1 and unit 2 pin 3 both sit at local `(-7.62, 0)`, both resolve to the same absolute coordinate, two `format_net_label` calls land on one point, and that is a short. `errors` stays empty, `added_count` says 2, and ERC cannot flag it because as far as the file is concerned the labels genuinely are coincident.

It's also drift rather than an oversight, which makes it worse: the singular path already got this right in #35 — `sch_wiring.rs:1631` carries the comment "Unit-aware (#35): only this instance's unit owns the pin" and calls `extract_lib_pins_for_unit`. `sch_batch`'s copy never received that fix.

**The resolve.** #144 landed just ahead of this and replaced every `lib_symbols` lookup with `find_lib_symbol()`, which resolves by `(lib_name …)` when the instance carries one and falls back to `lib_id`. Your replacement closure looks the symbol up by `lib_id` directly — so keeping your side verbatim would have reintroduced #143 on precisely the multi-unit path this fixes: a sheet-local derived symbol resolving to the base definition, whose pins can sit at different coordinates, or to nothing when the base was never embedded. Both now hold: the per-unit fan-out stays, and each candidate unit resolves through `find_lib_symbol`.

Gate green: fmt, clippy, 572 tests, including all seven of yours — `each_unit_resolves_its_own_pin_position`, `field_edit_reaches_every_unit` and `delete_removes_every_unit` are the ones that fail against the old code.

`find_all_symbol_instance_blocks` keeping `find_symbol_instance_block` as `.into_iter().next()` is the right call — every existing single-placement caller is behaviourally unchanged.

**Two things for the record, neither blocking:**

1. **`bulk_move` changes its response shape** — `moved[]` goes from `{reference, old_x, old_y, new_x, new_y}` to `{reference, units, placements:[…]}`. I checked and no in-repo consumer breaks (skills, tests, `tool-directory.md`), but it is an undocumented API change and I'd rather it were stated than discovered.
2. **This is partial, which is fine** — still unit-blind after it: `sch_batch.rs` validate handlers, `sch_export.rs:223,341`, `sch_analysis.rs` (five sites), `design_review.rs` (five sites), and the single-symbol delete/move/rotate/edit in `sch_components.rs`. I'll file that as a follow-up rather than hold this.

Merging once CI is green.

PR #180, opened 2026-08-14
Make main the VNS working version: current upstream plus ten fixes

JYPochez:
Brings this fork's `main` up to current upstream and makes it the VNS working version:
**`mixelpixx/Konnect@5cd6454` plus ten fixes**, merged and gated together.

Until now `main` here was a byte-identical mirror of upstream, and the fork's actual work
lived on `publish` — a branch built from a vendored v0.2.0 tarball with **no shared ancestry**
with upstream. That made it useless as a base: a diff against upstream showed 208 files and
6060 deletions, because git could only read it as reverting everything upstream had done
since. Every fix here has now been replayed onto real upstream history instead.

## Correctness fixes

**Multi-unit symbols were treated as one placement.** A multi-unit part is one `(symbol …)`
instance *per unit*, all sharing the reference, and four tools took the first match.
`batch_connect_to_net` resolved every pin against instance #1 and transformed it by that
unit's placement, so two different nets landed on one coordinate and were **silently
shorted** — no error, and the tool returned success. Found in real use: `X_CLK_IN` and
`X_DATA_IN` both on U6 pin 1 of a 74HC14. `batch_edit_schematic_components` wrote fields into
unit 1 only; the delete tools left the other units as orphans; `bulk_move` tore the part
apart.

**No user symbol library resolved.** `resolve_lib_symbol` scanned a hardcoded list of install
directories for a file named after the library nickname and never read `sym-lib-table`. On any
non-standard KiCad install — a macOS bundle outside `/Applications`, a portable Windows copy —
*nothing* resolved. Footprints got table-aware lookup in v0.2.1; symbols never did.

**`add_schematic_text` wrote files KiCad could not open.** It inserted the node after the
symbol instances (KiCad 10 requires those last) and wrote literal newlines where the format
wants `\n`. Either alone makes the schematic fail to load with no indication of which element
is at fault, while the tool reports success. Worse, once the file will not load `kicad-cli
sch erc` leaves the previous report in place — so a *stale* ERC result is what gets read next.

**String escapes decoded wrongly.** `unescape` collapsed backslashes last, so `C:\\new` came
back as `C:\` plus a real newline. Hits Windows paths in `Datasheet`, 3D-model references and
`Sheetfile`.

## Added tooling

- **`sch_bus`** — `add_bus`, `batch_add_bus`, `add_bus_entry`, `connect_pins_to_bus`.
  `konnect-sexp` already had `format_bus` and `BusEntryDirection` and `SchematicBuilder`
  already round-tripped bus nodes, but nothing could create one. `connect_pins_to_bus` writes
  the stub, the entry *and* the member label per pin, because bus membership in KiCad is by
  name — a stub without a label joins nothing.
- **`set_schematic_page`** — nothing could set the sheet size, and content off the frame still
  exports and still nets up, so an undersized page is a silent defect.
- **`update_symbols_from_library`** — eeschema's *Update Symbols from Library*. Refuses any
  symbol whose pin anchors moved, since wires sit at the old coordinates.
- **`rename_project`** — renames the four project files *and* the internal `(project "name")`
  references. Renaming files alone makes KiCad treat the design as unannotated, losing every
  reference designator.
- **`batch_add_no_connect`** — `batch_delete_no_connect` existed with no batch add; one MCU's
  unused pins is routinely 15–20 round trips.
- **`reload_server`** — `exec`s into the rebuilt binary in place, same PID and stdio pipes, so
  the MCP session survives a rebuild. Verifies the new binary first.

## Validation

`cargo fmt --all -- --check`, `cargo clippy --workspace -- -D warnings`, and
`cargo test --workspace --lib --tests` — **515 passing**. `cargo check --locked
--manifest-path crates/schematic-viewer/Cargo.toml` passes too, which the viewer CI job
requires.

Counts reconciled across the merge: 19 toolsets, 195 registered tools + 7 meta-tools.
`sch_components` 17 → 19, `sch_wiring` 19 → 20, `project` 6 → 7, and the `tools_added`
assertion in `protocol_stdio.rs` 36 → 39.

Three merge hazards were resolved rather than papered over: two pairs of branches appended
test modules at the same point in `sch_batch.rs` and `sch_components.rs`, and every branch
claimed the same next `tool_count`.

## Relationship to the upstream PRs

Each fix is also offered upstream on its own branch, one per PR (`up/*`), each based directly
on `upstream/main` and carrying only its own change. This branch is the integrated result for
use here; those are the reviewable units. As they land upstream, rebuild this branch on the
new `upstream/main` rather than merging it back.

Not included, and still fork-only: `publish-to-fork.sh`, `version_history.md`, `CLAUDE.md`,
`docs/CODE_REVIEW.md` and the KiCad bug reports.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1109,6 +1109,7 @@ name = "konnect-schematic-editor"
 version = "0.2.2"
 dependencies = [
  "konnect-sexp",
+ "serde_json",
  "tempfile",
  "thiserror 1.0.69",
  "uuid",
--- a/DEV.md
+++ b/DEV.md
@@ -227,9 +227,9 @@ Source: [`crates/konnect-core/src/observability.rs`](crates/konnect-core/src/obs
 
 ## Tool Routing (Starter Kit + On-Demand Loading)
 
-The server does NOT expose all 187 tools (193 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
+The server does NOT expose all 187 tools (194 total with the 7 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
 
-- **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 6 meta-tools, baseline `tools/list` is ~19 tools ≈ 2K tokens.
+- **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 7 meta-tools, baseline `tools/list` is ~20 tools ≈ 2K tokens.
 - **On demand**: the LLM reads `list_toolboxes` → calls `load_toolset(name)` to expose a toolset's tools in subsequent `tools/list` responses. `unload_toolset(name)` prunes them when the task shifts.
 - **`tools/list_changed` notification**: sent on every load/unload so MCP clients refresh their local tool cache.
 - **Error recovery**: if the LLM calls an unloaded tool, `handler.rs` returns an actionable error naming the toolset that owns it (so the LLM can load it and retry in one hop — no extra `list_toolboxes` round-trip).
@@ -289,9 +289,9 @@ convention for other `kicad-cli`-calling code.
 
 ## Current Stats
 
-- **18 toolsets, 187 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
+- **18 toolsets, 187 tools** + 7 meta-tools (4 routing + 2 observability + 1 maintenance — see `tool-directory.md`)
 - Baseline `tools/list`: ~19 tools / ~2K tokens (starter kit + meta-tools)
-- Full-catalog `tools/list` (all loaded): 193 tools (187 registered + 6 meta) / ~25K tokens
+- Full-catalog `tools/list` (all loaded): 194 tools (187 registered + 7 meta) / ~25K tokens
 - **0 IPC stubs** (all protobuf methods implemented)
 - **0 unimplemented tools**
 - **3 CLI commands removed in KiCAD v10** (specctra DSN/SES, pcb sync — return clear errors)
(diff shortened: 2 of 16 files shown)


pauliuszaleckas:
I guess this PR was created by accident?

mixelpixx:
Thanks for putting this together, and for taking the trouble to replay the fork's work onto real upstream history rather than leaving it on a vendored branch — that part was the right call and it made this reviewable at all.

I'm going to take the ten individual PRs (#170–#179) instead of this bundle, for a concrete reason rather than a policy one.

I verified this branch **is** the exact union of the ten: every single-owner file diff is byte-identical to the corresponding PR, and the multi-owner files reconcile line-for-line. So nothing is missing. The problem is the history it would bring with it:

1. **Three intermediate commits carry literal conflict markers.** `df2deb7df`, `8503c8717` and `febd5c077` all contain `<<<<<<< HEAD` / `=======` / `>>>>>>> up/update-symbols` in `crates/konnect-core/src/tools/sch_components.rs`. Those commits cannot compile. CI is green because it only builds the head, so this doesn't show up as a failure — but it would poison `git bisect` permanently, and `bisect` is how we found the `add_via` regression in v0.2.1.
2. **A duplicated attribute survives the cleanup** — `sch_components.rs:2209-2210` has `#[cfg(test)]` twice over `mod page_tests`. Legal Rust, so it compiles, but it's residue from the conflict resolution.
3. **The reconcile commit contradicts itself.** `53e6134f0` takes `tool-directory.md` to "19 toolsets, 195 registered + 7 meta = 202", but `DEV.md` still says "18 toolsets, 187 tools ... 194 total" and `DEV.md:55` still says "6 always-visible meta-tools", while `README.md:16` is untouched at "187 tools across 18 on-demand toolsets". That's the drift CONTRIBUTING asks contributors to avoid, landing in the commit whose job was to fix it.

The ten separate PRs don't have any of these problems — they're each individually correct against main, which is exactly the "short PR series" shape CONTRIBUTING asks for. I'd rather merge them one at a time and keep a history that builds at every commit.

I've reviewed all ten and will comment on each with its own verdict. Short version: #174 and #171 are clean merges, #179 is a genuine data-corruption fix I want, #170/#173/#175 need small changes, and #172/#176/#177/#178 need rework or are superseded — details on each PR.

Closing this one in favour of the series. None of the work is lost.

JYPochez:
oups yes ! was meant to be local to my repo ! really sorry !

Issue #182, opened 2026-08-14
sch_batch, sch_export, sch_analysis and design_review still resolve pins against unit 1


mixelpixx:
Follow-up from #179, which fixed the multi-unit pin resolution in `handle_batch_connect_to_net` (the one that silently shorted two nets onto one coordinate). That fix was deliberately scoped; several other call sites still take the first instance carrying a reference, or use unit-blind `extract_lib_pins`.

A multi-unit part is placed once **per unit**, all sharing the designator, so "the first instance" is always unit 1 — and a pin owned by unit N transformed by unit 1's placement lands somewhere it doesn't belong.

Still unit-blind on main after #179 and #148:

- `crates/konnect-core/src/tools/sch_batch.rs` — `handle_validate_wire_connections`, `handle_validate_component_connections`
- `crates/konnect-core/src/tools/sch_export.rs` — two sites
- `crates/konnect-core/src/tools/sch_analysis.rs` — five sites
- `crates/konnect-core/src/tools/design_review.rs` — five sites
- `crates/konnect-core/src/tools/sch_components.rs` — the single-symbol delete/move/rotate/edit paths

Severity differs by site. The validators and analysis tools produce **wrong reports** rather than wrong files, which is bad but recoverable. The `sch_components` mutators are the ones to look at first: `delete` leaves the other units behind as orphans, and `move`/`rotate` tear a multi-unit part apart. #179 fixed the batch equivalents of those, so the shapes to copy already exist — `find_all_symbol_instance_blocks` (`tools/mod.rs`) and `extract_lib_pins_for_unit`.

The correct pattern is established in three places now: `sch_wiring.rs::resolve_placed_pin` (unit-aware, and after #148 also `lib_name`-aware via `find_lib_symbol`), and `tools/mod.rs::placed_pins`.

## Acceptance

- [ ] Each site either resolves per-unit or has a comment saying why unit 1 is correct for it.
- [ ] Regression tests using a two-unit fixture where both units own a pin at the same *local* coordinate — that is the shape that makes the bug visible.

mixelpixx:
Adding a call site found while fixing #203: `handle_add_component_annotation` (`sch_components.rs`) uses `find_symbol_instance_block` (singular, first block in file order), so an annotation on a multi-unit part lands only in its first unit block. Same unit-blind shape as the rest of this issue's list.

Noted by @triglav-modular in #203, which is otherwise fixed on `main`.

mixelpixx:
[...]

PR #185, opened 2026-08-14
fix(review): fail closed on incomplete coverage

### LINKED ISSUE #184: [BUG] Make run_design_review fail closed when extraction or audit coverage is incomplete
## Summary

`run_design_review` can currently return:

> `LOOKS GOOD — no critical issues found`

when an audit evaluated no usable design objects or when its result was not successfully extracted. A clean review and an incomplete review are therefore indistinguishable.

This issue proposes making the consolidated review **fail closed** whenever extraction or audit coverage is incomplete. I am opening it before implementation, per `CONTRIBUTING.md`, so the result contract and scope can be agreed first.

## Evidence on current `main`

In `crates/konnect-core/src/tools/design_review.rs`:

1. `handle_run_design_review` runs the individual audits and passes each result to `extract_findings`.
2. `extract_findings` returns an empty vector unless the first content block is valid JSON containing a `findings` array. Any other result shape is silently converted to “no findings.”
3. Several audits silently skip symbols whose embedded library definition cannot be resolved:

   ```rust
   let lib_sym = match find_lib_symbol(&lib_syms, inst) {
       Some(s) => s,
       None => continue,
   };
   ```

4. The final verdict depends only on finding counts. If both error and warning counts are zero, the result is always `LOOKS GOOD`.
5. The consolidated response contains no audit completion status or coverage counts—no sheet, symbol, resolved-symbol, footprint, pad, or net totals—so a caller cannot tell whether the whole design was actually inspected.

A populated hierarchical or unusual-symbol project can therefore appear clean when part of the design was skipped or reduced to zero extracted objects.

## Relationship to #182

[#182](https://github.com/mixelpixx/Konnect/issues/182) identifies specific unit-1 assumptions in `design_review` and other schematic paths. That work is necessary, but narrower: fixing those call sites does not give the review aggregator a general way to detect future parser, hierarchy, library-resolution, or audit-result failures.

This issue is about the **review completion and coverage contract**, not duplicating #182's multi-unit fixes.

## Proposed direction

Give each audit a structured result envelope, for example:

- `status`: `complete`, `partial`, or `failed`;
- `findings`;
- `coverage`: relevant source and evaluated counts;
- `diagnostics`: skipped/unresolved objects and parser or backend failures.

The consolidated review should then:

1. require every requested audit to report a recognized status;
2. surface unresolved/skipped objects instead of silently continuing;
3. treat suspicious extraction gaps as incomplete, not clean;
4. include source/evaluated counts in the response;
5. never emit `LOOKS GOOD` unless all requested audits completed with adequate coverage.

Useful schematic coverage includes root/child sheet counts, symbol instances, resolved symbol definitions, and named nets. If a board is supplied, include footprints, pads, and nets.

## Questions to agree before implementation

1. Should incomplete coverage return a structured tool error, or a third verdict such as `INCOMPLETE — review could not evaluate the full design` with diagnostics?
2. Should native `kicad-cli` count comparison happen on every review, only when counts look suspicious, or only in the weekly E2E tests?
3. Which zero counts are legitimate, and which should require explicit confirmation from the caller?
4. Should individual audit tools adopt the same envelope in this change, or should `run_design_review` adapt their existing responses first to keep the PR smaller?

## Acceptance criteria

- [ ] Every audit requested by `run_design_review` reports `complete`, `partial`, or `failed`.
- [ ] A malformed, error-shaped, or missing-`findings` audit result cannot be converted into an empty successful audit.
- [ ] The consolidated result includes source and evaluated counts sufficient to understand coverage.
- [ ] Unresolved library symbols and other skipped design objects are reported.
- [ ] A populated hierarchical fixture with components on a child sheet cannot receive `LOOKS GOOD` if child-sheet extraction is absent or incomplete.
- [ ] A regression test proves that incomplete extraction produces the agreed structured error/verdict and preserves diagnostics.
- [ ] A genuinely complete design with zero findings can still receive `LOOKS GOOD`.

## Validation

The implementation should include focused handler tests plus a KiCad-backed E2E fixture where available. The PR should follow the repository's standard locked tests, documentation tests, Clippy, and formatting checks.

neusse:
Fixes #184.

## Problem

`run_design_review` audited only the root schematic and based its verdict solely on returned findings. A hierarchical design could therefore receive `LOOKS GOOD` while child sheets were never opened. The aggregator also had no status, coverage, or diagnostics contract, and malformed audit results were indistinguishable from an empty findings list.

## Changes

- Reuse `build_hierarchy_node` and run every schematic audit against every reachable schematic file.
- Return additive `status`, per-audit summaries, coverage counts, and structured diagnostics.
- Use the exact `INCOMPLETE — review could not evaluate the full design` verdict whenever coverage is partial or failed, while preserving findings that were gathered.
- Treat zero symbols, zero named nets, a supplied board with zero footprints, unresolved symbols, parse/hierarchy failures, and audit failures as incomplete coverage.
- Report multi-unit symbols as incomplete until #182 makes the affected audit paths unit-aware.
- Reject errored or malformed audit responses instead of treating them as clean results.
- Update the bundled review skill and tool directory so callers never soften an incomplete result into approval.
- Add a KiCad E2E cross-check between Konnect's runtime symbol count and KiCad's exported netlist.

Individual audit response schemas are unchanged. The new design-review fields are additive; the intentional compatibility change is that unverified designs can no longer receive a passing verdict.

## Validation

- `cargo test -p konnect-core --locked review_completion_tests -- --nocapture` (7 passed)
- `cargo test -p konnect --test e2e_kicad --no-run --locked`

The real-KiCad E2E remains ignored locally by design; it compiled successfully and will execute in the repository's KiCad-enabled workflow.

## Risk and rollback

The main risk is stricter verdicts for sparse, unresolved, or multi-unit designs. That is fail-closed behavior requested by #184. Rollback is isolated to the aggregator, its hierarchy helper visibility, documentation, and tests.

### DIFF
--- a/crates/konnect-core/src/tools/design_review.rs
+++ b/crates/konnect-core/src/tools/design_review.rs
@@ -7,16 +7,18 @@
 
 use crate::mcp::protocol::CallToolResult;
 use crate::tool;
-use crate::tools::{get_path, ToolContext, ToolDef};
+use crate::tools::{get_path, project_name_for, sch_hierarchy, ToolContext, ToolDef};
 use konnect_schematic_editor as cse;
 use konnect_sexp::{
     parser::parse_sexp,
     schematic::{
-        extract_lib_pins, extract_symbol_instances, find_lib_symbol, pin_endpoint, read_schematic,
+        extract_labels, extract_lib_pins, extract_lib_pins_for_unit, extract_symbol_instances,
+        find_lib_symbol, pin_endpoint, read_schematic,
     },
 };
-use serde_json::json;
+use serde_json::{json, Value};
 use std::collections::HashSet;
+use std::path::{Path, PathBuf};
 use tracing::info;
 
 // ─── Tool definitions ─────────────────────────────────────────────────────────
@@ -89,6 +91,8 @@ pub fn tools() -> Vec<ToolDef> {
         tool!(
             "run_design_review",
             "Run all available audit checks and produce a consolidated design review report. \
+             Audits every reachable schematic sheet and reports status, coverage, and diagnostics. \
+             Returns an INCOMPLETE verdict instead of approval when coverage is partial or failed. \
              This is the tool to call when the user asks 'is my board ready?' or 'review my design'.",
             json!({
                 "type": "object",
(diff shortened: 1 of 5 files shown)


mixelpixx:
Merged main into your branch (clean) and verified against the contract we agreed on #184 rather than from the description. It matches on every point, and the one thing I most wanted proved, I proved.

**The headline acceptance test genuinely catches the bug.** I added a temporary switch that restores the old root-only behaviour and re-ran:

```
hierarchical_child_is_included_in_coverage_and_audits ... ok       (with the fix)
hierarchical_child_is_included_in_coverage_and_audits ... FAILED   (root sheet only)
```

That was the criterion I said would matter — a populated hierarchical design must not be able to receive `LOOKS GOOD` when a child sheet was never opened. It now can't.

**Against the four decisions from the triage:**

- Third verdict, not a tool error — and the string is exactly `INCOMPLETE — review could not evaluate the full design`, with findings preserved.
- Aggregator adapted, **individual audits untouched** — I checked: no `handle_audit_*` or `handle_check_*` signature or response shape moved. That keeps the public API stable and the diff reviewable, which is what I asked for.
- `build_hierarchy_node` reused via a one-line visibility change rather than a second traversal.
- Suspicious zeros treated as incomplete: blank schematic, zero footprints on a supplied board, unresolved symbols — all covered by name.

**The skill update is the part I'd have forgotten to ask twice for**, and you got it right. "Report that verdict verbatim ... do not describe the design as ready, passing, clean, or looking good" closes the failure mode where the model reads `INCOMPLETE` and paraphrases it into reassurance. A verdict a model can soften is not a verdict.

`clean_single_sheet_can_still_look_good` is the test I look for in changes like this — it guards the over-correction, where fail-closed quietly becomes fail-always. And `multi_unit_symbol_is_partial_until_issue_182_lands` is the right way to handle the #182 overlap: it reports the gap now rather than pretending it isn't there, and it will need revisiting when #182 lands, which the name makes obvious.

Gate green on the merge: fmt, clippy `--all-targets`, doctests, **644 tests**, all seven of your completion tests passing.

Closes #184. Turning a well-specified issue into a merged implementation the same day, with the acceptance criteria met point for point, is genuinely good work — thank you.

Merging once CI is green.

PR #186, opened 2026-08-14
fix(sch): place instance fields at their library anchors

### LINKED ISSUE #101: Schematic property placement: use library anchors instead of hardcoded ±3.81, account for rotation

pauliuszaleckas:
Fixes #101.

`positioned_property` put Reference at `y − 3.81` and Value at `y + 3.81` for every
symbol, ignoring the library's own anchors and the instance rotation. That matched
eeschema only for down-pointing symbols: `power:VCC` and every other up-pointing rail
anchors Value *above* the graphic, and `Device:R` anchors both fields beside the body
at 90°.

The anchors are now read from the definition embedded in the schematic's `lib_symbols`
— the copy KiCad actually draws — and put through the placement transform. The old
offsets remain as the fallback for a symbol that defines no anchor.

### Position and angle are stored differently

Easy to get backwards, so it is worth stating:

- **Position is absolute.** The anchor takes the same flip-rotate-translate as a pin.
- **The angle is relative.** KiCad adds the symbol's rotation when it draws, so the
  library value is written through unchanged.

Verified by rendering a 90° `Device:R` with `kicad-cli sch export svg` — storing the
rotated angle draws the reference vertically across the horizontal body.

### The anchor alone is not enough

`(justify …)` comes with it. Fields that share a row rely on it to stay apart:
`Regulator_Linear:AP2112K-3.3` anchors Reference at `(-5.08, 5.715)` and Value at
`(0, 5.715)`, both `justify left`, so centring them prints `U2` on top of
`AP2112K-3.3`. Like the angle, justification is expressed in the text's own frame, so
the library value is written through unchanged, and a centred field keeps writing no
`(justify …)` at all.

### Scope

Applies to new placements — `add_schematic_component` and `add_power_symbol`, which
share `positioned_property`. A sheet placed before this change keeps its old field
positions until each symbol is re-placed; repairing existing sheets is a separate
concern.

### Testing

`cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings`, the workspace
test suite and the doctests all pass. Nine new unit tests cover anchor extraction,
the text rotation read off each anchor, justify parsing and re-emission in KiCad's
token order, the centred (no-justify) case, opposite-pointing power symbols, and the
three cases with no anchor to read: an unembedded symbol, a unit sub-symbol, and a
property with no `(at …)`.

### DIFF
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -520,6 +520,11 @@ pub(crate) fn add_pin_midwire_junctions(
 /// Hidden properties get KiCAD 10's property-level `(hide yes)` — a sibling
 /// before `(effects)`, exactly as eeschema writes instances (PR #96); the
 /// legacy hide-inside-effects form renders the same but round-trips dirty.
+///
+/// `justify` comes from the library field and is written through unchanged,
+/// like the angle in [`field_at`]: it is expressed in the text's own frame, so
+/// it stays true however the instance is rotated. Centred fields write no
+/// `(justify …)`, which is how KiCad spells centred.
 pub(crate) fn positioned_property(
     name: &str,
     value: &str,
@@ -527,6 +532,7 @@ pub(crate) fn positioned_property(
     y: f64,
     rotation: f64,
     hide: bool,
+    justify: konnect_schematic_editor::library::FieldJustify,
 ) -> konnect_schematic_editor::Property {
     use konnect_schematic_editor::sexp::{atom, SexpNode};
     use konnect_schematic_editor::types::fmt_f64;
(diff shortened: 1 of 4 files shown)


mixelpixx:
Merged main into your branch (clean) and verified this independently rather than from the description. It's a good piece of work and I want to record why.

**The part that matters is the part that's easy to get backwards**, and you got it right: position is absolute and must be transformed, angle is relative and must not be. I confirmed both against the real `Device:R`, whose library anchors are Reference `(2.032, 0) rot 90` and Value `(0, 0) rot 90`:

```
R1 placed at rotation 0   -> Reference (51.562, 49.53) rot 90
                             49.53 + 2.032 = 51.562, X offset applied
R2 placed at rotation 90  -> Reference (100.33, 47.498) rot 90
                             the same offset became a Y displacement
                             49.53 - 2.032 = 47.498
```

The angle stays 90 in both. Rotating it there too would have written 180 and drawn the reference upside down relative to the body — the double-count your doc comment warns about. Verified end-to-end through the real server, then exported the result with `kicad-cli sch export svg` (plots clean) and ran ERC (four violations, all "Pin not connected" on two standalone resistors — nothing from field placement).

Against the old hardcoded `y − 3.81`, R1's reference would have landed at `(49.53, 45.72)`: below the body, where `Device:R` anchors it beside. That's #101 exactly.

**Reading anchors from the schematic's embedded `lib_symbols` rather than re-resolving from disk is the right call**, and it's the same principle #144 established for `find_lib_symbol` — the embedded copy is what KiCad actually draws, so a locally edited symbol places its fields where the user sees them, not where the base definition would.

Test coverage is thorough. `unit_sub_symbol_properties_are_not_anchors` and `a_property_without_an_at_is_skipped` are the two I'd have asked for and they're already there; `opposite_pointing_power_symbols_anchor_value_opposite_ways` is the concrete case from the issue — a `power:VCC` anchoring Value *above* the graphic where GND anchors below, which the old fixed offset could never express.

Keeping the old offsets as the fallback for a symbol that defines no anchor is right too — it means this can't regress a symbol that was previously fine.

Gate green on the merge: fmt, clippy `--all-targets`, doctests, **637 tests**.

Closes #101 — the one I explicitly kept open when #148 landed, because that was net-label orientation and this is instance-field anchors. Thanks for picking it up.

Merging once CI is green.

PR #188, opened 2026-08-14
fix(library): project resolution escaping into an unrelated project, and downgraded argument errors

Pytonballoon810:
> **Stacked on #160.** This branch carries that PR's two commits as well, because a fork branch cannot be a base for an upstream PR. Merge #160 first and this diff reduces to its own two commits, `fix(library): …` and `refactor(library): …`. Neither of them touches the graphics tools; they are here rather than in #160 so that PR's diff stays about one thing.

Two things found while verifying #160 against real KiCAD, both in `library.rs` and both independent of it.

## 1. An unrelated project above you captures library resolution

`project_root_for` walks ancestors for a `.kicad_pro` and takes the first hit. The walk is unbounded, so **any** unrelated project file in **any** ancestor captures the search and every library lookup then resolves against the wrong `KIPRJMOD`. A project nested inside another project's folder is the realistic case, and it fails silently — you get someone else's `Device` library, not an error.

What made me look was duller. Two tests fail on `main` **right now** on a machine with KiCad installed, and pass on CI:

```
tools::library::symbol_source_tests::a_schematic_with_no_project_falls_back_to_its_own_directory
tools::sch_components::tests::add_component_writes_eeschema_style_instance_path
```

Both build a hermetic fixture in a tempdir. `add_component_writes_eeschema_style_instance_path` says so in its own comment — the project `sym-lib-table` "is what makes this hermetic ... the global table's own `Device` entry resolves to whatever KiCad the developer has installed and would shadow the stub". That is exactly what happens: `tempfile::tempdir()` creates a directory whose parent is the system temp directory, something had left twenty `.kicad_pro` files there, and the walk climbed straight out of the fixture and into them. The stub is never consulted. On CI temp is clean, the walk falls through to the file's own directory, and both pass — which is why v0.3.0 shipped green.

The fix ends the search when the file's own directory carries a `sym-lib-table` or an `fp-lib-table`. A directory that states where its libraries come from is a more specific answer than a `.kicad_pro` somewhere above it. A sheet under `<proj>/sheets/` normally carries no table of its own, so hierarchical resolution is untouched and `a_sheet_in_a_subdirectory_resolves_against_the_project_table` still passes.

New test `a_table_beside_the_file_beats_an_unrelated_project_further_up` reproduces the hazard deterministically — nested project under an outer one, no reliance on ambient litter. I checked it fails without the fix rather than assuming it would.

## 2. Missing-argument errors downgraded to handler errors

`require_str` returns a structured `InvalidArgument` `CallToolResult` naming the offending field. Seven handlers in `library.rs` did `map_err(|e| anyhow!("{:?}", e))?` on it, flattening the taxonomy into a debug-formatted string and surfacing as a generic `handler_error`. `pcb_components.rs` already returns the result directly; this makes `library.rs` agree.

Visible at the client — a missing argument now reports:

```json
{"error": {"kind": "invalid_argument", "field": "uuid", "reason": "missing or not a string"}}
```

where it used to be `handler_error`, which is what lets a caller tell "you forgot an argument" from "the tool tried and failed". An unknown-but-present value correctly stays a `handler_error`.

Copilot flagged one instance of this on #159/#160; the pattern was in seven more. No behaviour change beyond the error shape — every affected handler already returned `anyhow::Result<CallToolResult>`.

## Tests

```
cargo test --workspace --lib --tests --no-fail-fast   # all green
cargo fmt --all -- --check                            # clean
cargo clippy -p konnect-core -p konnect-ipc --lib --tests   # no warnings
```

This is the first fully green `--workspace` run I have had on this machine; the two failures above were the only ones outstanding.

## Risk

The resolution change is the one worth a second opinion, since it changes which directory wins for a file that has a table beside it *and* a project above it. My reading is that the table is the more specific statement, and the existing sub-sheet test pins the case where it should still walk. If you would rather it stayed strictly "nearest `.kicad_pro` wins" and the two tests were made hermetic another way — a fixture that plants its own `.kicad_pro` — say so and I will redo it that way; that also fixes the tests but leaves the nested-project hazard in place.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -206,7 +206,7 @@ if !path.exists() {
 
 Adding a new kind: edit `mcp/error.rs`, add the variant, add the match arm in `short_code()`, use it from the handler. The `short_code_matches_serialized_kind_field` test will fail loudly if they drift.
 
-The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 187 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
+The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 189 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
 
 ## Observability
 
@@ -227,7 +227,7 @@ Source: [`crates/konnect-core/src/observability.rs`](crates/konnect-core/src/obs
 
 ## Tool Routing (Starter Kit + On-Demand Loading)
 
-The server does NOT expose all 187 tools (193 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
+The server does NOT expose all 189 tools (195 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
 
 - **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 6 meta-tools, baseline `tools/list` is ~19 tools ≈ 2K tokens.
 - **On demand**: the LLM reads `list_toolboxes` → calls `load_toolset(name)` to expose a toolset's tools in subsequent `tools/list` responses. `unload_toolset(name)` prunes them when the task shifts.
@@ -299,9 +299,9 @@ convention for other `kicad-cli`-calling code.
 
 ## Current Stats
 
-- **18 toolsets, 187 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
+- **18 toolsets, 189 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
 - Baseline `tools/list`: ~19 tools / ~2K tokens (starter kit + meta-tools)
-- Full-catalog `tools/list` (all loaded): 193 tools (187 registered + 6 meta) / ~25K tokens
+- Full-catalog `tools/list` (all loaded): 195 tools (189 registered + 6 meta) / ~25K tokens
 - **0 IPC stubs** (all protobuf methods implemented)
 - **0 unimplemented tools**
 - **3 CLI commands removed in KiCAD v10** (specctra DSN/SES, pcb sync — return clear errors)
--- a/DEV.md
+++ b/DEV.md
@@ -206,7 +206,7 @@ if !path.exists() {
 
 Adding a new kind: edit `mcp/error.rs`, add the variant, add the match arm in `short_code()`, use it from the handler. The `short_code_matches_serialized_kind_field` test will fail loudly if they drift.
 
-The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 189 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
+The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 191 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
 
 ## Observability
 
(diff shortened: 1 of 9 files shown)


mixelpixx:
Excellent find, and the diagnosis is exactly right. I've taken the fix commit (`31757da`) directly onto `main` as `ec705c3`, with your authorship preserved — I didn't want it waiting behind #160.

**I reproduced it rather than taking it on trust.** Planted a single `.kicad_pro` in the system temp directory and `a_schematic_with_no_project_falls_back_to_its_own_directory` went from pass to FAILED, deterministically; removed it, back to green. With your fix applied and the stray file still present, the full workspace suite is green at 625 tests. I also neutered the fix to confirm your new `a_table_beside_the_file_beats_an_unrelated_project_further_up` genuinely fails without it — it does, as does the older test.

One correction to your report: on my machine only **one** of the two tests failed with a single planted file. `add_component_writes_eeschema_style_instance_path` stayed green. Presumably yours needed the particular arrangement of the twenty files you'd accumulated. Doesn't change anything — the underlying bug is real and your fix is right.

**This is a regression I introduced today.** `project_root_for` came in with #136, which I merged and shipped in v0.3.0 a couple of hours ago. It escaped precisely as you say: CI's temp directory is clean, so the suite was green when I tagged. That's a genuinely nasty failure mode — a test that asserts its own hermeticity in a comment while silently resolving against the developer's installed KiCad — and it's the second time this month the `sch_components` fixtures have been caught doing that (#136's `STUB_MARKER` was the first). Worth it.

**One residual I want on record, not blocking.** The walk is still unbounded when the file's directory carries *neither* a table nor a `.kicad_pro` — a schematic at `~/Documents/work/foo.kicad_sch` with an unrelated `~/Documents/Other.kicad_pro` still climbs into it. Your fix covers the case that actually bites and the one that's reproducible; properly bounding the walk is a separate design question (stop at a VCS root? a depth limit? never cross above home?) and I'd rather not decide it inside a hotfix. Filing a follow-up.

**What's left on this PR.** The error-taxonomy commit (`6b008e3`) is a real improvement and I want it — `map_err(|e| anyhow!("{:?}", e))` flattening a structured `InvalidArgument` into a debug string is exactly the kind of thing that makes a tool hard to drive. But it changes error *kinds* on seven handlers, and I've just shipped a release; I'd rather it land with #160 on a normal review cycle than ride along in a patch. So this PR stays open for that plus the #160 stack.

On **#160** itself: still needs a live IPC verification pass against real KiCAD before I merge it — 353 lines of outbound message construction is exactly where a schema-valid-but-malformed message hides, and no offline test can catch it. Happy to run that myself. It also still needs `README.md` in the tool-count update; there's a test on `main` now (`crates/konnect/tests/doc_tool_counts.rs`) that will fail on the current diff and tell you so.

Thank you — finding this within two hours of the tag is the reason it's a footnote rather than a bug report next month.

mixelpixx:
The fix is shipped in **v0.3.1** — https://github.com/mixelpixx/Konnect/releases/tag/v0.3.1

Thanks again for catching it inside the release window. Notes credit you and explain the CI-hermeticity failure mode, since it is worth other people knowing about.

This PR stays open for the error-taxonomy commit and the #160 stack.

mixelpixx:
Taken the error-taxonomy commit too — cherry-picked as `dd57a81`, your authorship preserved. Both of this PR's own commits are now on `main`.

Verified the shape through the real server rather than from the diff:

```
get_symbol_info {}   ->  {"error":{"field":"lib_id","kind":"invalid_argument",
                                   "reason":"missing or not a string"}}
```

where it previously produced a `handler_error` carrying a debug-rendered string.

**I added the test it was missing** (`7b983bc`). Your description said "no behaviour change beyond the error shape" — but the error shape is exactly what a caller sees, so it is worth pinning. Confirmed it catches the old pattern by restoring one call site and watching the handler bubble an `Err` instead of a typed result.

Writing that test turned up something worth its own issue, now [#194](https://github.com/mixelpixx/Konnect/issues/194): **`get_path` never got the same treatment**, so a missing *path* argument is still a `handler_error`. `list_symbols_in_library {}` reports `handler_error`; `get_symbol_info {}` reports `invalid_argument`. Whether a caller gets a typed error depends on which helper the handler happened to use for that argument — and path arguments are the most commonly omitted ones, so it is the case where the taxonomy would help most.

The tell was in the test itself: it has to supply each tool's path arguments explicitly, or `get_path` fails the call before the `require_str` under test is reached and the assertion goes vacuous. A test having to work around an inconsistency is usually the inconsistency asking to be fixed.

**What's left on this PR is #160's two commits.** Two things still needed there, unchanged from before:
1. A live IPC verification pass against real KiCAD. 353 lines of outbound message construction is exactly where a schema-valid-but-malformed message hides, and no offline test catches it. Happy to run it — I did the same for #137 earlier today and it took a few minutes.
2. `README.md` in the tool-count update. `crates/konnect/tests/doc_tool_counts.rs` on `main` will fail on the current diff and tell you precisely.

Note `main` has moved a fair way today — v0.3.0, v0.3.1, and since then #186 (field anchors) and #185 (design-review coverage) — so expect a real rebase.

Thanks again. Two solid fixes out of one afternoon's incidental testing, and the project-root one caught a regression inside its release window.

mixelpixx:
Closing — the two fixes this PR was opened for are on `main`, and what is left
of the branch belongs to #160.

Both landed yesterday, byte-identical to the commits here (compared by
`git patch-id`, not by subject line):

- `31757da` -> `ec705c3` — a table beside the file wins over an unrelated project above it
- `6b008e3` -> `dd57a81` — stop downgrading missing-argument errors to handler errors

The first one is the fix I most wanted; a nested project silently capturing
every library lookup is the kind of thing that only shows up as "the wrong
symbol appeared" much later. Thank you for it.

The branch still carries `fd76145` and `ed104a2`, the footprint-graphics pair.
That is the same work as #160, so keeping it alive in two PRs means whichever
lands second conflicts with itself. #160 is where I have been leaving review
notes, so I would rather consolidate there — nothing is being dropped by
closing this.

One thing that has changed since we last talked about it: #205 landed, adding
footprint-graphics editing for **library files** with layer and mode selectors.
It does not overlap the on-board IPC half in #160 — editing a placed footprint
through a running KiCad is still wanted and still unimplemented — but #160 will
want a rebase onto it so the two share the count updates rather than fight over
them. Details are in the #160 thread.

Also worth knowing for the rebase: the tool counts moved a long way yesterday
(188 -> 200 registered, 18 -> 19 toolsets), and `doc_tool_counts.rs` enforces
them across README.md, DEV.md and tool-directory.md. Sum `tool_count` from the
registry rather than adjusting by hand.

PR #193, opened 2026-08-14
fix(pcb_board): honour the board argument and fail closed on IPC rejection

### LINKED ISSUE #191: add_board_outline ignores the board argument and writes into the board open in KiCad
`add_board_outline` writes into **the board currently open in KiCad**, whatever path is passed in `board`. The target file is left untouched.

**Environment:** Konnect v0.3.0 (build 2026-08-14), KiCad 10.0.4, Windows 11, PCB editor open.

## Repro

Board A open in the PCB editor, board B a different project on disk (not open):

```
add_board_outline(board="<path to board B>", x1=50, y1=50, x2=70, y2=70, corner_radius=1)
  → {"height":20.0,"source":"ipc","width":20.0,"x1":50.0,…}
```

- the returned `source` is `ipc`, although a file path for a different board was given
- board B on disk: unchanged, still no outline segments
- board A: gains the rectangle in memory, and with it into the next save

## Where it comes from

This looks like the unconsidered edge of #5, which moved `set_board_size`, `add_board_outline` and `add_board_text` to "IPC first, file fallback" (`with_ipc(...).is_ok()` → IPC, otherwise file) and introduced the `source` field. That decision never takes the `board` argument into account — the PR only considered the active-session case, not path-based board selection.

**The same pattern therefore likely affects `set_board_size` and `add_board_text` as well**; only `add_board_outline` was exercised here.

Note the schema changed in v0.3.0 — `board` is now optional, which makes "operate on the open board" a legitimate default. The issue is that an **explicitly given** path is silently disregarded rather than honoured or rejected.

## Cost in practice

During a test session this rectangle twice had to be undone by hand in the GUI, and once slipped into an unrelated `save_project`, landing in a finished layout that then had to be restored from a backup.

## Suggested fix

If `board` is present and does not resolve to the board currently open over IPC, either operate on that file (file mode) or return a clear error (`board parameter does not match the open board`). Worth auditing every IPC-capable tool taking a `board` argument for the same pattern — starting with the other two tools from #5.

anyn99:
## Summary

The board-graphics tools sent their work to IPC whenever KiCad answered at all, which meant they ignored the caller's `board` argument: a request naming board B, while KiCad had board A open, edited **A**. The response even said `"source": "ipc"` while B on disk stayed empty.

Closes #191

While fixing that, a second failure mode in the same line of code had to be separated out: falling back to a file edit on *any* IPC error. When KiCad holds the board open, that edit is invisible to it and is discarded by its next save — the tool reports success and nothing happens.

## Approach

Both halves are decided the way #137 already established for `add_mounting_hole`, now factored into one helper (`attempt_ipc_write`) and used by **all five** board-mutating tools:

- **Board identity** — `ensure_board_is_active` gates the IPC path, so a request naming a board KiCad does not have open can never mutate the board it does.
- **Fallback gate** — the typed transport classification (`IpcFailure::Unreachable` vs `Rejected`), never a text match on the message. Only a request that never reached a live KiCad may edit the file; a KiCad that answered, even with an error, fails closed.

`add_mounting_hole` moves onto the shared helper too, so the rule lives in exactly one place instead of being reimplemented per tool. It gains the identity check it previously delegated to `place_footprint`. `BoardWrite<T>` is generic so the mounting hole can still return its placed footprint.

**Where the bug came from:** #5 moved `set_board_size`, `add_board_outline` and `add_board_text` to "IPC first, file fallback" (`with_ipc(..).is_ok()`) and introduced the `source` field. That change only ever considered the active-session case; path-based board selection was never part of it, so the `board` argument quietly stopped mattering for these tools. `import_svg_logo` inherited the same pattern later.

`get_board_extents` keeps its file fallback — it only reads, so it cannot overwrite anything — but it now reads the board that was asked for.

## Compatibility and safety

No public API, schema, or config change: same tool names, same arguments, same response fields.

Behaviour changes in two places, both narrowing what can happen:

1. A request naming a board other than the one KiCad has open no longer mutates the open board. With KiCad reachable it now returns an error naming the requested board; with KiCad unreachable it edits the requested file as before.
2. A reachable KiCad that refuses the operation no longer causes a silent file edit; the tool returns the rejection and leaves the file byte-identical.

Callers that relied on "any IPC error means edit the file" will see errors where they previously saw a success that did nothing observable. This matches the repo's own review rule: *"IPC mutations verify the requested board and do not leave partial batches."*

## Validation

```
cargo fmt --all -- --check                        clean
cargo test --workspace --locked --lib --tests     all green, no failures
cargo test --workspace --locked --doc             all green
cargo clippy --workspace --locked -- -D warnings  clean
```

**New regression coverage** — four tests, using the mock IPC endpoint introduced with #137:

- a reachable KiCad that rejects must leave the board file byte-identical, for `add_board_outline`, `add_board_text` and `set_board_size`
- an unreachable KiCad must still produce the file edit (four `Edge.Cuts` segments)

The existing `add_mounting_hole` tests pass unchanged, which is what establishes that moving it onto the shared helper kept its behaviour. Test helpers are shared with the mounting-hole module rather than duplicated.

**Verified against real KiCad** (10.0.4, Windows, PCB editor open on project A):

| check | result |
|---|---|
| `add_board_outline(board=<B>)`, B not open | rejected, error names the requested board |
| board A and board B on disk afterwards | both md5-identical to before |
| `get_board_extents(board=<B>)` | falls back to file, reads B |
| IPC path against the open board still works | confirmed via `find_component` / `get_component_pads` |

## Review checklist

- [x] No tools added or removed, so no count or doc updates are due.

## Follow-up worth considering (not in this PR)

`IpcFailure` has two cases, but board work has three. "KiCad is running, but the requested board is not open" currently arrives as `Rejected`, because `find_open_board` fails with a plain error — so the tools refuse, even though a file edit would be both safe and correct there (no editor holds that file). In practice that means: with KiCad open on project A, board B cannot be edited through these tools at all.

A marker type alongside `TransportUnreachable`, set by `find_open_board`, plus a third `IpcFailure` variant would let every board tool decide three-way — unreachable → file, board-not-open → file, rejected → fail closed — without matching on message text, which the codebase rules out. Happy to open that as a separate issue or PR if it sounds right.

### DIFF
--- a/crates/konnect-core/src/tools/pcb_board.rs
+++ b/crates/konnect-core/src/tools/pcb_board.rs
@@ -52,6 +52,64 @@ where
     }
 }
 
+/// What a board-mutating tool should do after its IPC attempt.
+pub(crate) enum BoardWrite<T = ()> {
+    /// KiCAD applied the change; report `"source": "ipc"`. Carries whatever the
+    /// IPC call returned, for tools that echo it back — the placed footprint,
+    /// for instance.
+    Ipc(T),
+    /// No live KiCAD on this transport, so a direct file edit cannot race an
+    /// editor; proceed with the S-expression path.
+    File,
+    /// KiCAD answered and refused. The caller must return this result and must
+    /// NOT touch the file.
+    Refused(CallToolResult),
+}
+
+/// Run `f` over IPC against the board named by `board_path`, deciding what the
+/// caller may do next.
+///
+/// Two failure modes that used to look alike are kept apart here, both of which
+/// silently corrupted work before:
+///
+/// * The board reached over IPC is whichever one KiCAD has open, so a request
+///   naming a *different* board would edit the wrong one — `ensure_board_is_active`
+///   rejects that up front (issue: `add_board_outline` writing into the open board).
+/// * A file-only edit is invisible to a KiCAD holding this board open and is
+///   discarded by its next save. So the fallback gate is the typed transport
+///   classification, never a text match: only a request that never reached a live
+///   KiCAD may edit the file; a KiCAD that answered — even with an error — fails
+///   closed. `handle_add_mounting_hole` established that rule; this is it made
+///   reusable, so every board-mutating tool decides the same way.
+pub(crate) async fn attempt_ipc_write<T, F>(
+    addr: String,
+    board_path: &std::path::Path,
+    what: &str,
+    f: F,
+) -> anyhow::Result<BoardWrite<T>>
+where
+    T: Send + 'static,
+    F: FnOnce(&konnect_ipc::client::KiCadIpcClient) -> anyhow::Result<T> + Send + 'static,
+{
+    let requested = board_path.to_path_buf();
+    match crate::tools::pcb_components::with_ipc_classified(addr, move |c| {
+        c.ensure_board_is_active(&requested)?;
+        f(c)
+    })
+    .await?
+    {
+        Ok(value) => Ok(BoardWrite::Ipc(value)),
+        Err(konnect_ipc::IpcFailure::Rejected(message)) => {
+            Ok(BoardWrite::Refused(CallToolResult::error(format!(
+                "KiCAD rejected the {what} over IPC: {message}. \
+                 The board file was not modified — KiCAD is reachable and may hold this \
+                 board open, so editing the file directly could be silently overwritten."
+            ))))
+        }
+        Err(konnect_ipc::IpcFailure::Unreachable(_)) => Ok(BoardWrite::File),
+    }
+}
+
 // ─── S-expression format helpers ──────────────────────────────────────────────
 
 fn format_gr_line(x1: f64, y1: f64, x2: f64, y2: f64, layer: &str, width: f64) -> String {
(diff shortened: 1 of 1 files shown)


mixelpixx:
Merged main into your branch (clean) and reproduced #191 end-to-end against real KiCAD before and after, because "edits the wrong file" is the one class where I want to see it with my own eyes.

**The bug is confirmed on `main`.** `pcb_board.rs:432` and four siblings do `with_ipc(...).is_ok()` with no identity check at all — nothing in those handlers reads the `board` argument before dispatching. So the request goes to whatever KiCad has open, and `"source": "ipc"` is returned regardless.

**The fix, verified live.** Board A (a demo board) open in pcbnew, board B a separate file on disk, never opened — your exact repro:

```
add_board_outline(board=B, x1=50, y1=50, x2=70, y2=70)
  -> isError: true
     "requested board '…/B_closed.kicad_pcb' is not open in KiCAD
      (open boards: …/A_open.kicad_pcb)"
```

Then the two things that actually matter:

- **B is byte-identical** afterwards — `0ca5d56c…` before and after. A reachable-but-refusing KiCad produced no consolation file edit.
- **A is untouched too**, and structurally guaranteed: `attempt_ipc_write` runs `c.ensure_board_is_active(&requested)?` *before* `f(c)`, so the `?` short-circuits and no `create_items` is ever sent. That is the right ordering — checking after dispatch would be no check at all.
- **The correct board still works**: `add_board_text` against A returns `{"source":"ipc", …}` and succeeds.

**On the design.** Factoring this into one `attempt_ipc_write` rather than repeating the rule five times is the right call, and moving `add_mounting_hole` onto it is what stops the two copies drifting — that tool got the discipline in #137 and would otherwise have been the odd one out. Gating on the typed `IpcFailure::Rejected` vs `Unreachable` rather than a text match matters more than it looks: message-matching is how this class of gate rots silently when upstream rewords an error.

Keeping the file fallback for `get_board_extents` while still making it read the board that was asked for is exactly the right distinction — it cannot overwrite anything, so the fallback is safe, but it was still answering about the wrong board.

Your four regression tests cover both directions (rejecting → file untouched, unreachable → file edited), which is the pair that matters. Gate green: fmt, clippy `--all-targets`, doctests, **650 tests**.

**Worth stating for the record**: this predates the recent releases — it came in with #5, when those three tools moved to "IPC first, file fallback" and path-based board selection was never considered. It is not a v0.3.x regression, it is a long-standing bug that your repro finally pinned down.

Closes #191. Merging once CI is green.

Issue #103, opened 2026-07-29
Plugin leaks orphan konnect server processes: PID file only tracks the last one


Nigh:
## Summary

The KiCad plugin (`plugin/__init__.py`) can accumulate multiple orphan `konnect` server processes that are never reaped. On my machine I found **5** live `bin/konnect` processes with KiCad closed.

## How it happens

1. `_run_server` launches the server with `subprocess.Popen` from a daemon thread. When KiCad exits, the thread dies but nothing terminates the child, so the server is orphaned. This much is by design — `start_server` has a preflight `_kill_tracked()` with the comment "Orphan from a previous KiCAD session may still be holding the port."
2. The problem is that `~/.cache/konnect/cache/server.pid` records only the **most recent** PID, and `_kill_tracked` kills only that one. Any time more than one orphan exists, the older ones are never tracked and never killed:
   - multiple KiCad sessions where the PID file was overwritten or removed by `_run_server`'s `finally` block,
   - MCP clients (Claude Desktop, Cursor, …) spawning their own instances of the same binary while the plugin has its own,
   - a crash path that deletes the PID file while the process survives.

Each orphan holds ~17 MB and keeps running a stale binary after the plugin is updated.

## Suggested fixes (either works)

- Terminate the child on KiCad exit: keep the `Popen` handle and register an `atexit` handler (and/or hook the wx app shutdown) that calls `terminate()`/`kill()` on it.
- Or make the PID file a list: append every spawned PID, and have `_kill_tracked` kill and prune all of them.

## Environment

- Linux, KiCad 10.0, Konnect v0.2.1 installed via PCM

mixelpixx:
Confirmed, and the diagnosis is right about the PID file — but the problem is bigger than the ActionPlugin path, so I want to make sure the fix actually covers your five orphans.

**Three ways a server gets spawned; `server.pid` tracks one:**

1. **ActionPlugin** (`plugin/__init__.py`) — tracked, and it has the race you describe plus one more: the `finally` block removes the PID file without checking whose PID it holds. So if instance A exits after instance B started, A deletes B's record and B is orphaned permanently.

2. **KiCad 10's IPC entrypoint** — `plugin.json` declares `"entrypoint": "bin/konnect"` with an `exec` runtime, so KiCad launches the binary **directly**, never going through `__init__.py`. No PID is recorded at all. This is likely where most of your five came from, and neither of your proposed fixes would catch it.

[...]

PR #199, opened 2026-08-15
fix(plugin): keep the server PID record correct across sessions and reap on exit

tonydzi:
the plugin-side half of #103, scoped so it does not collide with the `<cache>/run/<pid>.lock` work you described for `crates/konnect/src/main.rs`. this touches `plugin/__init__.py` only.

### problem

two defects, both in the record's lifetime, both named in your comment on #103:

1. **`_run_server`'s `finally` removed the PID file without checking whose PID it held.** two KiCAD sessions share one `server.pid`. session A starts and records A, session B starts and overwrites with B, then A's server exits and A deletes B's record. B keeps running and is now untracked, so no later `start_server` preflight can ever reap it — one of the shapes behind the reporter's five orphans.
2. **nothing ended the child on KiCAD exit.** `_run_server` sits on a daemon thread; interpreter shutdown drops the thread and leaves the child. with `transport = "http"` the server never reads stdin, so parent death is invisible to it and it holds the port (and a possibly pre-update binary) indefinitely.

### design

- `_write_pid_file(pid)` / `_clear_pid_file(pid)`: clear is compare-and-delete — it removes the file only while the recorded PID is still ours. an unreadable or foreign record is left alone.
- `_terminate_owned_server()` registered with `atexit`: `terminate()`, then `kill()` after 5s, then clear our record. best-effort by construction — a hard kill of KiCAD runs no handler, which is exactly why the preflight reap stays.

no behaviour change for the single-session case, no new dependency, no config or path surface touched. this is belt-and-braces only: it does not see the IPC-entrypoint or external-MCP spawn paths, so the `main.rs` lock file is still the fix that covers all three.

### tests run

no python test harness exists in the repo, so i ran the repro standalone against both trees. it stubs `pcbnew` / `wx` / `settings_dialog` and needs nothing but stock python3:

```
$ python3 repro_103.py <pristine main>
[FAIL] pid record: A's cleanup deleted B's record -- B is orphaned untracked
[FAIL] kicad exit: no shutdown handler: the spawned server outlives KiCAD

$ python3 repro_103.py <this branch>
[PASS] pid record: B's record survived A's cleanup
[PASS] kicad exit: server reaped on exit (rc=-15), record cleared
```

case 1 runs the real `_run_server` against a short-lived stand-in binary and writes B's PID into the file mid-flight, so it exercises the actual race rather than asserting on the helper. mutant check: deleting just the `if recorded != pid: return` guard turns case 1 red again and leaves case 2 green, so the assertion is load-bearing.

no rust file is touched, so the cargo checklist is unaffected; i did not run it (no rust toolchain on the machine i tested from).

<details>
<summary>repro_103.py</summary>

```python
"""Repro for Konnect #103, plugin side.

  python3 repro_103.py <path-to-Konnect-checkout>
"""

import importlib.util
import os
import subprocess
import sys
import tempfile
import threading
import time
import types


def load_plugin(repo):
    for name in ("pcbnew", "wx", "settings_dialog"):
        sys.modules[name] = types.ModuleType(name)
    sys.modules["pcbnew"].ActionPlugin = type("ActionPlugin", (), {"register": lambda self: None})
    sys.modules["wx"].MessageBox = lambda *a, **k: None
    sys.modules["wx"].GetTopLevelWindows = lambda: []
    sys.modules["wx"].OK = sys.modules["wx"].ICON_ERROR = 0
    sys.modules["wx"].ID_YES = sys.modules["wx"].ID_NO = 0
    sys.modules["settings_dialog"].KonnectSettingsDialog = object
    sys.modules["settings_dialog"].load_settings = lambda *a, **k: {}
    spec = importlib.util.spec_from_file_location(
        "konnect_plugin_under_test", os.path.join(repo, "plugin", "__init__.py")
    )
    mod = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(mod)
    return mod


def case_pid_record(plugin, tmp):
    plugin._CACHE_DIR = tmp
    plugin._PID_FILE = os.path.join(tmp, "server.pid")
    sleeper = os.path.join(tmp, "server-a.sh")
    with open(sleeper, "w") as f:
        f.write("#!/bin/sh\nsleep 1.5\n")
    os.chmod(sleeper, 0o755)
    plugin.BINARY_PATH = sleeper
    plugin.SETTINGS_PATH = os.path.join(tmp, "nope.json")

    thread = threading.Thread(target=plugin._run_server)
    thread.start()
    time.sleep(0.4)                                   # A is running and tracked
    b_pid = 424242
    with open(plugin._PID_FILE, "w") as f:            # B starts, takes the record
        f.write(str(b_pid))
    thread.join(10)                                   # A's server exits, A cleans up

    if not os.path.exists(plugin._PID_FILE):
        return False, "A's cleanup deleted B's record -- B is orphaned untracked"
    with open(plugin._PID_FILE) as f:
        recorded = f.read().strip()
    if recorded != str(b_pid):
        return False, f"record is {recorded!r}, expected B's {b_pid}"
    return True, "B's record survived A's cleanup"


def case_kicad_exit(plugin, tmp):
    if not hasattr(plugin, "_terminate_owned_server"):
        return False, "no shutdown handler: the spawned server outlives KiCAD"
    plugin._PID_FILE = os.path.join(tmp, "server2.pid")
    child = subprocess.Popen(["sleep", "30"])
    plugin._server_process = child
    plugin._write_pid_file(child.pid)
    plugin._terminate_owned_server()
    if child.poll() is None:
        child.kill()
        return False, "handler left the server running"
    if os.path.exists(plugin._PID_FILE):
        return False, "handler killed the server but left its record behind"
    return True, f"server reaped on exit (rc={child.returncode}), record cleared"


def main():
    repo = sys.argv[1] if len(sys.argv) > 1 else "."
    plugin = load_plugin(repo)
    failures = 0
    with tempfile.TemporaryDirectory() as tmp:
        for name, fn in (("pid record", case_pid_record), ("kicad exit", case_kicad_exit)):
            try:
                ok, detail = fn(plugin, tmp)
            except Exception as exc:
                ok, detail = False, f"{type(exc).__name__}: {exc}"
            print(f"[{'PASS' if ok else 'FAIL'}] {name}: {detail}")
            failures += 0 if ok else 1
    return 1 if failures else 0


if __name__ == "__main__":
    sys.exit(main())
```

happy to land it as a checked-in test instead if you want a python harness in the repo — say where it should live and i will follow up.

</details>

### risk

low. `plugin/__init__.py` only; the failure mode of the new code is doing nothing (a foreign or unreadable record is left in place), which is the pre-existing preflight's problem to solve. one behaviour worth calling out: on a clean KiCAD exit the server now stops, where before it survived — that is the point of the change, but it does mean a user who relied on the http server outliving KiCAD would notice.


mixelpixx:
Merging this — the compare-and-delete is the right shape, and the docstring explaining *why* an unconditional `os.remove` breaks multi-session use is doing real work for the next reader.

**I landed your repro as a test rather than leaving it in the description.** It's now `plugin/tests/test_pid_lifecycle.py` with a `Plugin (Python)` CI job. The plugin is Python, so cargo never saw it — its one piece of real logic had no regression cover at all, which is why this bug could exist. The tests stub `pcbnew`/`wx`/`settings_dialog` in `sys.modules` (the PID functions touch none of them), so no KiCAD is needed.

Four cases: a session clears its own record; **a foreign record survives another session's exit** (your bug); a missing or unparseable record is not an error; and the cache directory is created on write. I checked the middle one is load-bearing by restoring the unconditional remove — it fails, the other three still pass. That answers the question you asked in the PR about where a test should live.

**This does not close #103**, and your PR body is right that it shouldn't. It covers one of three spawn paths: the ActionPlugin. The KiCad 10 IPC entrypoint (`plugin.json` → `bin/konnect.exe`) never imports `__init__.py` at all and records no PID, and external MCP clients have their own lifecycles that must *not* be reaped. The `<cache>/run/<pid>.lock` sweep in `main.rs` is still the real fix; this is the belt-and-braces half, which is worth having on its own.

Two things I'll carry into the release notes: on a clean KiCAD exit the HTTP server now stops where it used to survive, and because a foreign PID record now persists rather than being deleted, `start_server`'s preflight `_kill_tracked()` will reach another session's *running* server more often than before. That was already true on main; this just makes the path more reachable.

Issue #190, opened 2026-08-14
create_netclass writes an unparseable node into .kicad_pcb — board no longer loads


anyn99:
`create_netclass` reports success but leaves the board file in a state KiCad refuses to open. In the previous release the tool was a silent no-op; as of v0.3.0 it writes, and the write corrupts.

**Environment:** Konnect v0.3.0 (build 2026-08-14), KiCad 10.0.4, Windows 11.

## Repro

On a freshly created, otherwise untouched project:

```
create_project(name="ncprobe", path=<tmp>)
kicad-cli pcb drc ncprobe.kicad_pcb     → loads fine ("1 violation found")

create_netclass(board="<abs>/ncprobe.kicad_pcb", name="Power", clearance=0.15)
  → {"created_netclass":"Power","clearance":0.15,"trace_width":0.25,…}

kicad-cli pcb drc ncprobe.kicad_pcb     → "failed to load the board"
```

## What lands in the file

Appended as a direct child of `kicad_pcb`, right after the net list:

```
	(net 0 "")

      (netclass "Power"
        (clearance 0.15)
        (trace_width 0.25)
        (via_drill 0.4)
        (via_diameter 0.8)
      ))
```

Two problems:

1. **Wrong container** — `netclass` is not a valid node directly under `kicad_pcb`, so the parse fails.
2. **Wrong file** — since KiCad 7, net classes live in the project file: `<project>.kicad_pro` → `net_settings.classes[]` (`name`, `clearance`, `track_width`, `via_diameter`, `via_drill`, …). Writing them into `.kicad_pcb` cannot work even with correct syntax. That also explains the old no-op: the class never appeared where KiCad looks for it.

## Suggested fix

Write into `<project>.kicad_pro` under `net_settings.classes`, deriving the project path from the board path (the `.kicad_pro` sits next to it). If it cannot be located, fail loudly rather than touching the board.

## Acceptance test

Create a project, call `create_netclass`, then (a) `kicad-cli pcb drc` still loads the board and (b) the class appears in `.kicad_pro` → `net_settings.classes`.

triglav-modular:
Still present in **v0.3.2**, and on a different platform to the original report, so it isn't Windows- or fresh-project-specific.

**Environment:** Konnect v0.3.2 (`serverInfo` reports `{"name":"konnect","version":"0.3.2"}`), KiCad 10.0.5, macOS 15 (arm64), server driven over stdio.

[...]

PR #205, opened 2026-08-15
feat: edit and place footprints safely

### LINKED ISSUE #200: feat(library): edit footprint graphics atomically
## Problem

The `library` toolset can create a footprint and edit existing pads, but it
cannot safely add, replace, or delete graphical primitives in an existing
`.kicad_mod`. Agents that need an accurate non-rectangular courtyard or
fabrication outline currently have no supported alternative to direct text
editing.

The immediate acceptance case is adding one or two precise `B.CrtYd`
polygons to existing keyboard socket footprints, while preserving every pad
and unrelated item. The capability itself should remain footprint- and
project-agnostic.

## Proposed interface

Add one atomic MCP tool to the `library` toolset:

```json
{
  "footprint_path": "/abs/path/Part.kicad_mod",
  "selector": {"layer": "B.CrtYd"},
  "mode": "replace",
  "graphics": [
    {
      "type": "poly",
      "points": [
        {"x": -8.695, "y": 3.615},
        {"x": -8.690, "y": 3.566},
        {"x": -8.650, "y": 3.500}
      ],
      "stroke_width_mm": 0.05,
      "fill": "none"
    }
  ]
}
```

- Tool name: `set_footprint_graphics`
- Modes: `append`, `replace`, and `delete`
- Initial primitives: `line`, `arc`, `rect`, `circle`, and `poly`
- Selection: top-level footprint graphic primitives on one canonical KiCad
  layer; text, pads, models, properties, and graphics on other layers are
  preserved
- Polygons: callers provide at least three distinct finite points; Konnect
  closes the polygon, and normalizes a repeated final point
- Version handling: callers do not provide `uuid`/`tstamp` or KiCad syntax;
  Konnect emits syntax compatible with the source footprint

`replace` and `delete` should reject a selected primitive that is referenced
by an existing footprint group. Silently deleting it would leave a dangling
group member; changing group ownership is better handled by a future tagged
selector rather than hidden in the layer-level operation.

## Safety

The implementation can reuse the existing S-expression infrastructure:

1. Require a `.kicad_mod` path and a `footprint` root node.
2. Read the exact source with `read_consistent`.
3. Locate only direct child graphic blocks with
   `find_direct_child_blocks`.
4. Build targeted `SexpEdit` replacements, preserving unrelated source bytes.
5. Parse the complete result before writing.
6. Commit with `write_atomic_if_unchanged`, so a concurrent KiCad or editor
   save returns a conflict instead of being overwritten.

Missing files, invalid layers/geometries, grouped-item conflicts, and
concurrent writes should use the existing structured error taxonomy.

## Verification surface

Extend `get_footprint_info` without changing its existing fields:

- optional `include_graphics` and `graphics_layer` inputs;
- when requested, return each supported primitive's type, layer, geometry,
  point count/closed state where applicable, stroke width, fill, and item ID
  when present.

## Alternatives considered

- A CRUD tool family exposes more surface area, needs multiple round trips,
  and cannot replace a set atomically.
- A named group/tag selector is more precise when manual and generated
  graphics share a layer, but requires group ownership and membership
  lifecycle in the first release. It can be added compatibly after the
  layer-level operation.

## Acceptance criteria

- Two independent `B.CrtYd` polygons can coexist.
- Replace/delete affect only supported primitives on the selected layer.
- Append preserves existing selected graphics.
- Pads and unrelated blocks remain byte-equivalent.
- Invalid/non-finite geometry and non-canonical layers are rejected.
- Grouped selections and concurrent source changes are rejected safely.
- Output round-trips through `konnect_sexp::parse_sexp` and exports through
  KiCad 10.
- Tool schema, registry counts, `tool-directory.md`, developer docs, README,
  and the bundled library skill stay in sync.

TWT233:
## Summary

Add safe, atomic editing for graphics, pads, metadata, and 3D models in existing footprints, plus portable library-table upserts and hardened headless placement so the workflow works end to end without direct KiCad source edits.

Closes #200

## Approach

- Add `set_footprint_graphics` with `append`, `replace`, and `delete` modes for top-level `line`, `arc`, `rect`, `circle`, and `poly` primitives on one canonical layer.
- Extend `get_footprint_info` with opt-in graphic inspection so callers can verify geometry, point counts, stroke, fill, and item IDs.
- Preserve unrelated footprint source with targeted edits; reject grouped selected graphics instead of leaving dangling group members.
- Extend `edit_footprint_pad` with opt-in `match_all` and `new_number` for atomic duplicate-pad renumbering.
- Extend `create_footprint` pads with optional canonical `layers`, `rotation`, and `roundrect_rratio`; legacy payloads retain the existing front-side defaults.
- Update the bundled `kicad-library` and `kicad-pcb` skills so agents use the atomic footprint APIs, safe placement fallback, and front/back pad contracts instead of editing KiCad files directly.
- Add atomic `set_footprint_metadata` and `set_footprint_models` tools.
- Add `replace_existing` to `register_footprint_library` so stale project URIs can be updated in place while preserving options, descriptions, and unrelated entries.
- Harden the existing IPC-unreachable `place_component` file fallback for legacy `fp_text reference` footprints, duplicate-reference rejection, and revision-aware board writes.

The keyboard socket geometry used for acceptance remains outside Konnect. The implementation exposes only project-agnostic primitives, metadata, models, library registration, and placement behavior.

## Compatibility and safety

- Existing tool names and default behavior remain compatible. New arguments are optional and default to the previous single-item/idempotent behavior.
- All footprint mutations require a regular `.kicad_mod` file with a `footprint` root, validate the complete replacement, and commit with `write_atomic_if_unchanged`.
- `replace`/`delete` of grouped graphics returns a structured conflict; no group membership is silently changed.
- Invalid layers, non-finite or degenerate geometry, invalid metadata/model payloads, missing files, stale revisions, malformed tables, and duplicate library nicknames are rejected before writes.
- Project library upserts preserve existing entry metadata and convert same-repository project paths to portable `${KIPRJMOD}` URIs.
- Headless placement is used only when IPC is unreachable. A reachable KiCad rejection still fails closed so a live editor cannot be raced.
- Headless placement preserves full pads, graphics, attributes, and models, rejects duplicate references, validates one complete board form, and uses revision-aware atomic persistence.
- Rollback is a normal branch revert. No migration or persistent config change is required; installed tool binaries are not modified by this PR.

## Validation

- [x] `cargo test --workspace --locked --lib --tests` (all non-environment tests passed; real-GUI/e2e tests remained explicitly ignored)
- [x] `cargo test -p konnect --test protocol_stdio --locked` (9 passed)
- [x] `cargo test -p konnect --test doc_tool_counts --locked` (3 passed)
- [x] `cargo test -p konnect --test asset_references --locked` (3 passed)
- [x] `KICAD_CLI=~/.local/bin/kicad-cli python -m unittest -v test.test_socket_library_update test.test_lh60_sockets` in the external LH60 consumer (18 passed)
- [x] Deployed-binary MCP probe: bottom-side pad layers, 180° rotation, and roundrect ratio serialized exactly as requested
- [x] External LH60 consumer: generated 21 G/K/Dual footprints, exported all 21 to SVG, and passed a clean KiCad 10 coupon DRC with 0 violations / 0 unconnected items
- [x] KiCad 10 DRC on an MCP-generated conflict coupon: one `courtyards_overlap`, 0 unconnected items
- [x] Disposable MCP project probe: create project, set outline, register library, place a legacy footprint through file fallback, then KiCad 10 DRC: 0 violations, 0 unconnected items

## Review checklist

- [x] The diff is focused and contains no generated output, personal data, or keyboard-specific geometry.
- [x] New names follow `docs/NAMING_CONVENTIONS.md`; existing public names are unchanged.
- [x] New behavior and failure paths have regression coverage, including old/current footprint syntax and stale revisions.
- [x] IPC mutations verify the requested board and the file fallback remains gated on typed IPC-unreachable failures.
- [x] Tool counts and docs are synchronized: library 17 tools, 191 registered tools, 197 including meta-tools.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -65,7 +65,7 @@ Konnect/
 │   │           ├── sch_export.rs     # 6 tools (SVG/PDF/netlist/ERC)
 │   │           ├── sch_hierarchy.rs  # 12 tools (typed Sheet model, sheet CRUD + hierarchy/page queries + pin lifecycle)
 │   │           ├── pcb_board.rs      # 11 tools (S-expr file editing, IPC fallback, SVG logo import)
-│   │           ├── pcb_components.rs # 13 tools (IPC real-time via NNG+protobuf)
+│   │           ├── pcb_components.rs # 13 tools (IPC real-time + safe headless single-placement fallback)
 │   │           ├── pcb_routing.rs    # 12 tools (traces, vias, nets, netclasses)
 │   │           ├── pcb_export.rs     # 13 tools (Gerber, PDF, 3D, DRC, DXF/GenCAD/IPC-2581/ODB++)
 │   │           ├── library.rs        # 17 tools (symbol/footprint library management)
--- a/DEV.md
+++ b/DEV.md
@@ -68,9 +68,10 @@ Konnect/
 │   │           ├── pcb_components.rs # 13 tools (IPC real-time via NNG+protobuf)
 │   │           ├── pcb_routing.rs    # 12 tools (traces, vias, nets, netclasses)
 │   │           ├── pcb_export.rs     # 13 tools (Gerber, PDF, 3D, DRC, DXF/GenCAD/IPC-2581/ODB++)
-│   │           ├── library.rs        # 16 tools (symbol/footprint library management)
+│   │           ├── library.rs        # 17 tools (symbol/footprint library management)
 │   │           ├── footprint_graphics.rs # footprint primitive validation, inspection, and atomic edits
 │   │           ├── footprint_metadata.rs # footprint description, tags, and attribute edits
+│   │           ├── footprint_models.rs # footprint 3D model validation and atomic edits
 │   │           ├── integration.rs    # 9 tools (JLCPCB SQLite, Freerouting, datasheets)
 │   │           ├── verification.rs   # 8 tools (DRC, design rules, KiCAD UI)
 │   │           ├── config.rs         # 7 tools (user/project config, design rules)
@@ -208,7 +209,7 @@ if !path.exists() {
 
 Adding a new kind: edit `mcp/error.rs`, add the variant, add the match arm in `short_code()`, use it from the handler. The `short_code_matches_serialized_kind_field` test will fail loudly if they drift.
 
-The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 190 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
+The dispatch-level errors (not-loaded/unknown/handler-panic) are fully structured. So are **all missing-argument errors** across all 191 tools — `tools/mod.rs::require_str` / `require_f64` emit `ToolErrorKind::InvalidArgument { field, reason }` automatically. Most in-handler errors still use `CallToolResult::error("free text")` or bubble `anyhow::Error`; migrating them is incremental. `project.rs::handle_get_project_info` demonstrates the structured `FileNotFound` pattern.
 
 ## Observability
 
@@ -229,7 +230,7 @@ Source: [`crates/konnect-core/src/observability.rs`](crates/konnect-core/src/obs
 
 ## Tool Routing (Starter Kit + On-Demand Loading)
 
-The server does NOT expose all 190 tools (196 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
+The server does NOT expose all 191 tools (197 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
 
 - **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 6 meta-tools, baseline `tools/list` is ~19 tools ≈ 2K tokens.
 - **On demand**: the LLM reads `list_toolboxes` → calls `load_toolset(name)` to expose a toolset's tools in subsequent `tools/list` responses. `unload_toolset(name)` prunes them when the task shifts.
@@ -301,9 +302,9 @@ convention for other `kicad-cli`-calling code.
 
 ## Current Stats
 
-- **18 toolsets, 190 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
+- **18 toolsets, 191 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
 - Baseline `tools/list`: ~19 tools / ~2K tokens (starter kit + meta-tools)
-- Full-catalog `tools/list` (all loaded): 196 tools (190 registered + 6 meta) / ~25K tokens
+- Full-catalog `tools/list` (all loaded): 197 tools (191 registered + 6 meta) / ~25K tokens
 - **0 IPC stubs** (all protobuf methods implemented)
 - **0 unimplemented tools**
 - **3 CLI commands removed in KiCAD v10** (specctra DSN/SES, pcb sync — return clear errors)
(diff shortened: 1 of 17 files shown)


TWT233:
Re-submitting for review after the latest update.

Added `create_footprint` support for explicit pad layers, rotation, and roundrect ratio while preserving legacy defaults. Fresh local workspace tests/clippy/fmt pass, all 7 GitHub CI checks are green, and the deployed binary generated an external 21-footprint LH60 library whose clean KiCad 10 coupon reports 0 violations / 0 unconnected items. No unresolved review threads.

mixelpixx:
Reviewed and merging. This is a lot of work and it is well built — the six-step recipe from #200 (`read_consistent` → `find_direct_child_blocks` → targeted `SexpEdit` → `parse_sexp` validate → `write_atomic_if_unchanged`) is followed throughout, and all four count locations moved together, which is the thing that usually gets missed.

**On #205 vs #160.** I'd assumed these overlapped; they don't. #160 edits graphics on a *placed board footprint over IPC*; yours edits the *library file* with a layer+mode selector, which is the selector design issue #200 explicitly chose over a UUID one ("a named group/tag selector … requires group ownership and membership lifecycle in the first release"). Different target, different transport. I've asked Pytonballoon810 to rebase #160's IPC half on top of this and drop the duplicated library commit.

**Two things I did on your branch** (thanks for `maintainerCanModify`):

- **Resolved the count collision with #206**, which landed an hour ago. Neither side's number was right: three library tools here plus one in `sch_components` there makes **192 registered + 6 meta = 198**. `doc_tool_counts.rs` checks the arithmetic now, so a wrong rebase fails loudly.
- **Replaced `tools.insert(2|3|4, …)`** with an insert anchored on `create_footprint`'s name. The grouping intent is preserved; the literal indices would silently reorder the catalogue the first time a tool moves in that list.

**What I verified rather than took on faith:**

- **`fix(verification): persist KiCad design rules safely` is the same bug I fixed this morning in #190** — design rules, like netclasses, live in `.kicad_pro`, not the board. You found it independently. I checked your JSON path against KiCad's own `cm5_minima` demo: `board.design_settings.rules` with `min_clearance`/`min_hole_to_hole`/etc. matches exactly, and `.kicad_dru` is right for custom rules.
- **`fix(library): emit canonical KiCad 10 symbols`** — compared against stock `Device.kicad_symdir/R.kicad_sym`: `version 20251024`, `generator_version "10.0"`, `exclude_from_sim`/`in_pos_files`/`duplicate_pin_numbers_are_jumpers`, `(show_name no) (do_not_autoplace no)` on properties, and an empty `Datasheet` rather than `~`. All match.
- **`place_component` hardening preserves the #193 gate** — only `IpcFailure::Unreachable` falls back to the file; a `Rejected` still refuses without touching it. The `write_atomic` → `read_consistent` + `write_atomic_if_unchanged` swap and the structural duplicate-reference check are both improvements.
- **Live against KiCAD 10.0.5**, end to end through the real server: created a footprint, `set_footprint_graphics` replaced the `F.SilkS` layer with a rect and a circle, placed it on a board, and `kicad-cli pcb drc` parsed the result — the graphics survive into the placed instance, and `get_footprint_info(include_graphics)` reads them back. Duplicate reference correctly refused.

**One behaviour change worth calling out in the notes** (I'll carry it into the release): `replace_reference` **bails** when a footprint has neither `(property "Reference"` nor `(fp_text reference`, where `replace_property_value` used to silently no-op and place with the library default. That's the right direction — fail closed — but it is a visible change for malformed library data.

**One nit I checked and am leaving alone:** `create_symbol` writes `(pin_numbers hide)` where KiCad 10 writes `(pin_numbers (hide yes))`, while `pin_names` right beside it uses the sub-node form. I rendered both to SVG through `kicad-cli sym export svg` — **byte-identical output**, so KiCad parses them the same and this is cosmetic, not a defect. Worth aligning some day for clean diffs against KiCad-saved files; not worth holding this.

**Scope note for next time:** this grew from #200's three tools to eight features across 19 commits — pad renumbering, metadata, 3D models, library upsert, bottom-side pads, `place_component`, design rules, canonical symbols. The commit history is clean and individually scoped, which is why it was reviewable at all, but a series this broad is easier to land as two or three PRs. The `place_component` and design-rules commits in particular are unrelated to footprint graphics and each touch code with its own history.

Gate green: fmt, clippy `--all-targets -D warnings`, **738 tests**, doctests. Ships in v0.4.0. Thank you — and please do open the follow-ups you mentioned.

Issue #202, opened 2026-08-15
bulk_move_schematic_components leaves Reference and Value text at the old coordinates


triglav-modular:
**Environment:** Konnect v0.3.2 (`serverInfo` reports `{"name":"konnect","version":"0.3.2"}`), KiCad 10.0.5, macOS 15 (arm64), server driven over stdio.

`bulk_move_schematic_components` moves the symbol but leaves its Reference and Value fields at their old absolute coordinates, so the text is stranded where the part used to be.

## Repro

```python
bulk_move_schematic_components(sch, references=["R999"], dx=15.24, dy=10.16)
```

Measured from the file before and after:

| | before | after | delta |
|---|---|---|---|
| symbol `(at …)` | (279.40, 30.48) | (294.64, 40.64) | +15.24, +10.16 |
| `Reference` property | (279.40, 26.67) | (279.40, 26.67) | **0.00, 0.00** |
| `Value` property | (279.40, 34.29) | (279.40, 34.29) | **0.00, 0.00** |

**Expected:** field positions translate with the symbol.
**Actual:** they keep their old coordinates.

The tool reports the move correctly, including the v0.3.0 per-unit `placements` array, so the return value gives no hint:

```
{'dx': 15.24, 'dy': 10.16, 'errors': [],
 'moved': [{'reference': 'R999',
            'placements': [{'old_x': 279.4, 'old_y': 30.48,
                            'new_x': 294.64, 'new_y': 40.64}]}]}
```

## Impact

Netlist and ERC are both unaffected, so this is invisible to every automated check — it only shows up in a render or the editor. Relocating a 14-part block left reference designators scattered across the far side of an A4 sheet while the graphics sat correctly in their new position.

The only workaround I found is `batch_delete_schematic_components` followed by re-adding at the new coordinates, which re-lays the fields out correctly but loses the footprint field and any annotations, so both have to be re-applied.

Possibly related to the field-anchor work in v0.3.2 ("Symbol Fields Were Placed at Fixed Offset") — that fixed *placement* anchors, but the move path still appears to translate only the symbol node.

mixelpixx:
Fixed on `main` in de70351 (auto-closed by the push). Confirmed exactly as reported: `bulk_move` took only the **first** `(at …)` in the symbol block — the symbol's own placement — and never visited the property anchors, which are absolute coordinates in `.kicad_sch`.

[...]

PR #206, opened 2026-08-15
feat(sch): add reset_schematic_field_positions

pauliuszaleckas:
#186 made `add_schematic_component` place Reference and Value at the anchors their
library defines, and said outright that repairing sheets written before it was a
separate concern. This is that concern.

Every symbol on an older sheet keeps the old fixed `y ± 3.81` offset: a connector's
reference sits inside its body, a rail's name below an up-pointing arrow. Nothing on
the tool surface could repair them — `move_schematic_component` translates the fields
along with the symbol, `edit_schematic_component` changes a field's value and not its
position, and `update_symbols_from_library` refreshes `lib_symbols` without touching
instances. The only fix was KiCad's own **Edit → Reset field text positions**, by hand.

`reset_schematic_field_positions` is that command as a tool: for each instance, put
Reference and Value back on the library anchor, carried through the symbol's rotation
and mirror — the same placement transform #186 introduced, called on existing symbols
instead of new ones.

### Behaviour

- **`dry_run`** reports what would move without writing.
- **`references`** filters to a subset; omitted, it resets every symbol on the sheet.
- **References it did not find are reported**, rather than the tool silently doing
  nothing and claiming success.
- **Idempotent** — a second call moves nothing.
- A symbol whose library definition declares no anchor is left alone, matching the
  fallback #186 kept.

### Scope of the write

Only each field's `(at …)` is rewritten. Field values, symbol positions, wires and
junctions are untouched: on a 27-symbol sheet the netlist exported before and after is
byte-identical.

### Related

Not a fix for #202 (`bulk_move_schematic_components` leaving field text behind), which
is a bug in the move itself. It does give a way to repair a sheet that has already hit
it.

### Testing

`update_symbols_from_library` tests fail on my machine both with and without this
branch — pre-existing on `main`, unrelated to this change.

Two new unit tests: stale fields are put back on their anchors and a second run
reports nothing left to move, and an unknown reference comes back in `not_found`.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -58,7 +58,7 @@ Konnect/
 │   │           ├── cli.rs            # kicad-cli v10 subprocess wrapper (verified against actual binary)
 │   │           ├── svg_import.rs     # SVG parsing + Bezier flattening for import_svg_logo (usvg-backed)
 │   │           ├── project.rs        # 6 tools (incl. open_schematic_viewer)
-│   │           ├── sch_components.rs # 18 tools (component placement with lib_symbols embedding)
+│   │           ├── sch_components.rs # 19 tools (component placement with lib_symbols embedding)
 │   │           ├── sch_wiring.rs     # 19 tools (incl. connect_pins, power symbol embedding)
 │   │           ├── sch_analysis.rs   # 15 tools (union-find net graph, connectivity)
 │   │           ├── sch_batch.rs      # 12 tools (single-read/single-write atomic operations)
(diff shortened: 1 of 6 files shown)


mixelpixx:
Merged. I pushed a merge of current `main` plus two small fixups to your branch (thanks for `maintainerCanModify`).

**Why the merge mattered here:** this tool writes through `Schematic::overwrite()`, and until #209 landed this morning that path truncated `(paper "User" W H)` to `(paper "User")` on every call — a sheet KiCad then refuses to load (#208). Merging #209 first was the ordering constraint for your PR, #212 and #213 alike. I re-checked it end-to-end: a `User`-paper sheet keeps its dimensions through a reset now.

**Two fixups:**

- **`not_found` came out of a `HashSet` unsorted**, so a caller asking about several unknown references got a different ordering each call. Now sorted, with a test that repeats the call eight times — an unsorted list passes once and then doesn't.
- **A field the library anchors but the placed symbol doesn't carry was dropped with a bare `continue`** — absent from `moved`, `unchanged` and `no_library_anchor` alike, so the caller couldn't tell it had been skipped. It's now reported in a `no_property` list. Both fixup tests fail with the fix neutered.

Your own tests are good and I want to note why: asserting the exact post-values `(at 103.632 50.8 90)` rather than "it changed", checking `dry_run` left the file alone, and the idempotence pass are all doing real work. The anchors-before-mutation borrow comment is also the kind of thing that saves the next reader ten minutes.

Verified end-to-end through the real server as well: dry run reports both fields and writes nothing, the apply puts Reference on its library anchor at (103.632, 50.8, 90), `not_found` comes back sorted, and `kicad-cli` loads the result.

Counts moved together as required — 189 registered, `sch_components` 19, `protocol_stdio` batch 38. Ships in v0.4.0.

Issue #120, opened 2026-08-01
Component and label moves don't add or prune junction dots


mixelpixx:
Follow-up from #104, which made placement and wire-drawing junction-aware but left the move paths alone.

After #104, `add_schematic_component`/`add_power_symbol`/`add_wire` insert a junction when a pin lands mid-segment on a wire. The move tools (`move_schematic_component`, `move_labels_by_offset`, `bulk_move_schematic_components`, from #102) do not, so:

- Moving a component **onto** a wire leaves the pin mid-segment with no junction → KiCad treats it as unconnected, silently.
- Moving a component **off** a wire leaves a stale junction dot behind. Harmless on its own, but at a point where two wires cross it now shorts them.

The second case is the concerning one: a move can create a connection the user didn't ask for.

Fix: after a move writes, run the same `add_pin_midwire_junctions` pass for the moved references, plus a prune step that drops junctions no longer justified by a pin or a wire T. The prune needs care — it must not remove junctions the user placed deliberately with `add_junction`, so it should only consider dots this tool would have created.

Related: #104's file shape is already non-canonical (eeschema always splits a wire rather than leaving a junction mid-segment — 0 of 4293 junctions in KiCad's demo corpus sit mid-segment). If we ever switch to split+junction, the move paths need to split/rejoin too, which is a good reason to settle that question before building the prune.

mixelpixx:
Narrowed by #214, not closed by it.

#214 (merged) prunes the dots a **deleted wire** leaves unjustified, and reworks `split_wire_at_point` so its two halves keep the dots they should. That covers the case where relocating a block deletes wires and re-adds them elsewhere.

What is still open here is what this issue was actually about: **moves that strand or require a dot without any wire being deleted.** Moving a component or a label changes which pins and wire ends coincide, and nothing currently re-evaluates junctions on that path. So:

- a move that takes a pin off a wire mid-segment leaves a dot with nothing to justify it
- a move that lands a pin on a wire mid-segment needs a dot and does not get one

The second is the one that bites, because it is a missing connection rather than a cosmetic leftover — the schematic reads as connected to a human and is not connected to ERC.

The machinery to do it now exists: `prune_orphaned_junctions` in `sch_wiring.rs` already answers "is this dot justified?" against wires and mid-segment pins. A move path would need the mirror of it — re-evaluate the dots at both the old and new coordinates — rather than anything new.

[...]

PR #212, opened 2026-08-15
fix(schematic): emit one junction dot per T, not one per wire

### LINKED ISSUE #201: add_wire re-stamps a junction at an existing junction point on every later call
**Environment:** Konnect v0.3.2 (`serverInfo` reports `{"name":"konnect","version":"0.3.2"}`), KiCad 10.0.5, macOS 15 (arm64), server driven over stdio.

Not the same as #120, which is about *moves* failing to add or prune junction dots. This is `add_wire` writing an additional `(junction)` at a coordinate that already has one — once for every later `add_wire` call, however unrelated.

## Repro

On any sheet, in empty space:

```python
add_wire(sch, x1=240.03, y1=20.32, x2=240.03, y2=30.48)   # vertical
add_wire(sch, x1=229.87, y1=20.32, x2=250.19, y2=20.32)   # T onto the first wire's endpoint
# -> 1 junction at (240.03, 20.32).  Correct.

add_wire(sch, x1=259.08, y1=40.64, x2=269.24, y2=40.64)   # unrelated, elsewhere
add_wire(sch, x1=259.08, y1=45.72, x2=269.24, y2=45.72)   # unrelated
add_wire(sch, x1=259.08, y1=50.80, x2=269.24, y2=50.80)   # unrelated
# -> 4 junctions stacked at (240.03, 20.32)
```

**Expected:** one junction at that point.
**Actual:** one additional junction per subsequent `add_wire`, at a coordinate those wires never touch.

## Why it's easy to miss

Connectivity is unaffected, so ERC passes and the exported netlist is byte-identical. Nothing surfaces until the file is inspected. In a real session adding 24 wires to one sheet, this left **23 stacked junctions on a single coordinate**.

A plain before/after count won't catch it either — the give-away is entries versus *distinct* coordinates:

```python
js = re.findall(r'\(junction\s*\(at ([-\d.]+) ([-\d.]+)\)', s)
dupes = {k: v for k, v in Counter(js).items() if v > 1}
```

Deleting a wire also leaves its junction behind, so relocating a block strands orphans — junctions with no wires passing through them at all.

pauliuszaleckas:
`find_t_junctions` reports every T on the sheet, not only the ones the new wire made, so
`add_wire`, `batch_add_wire` and `connect_to_net` re-emitted a junction dot at every T
already there — linear over a sequence of calls, quadratic in a batch. Laying out 100
wires over 6 T-points produced **445** junction entries where 6 were needed.

Fixes #201.

### The fix

The guard already existed twice in tree: three lines below each of these loops, on the
pins the new wire passes over (`1866129`), and inside `insert_wire_with_junctions` on the
string path (`620da07a`). Neither reached the T loop sitting beside it.

One `add_missing_junctions(&mut sch, &positions)` helper now covers both kinds of
junction at all three call sites, so each handler also loses its hand-rolled copy of the
pin guard. `connect_to_net`'s stub had the same defect and is not in #201.

`connect_pins`, `add_schematic_connection` and `batch_connect_pins` route through
`insert_wire_with_junctions` and were never affected.

Not routed through `insert_wire_with_junctions` as #201 suggests: that helper edits
schematic text, while these three handlers work on a loaded `cse::Schematic` and
`overwrite()` it. The guard reproduces its semantics — dedupe against the junctions
already in the sheet — without straddling both write paths in one handler.

Junctions still are not *removed* when the item that justified one moves away; that half
is #120 and is untouched here.

### Testing

`update_symbols_from_library` tests fail both with and without this branch — pre-existing
on `main`, unrelated.

Two new tests draw a rail plus three taps and assert one dot per T, once through repeated
`add_wire` calls and once through a single `batch_add_wire`. With the guard removed both
see three dots on the first T.

### DIFF
--- a/crates/konnect-core/src/tools/sch_wiring.rs
+++ b/crates/konnect-core/src/tools/sch_wiring.rs
@@ -430,6 +430,24 @@ fn pins_mid_segment(pins: &[(f64, f64)], x1: f64, y1: f64, x2: f64, y2: f64) -> 
         .collect()
 }
 
+/// Add a junction dot at each position that does not already carry one.
+///
+/// `find_t_junctions` reports every T on the sheet, not only the ones the new
+/// wire made, so an unguarded loop re-emits a dot at every existing T on every
+/// call — quadratic in a batch. `insert_wire_with_junctions` guards the same
+/// way on the string path.
+fn add_missing_junctions(sch: &mut cse::Schematic, positions: &[(f64, f64)]) {
+    for &(x, y) in positions {
+        if !sch
+            .junctions
+            .iter()
+            .any(|j| konnect_sexp::geometry::points_coincident(x, y, j.x, j.y, 0.01))
+        {
+            sch.add_junction(x, y);
+        }
+    }
+}
+
 pub(crate) fn insert_wire_with_junctions(
     content: String,
     x1: f64,
@@ -541,21 +559,11 @@ async fn handle_add_wire(
     let junctions = find_t_junctions(&existing_wires, 0.01);
 
     sch.add_wire(x1, y1, x2, y2);
-    for (jx, jy) in &junctions {
-        sch.add_junction(*jx, *jy);
-    }
+    add_missing_junctions(&mut sch, &junctions);
     // Pins the new wire passes over mid-segment also need junction dots.
     let (_, tree) = read_schematic(&sch_path)?;
     let pins = crate::tools::all_pin_endpoints(&tree);
-    for (px, py) in pins_mid_segment(&pins, x1, y1, x2, y2) {
-        if !sch
-            .junctions
-            .iter()
-            .any(|j| konnect_sexp::geometry::points_coincident(px, py, j.x, j.y, 0.01))
-        {
-            sch.add_junction(px, py);
-        }
-    }
+    add_missing_junctions(&mut sch, &pins_mid_segment(&pins, x1, y1, x2, y2));
     sch.overwrite()?;
 
     Ok(CallToolResult::json(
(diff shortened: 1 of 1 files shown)


triglav-modular:
I filed #201 — this matches what I measured, and the guard covers the case I reported. Confirming the diagnosis from the outside: adding 24 wires to one sheet in a single session left **23 junction entries stacked on one coordinate**, and the count grew by exactly one per subsequent `add_wire`, however unrelated that wire was. Good catch on `connect_to_net`; I only exercised `add_wire`, so that one wasn't in my report.

One thing worth not losing when #201 closes.

#201 also mentioned a **third** case, which this PR correctly leaves alone but which I don't think #120 covers either: **deleting a wire leaves its junction behind.**

#120 is scoped to component and label *moves*. Wire deletion is a separate path, and it bites in ordinary use — relocating a block means deleting its wires and re-adding them at the new position, which leaves junctions sitting at the old coordinates with no wires passing through them at all. I had to find and delete two of those by UUID after one move.

It's cheap to detect unambiguously, since a junction needs at least two wires through it to be meaningful:

```python
through = sum(1 for x1, y1, x2, y2 in wires
              if min(x1, x2) - eps <= jx <= max(x1, x2) + eps
              and min(y1, y2) - eps <= jy <= max(y1, y2) + eps)
# through == 0  ->  orphaned, nothing justifies it
```

Worth being careful about the boundary: `through == 1` with no wire *endpoint* at that point is legitimate — that's a pin landing mid-segment, exactly what the `pins_mid_segment` guard here creates. So only the zero case is safe to prune blindly.

Not suggesting it belongs in this PR — the add path and the remove path are cleanly separable and this one is nicely scoped. Just flagging that it may fall between #201 and #120 rather than inside either, in case it's worth its own issue.

PR #215, opened 2026-08-15
feat(pcb): add atomic schematic-to-board update

### LINKED ISSUE #187: [FEATURE] Add atomic update_pcb_from_schematic workflow
## Backlog priority and status

P1 #6 — workflow completeness and operational reliability.

**Accepted for v0.4.0.** The documentation correction and registry guard shipped in v0.3.2; the actual synchronization feature remains to be implemented. This body incorporates the maintainer's final contract ruling in [the accepted design response](https://github.com/mixelpixx/Konnect/issues/187#issuecomment-5300322819).

## Problem

Konnect has no public operation that carries a completed schematic into PCB layout. The standard schematic → board workflow therefore stops at the handoff where footprints and nets must be synchronized.

This cannot be a convenience wrapper: a partial or non-atomic implementation can destroy placement, disconnect routed copper, add excluded parts, or claim success after silently skipping design content.

## Evidence and precedent

The original report verified that no synchronization tool was registered, while `validate_for_manufacturing` already directed the user to KiCad's **Tools → Update PCB from Schematic** because KiCad 10 removed `kicad-cli pcb sync`. Konnect already has the required transactional primitive, `run_commit`.

The sibling [KiCAD-MCP-Server](https://github.com/mixelpixx/KiCAD-MCP-Server) establishes this as one required workflow-level capability, but its implementation history also supplies the failure cases this implementation must avoid:

- hierarchy was added after a root-only implementation failed ([PR #120](https://github.com/mixelpixx/KiCAD-MCP-Server/pull/120));
- missing footprints arrived later, with deletion deliberately out of scope ([PR #171](https://github.com/mixelpixx/KiCAD-MCP-Server/pull/171));
- repeated footprint resolution caused major performance problems ([issue #248](https://github.com/mixelpixx/KiCAD-MCP-Server/issues/248));
- reference-only matching can create duplicates ([issue #250](https://github.com/mixelpixx/KiCAD-MCP-Server/issues/250));
- incomplete extraction can be reported as partial success, new footprints stack at `(0, 0)`, and the operation has no reviewed plan/transaction boundary.

This is process precedent, not code to copy. Konnect will implement one cohesive Rust workflow using its existing KiCad CLI and official IPC modules.

## Public interface

Add one tool, `update_pcb_from_schematic`.

Inputs:

- `schematic`: saved root `.kicad_sch` path;
- `board`: target `.kicad_pcb` path, which must be the board active in pcbnew;
- `dry_run`: defaults to `true`;
- `expected_plan_revision`: required when applying and returned by dry run.

There is no `force` flag and no general override policy in v1. The dry-run plan is the review step; arbitrary partial application is unsupported because it can leave an electrically incomplete board while appearing intentional.

Output mirrors the established aggregate response shape:

- `status`:
  - dry run: `ready` | `noop` | `conflict`;
  - apply: `applied` | `noop` | `conflict`;
- `coverage`: paired planned/applied counts per category;
- `diagnostics`: structured entries for every conflict or unevaluable item.

The tool never returns the string `synced` and has no partial-success state.

## Source of truth and preflight

- Reuse `cli::export_netlist` with format `kicadsexpr`; do not add an XML path or reconstruct connectivity geometrically.
- The hierarchy must be saved because the export reads the on-disk schematic. This is enforced, not merely documented.
- A nonzero export, zero-component netlist, parse failure, unresolved footprint library reference, incomplete hierarchy, or ambiguous identity is a conflict.
- Resolve each distinct footprint definition once while planning.
- Read footprint identity/reference/net state, routed track/via/zone membership, and current board/item bounds before mutation.
- Compute staging positions during planning and retain the reviewed positions for apply.

`plan_revision` hashes exactly the state the plan read:

- netlist export content;
- board footprint identity/reference/net set;
- routing net membership consulted by conflict detection;
- bounds used for the staging strip.

Apply recomputes the plan and rejects a mismatch with a diagnostic that tells the caller to rerun dry run.

## Live IPC and atomicity

This tool is live-IPC-only permanently, not merely for v1. It enters through `attempt_ipc_write`, which calls `ensure_board_is_active` before any operation. If KiCad is unreachable, `BoardWrite::File` is converted to `conflict`; there is no S-expression/file fallback. Any future offline synchronization must use a different tool name and state its non-atomic contract explicitly.

All fallible non-mutating work happens before `begin_commit`: export, parsing, footprint resolution, net-code resolution, conflict checks, item construction, and revision verification.

Every mutation then runs inside exactly one `run_commit("Update PCB from schematic", ...)` and every overall/per-item failure propagates with `?`. No failure may be swallowed. A rejection drops the commit; successful apply creates one pcbnew undo entry. No file I/O or schematic write occurs inside the commit.

The response explicitly tells the user that one Ctrl-Z in pcbnew reverses the whole applied synchronization.

## Conflict policy

| Case | Required v1 behavior |
| --- | --- |
| Footprint moved/rotated/flipped/locked on the board | Board wins. Preserve KIID, X/Y, side, rotation, lock state, and artwork permanently. |
| In schematic, absent from board, footprint ID resolvable | Add at its reviewed deterministic staging position. |
| On board, absent from schematic | Preserve and report as board-only. Never delete. |
| Footprint ID changed in schematic | Conflict; do not replace or swap the board footprint. |
| `on_board = no`, absent from board | Skip and count in `skipped`. |
| `on_board = no`, already on board | Conflict; do not silently delete. |
| DNP | Keep/add, mark DNP, and assign nets normally. |
| `in_bom = no` | No board effect. This tool does not reconcile BOM membership. |
| Pad net changed and no routed copper/zone membership is affected | Reassign. |
| Net change would affect routed track, via, or zone membership | Conflict and refuse the entire apply. |
| Reference changed and matched by stable schematic path | Update the board reference. |
| Reference changed but only reference matching is available | Conflict; rename versus new component is ambiguous. |
| Duplicate reference/identity or multiple schematic instances matching one board footprint | Conflict. |

None of these defaults is overridable in v1.

## Staging

Missing footprints are placed in a deterministic, non-overlapping strip outside current board/item bounds. The origin and every footprint position are computed and stored in the dry-run plan, then reused unchanged during apply. The response returns each staged position so existing placement tools can consume it.

## Definition of success

Coverage reports paired `planned` and `applied` counts for:

- footprints added;
- footprints updated;
- pads reassigned;
- board-only footprints preserved;
- schematic instances skipped by `on_board = no`;
- conflicts.

Apply returns `applied` only when every planned mutation is accepted, every category has `applied == planned`, and conflicts are zero. If any category differs, the commit has been dropped, status is `conflict`, and the board is unchanged. Intentional exclusions are classified and counted; unresolved or skipped work can never be hidden inside success.

Idempotence is required: apply followed by dry run returns `noop` with every change count zero.

## Acceptance criteria

- Dry run does not mutate KiCad or the board file and reports the complete reviewed plan and staging positions.
- Apply requires the exact reviewed revision and produces one undo entry.
- Any per-item rejection propagates, drops the entire commit, and returns `conflict`; injected mid-apply failure proves no partial board state survives.
- Existing footprint placement, orientation, side, lock state, KIID, artwork, routing, vias, zones, and graphics are preserved.
- Missing eligible footprints and nets are added; safe existing pad/net assignments and schematic links are updated.
- Board-only footprints remain untouched and are reported.
- Full hierarchy, repeated sheet instances, multi-unit symbols, unlabeled nets, DNP, `on_board`, and `in_bom` semantics are covered.
- Tests cover blank, partially populated, already synchronized, changed footprint ID, renamed net, unresolved footprint, duplicate identity/reference, stable-path rename, stale plan, routed-copper conflict, and injected mid-apply rejection.
- Multiple components using the same footprint resolve that definition once per plan.
- A KiCad-backed E2E compares resulting connectivity with KiCad's export, proves apply → dry run is `noop`, and verifies the single undo transaction.
- The tool must be registered before shipped prose names it. Tool counts, `tool-directory.md`, bundled skill text, README/DEV totals, and public response documentation change together under the v0.3.2 guards.

## Reviewable implementation sequence

1. Pure plan computation and tests.
2. IPC apply through `attempt_ipc_write` and `run_commit`, with failure/rollback tests.
3. Tool registration, skill/docs/count updates, and KiCad-backed E2E coverage.

A focused PR with these as clean commits is preferred.

neusse:
## Summary

Closes #187.

Adds `update_pcb_from_schematic` to the `sch_export` toolset as the accepted live-KiCad, plan/apply workflow:

- exports the saved hierarchy with `kicad-cli sch export netlist --format kicadsexpr`
- returns `ready`, `noop`, or `conflict` for dry runs and requires `expected_plan_revision` for apply
- preserves board-owned footprint identity, placement, side, rotation, lock state, and artwork
- stages resolvable missing footprints deterministically outside the reviewed board bounds
- preserves and reports board-only footprints; rejects footprint swaps, ambiguous matches, invalid exports, unsaved hierarchy, and routed-net reassignment
- applies all mutations through `attempt_ipc_write` in one `run_commit`, with no file fallback and one KiCad undo entry
- reports paired planned/applied coverage plus structured diagnostics

## Design and compatibility

The planner is a pure module with focused tests. Export, hierarchy validation, footprint resolution, plan construction, revision validation, and routing preflight all happen before the commit. Only IPC mutations run inside the single commit, and every per-item result propagates with `?` so any failure drops the whole batch.

This adds one tool without changing existing tool inputs or behavior. Registry/documentation totals move to 189 registered tools plus 6 meta tools (195 total). README, DEV guide, tool directory, troubleshooting, and the bundled KiCad PCB skill are updated together.

## Verification

- live KiCad 10.0.5 hierarchical round trip on a disposable project:
  `cargo test -p konnect --locked --test live_kicad_tools schematic_sync_apply_then_dry_run_is_noop -- --ignored --exact --nocapture`

The live test verified dry-run `ready` -> apply `applied` -> dry-run `noop` against an open board and saved hierarchical schematic.

## Risk and rollback

The tool is deliberately live-IPC-only. A missing/rejected IPC connection returns a conflict instead of touching the board file. Routed-net changes and unevaluable inputs fail closed before `begin_commit`; apply mutations share one KiCad commit, so failures drop the batch and Ctrl-Z reverses a successful sync as one action.


mixelpixx:
Reviewed against the contract, and this implements it faithfully — I checked each ruling in the code rather than against the description:

- **Atomicity** — `attempt_ipc_write` → planning → revision check → *then* `run_commit` wrapping only `create_items_in`/`update_items_in`, both with `?`. Nothing that can fail without mutating happens inside the commit. `BoardWrite::File` returns a conflict rather than falling back, which was the rule I said should hold permanently rather than "in v1".
- **Never swallow a per-item status** — I checked the new `_in` variants specifically, since that was the failure mode I was most worried about. `update_items_in` carries the `ItemStatusCode` check and additionally bails on a *missing* status, which is stricter than the old path.
- **Conflict policy** — `footprint_id_changed`, `on_board_exclusion_conflict`, `routed_pad_net_change`, `reference_identity_conflict`, `reference_only_rename_ambiguous`, `reference_rename_collision`, the `duplicate_*` family. DNP is compared and carried onto the board; `in_bom` is absent from the module entirely, which is right. **No `force` flag**, and `dry_run` defaults to `true`.
- **Success semantics** — `status`/`coverage`/`diagnostics` in #185's shape, paired planned/applied counts, no partial-success state.

Then I ran your own acceptance test against KiCAD 10.0.5, and **it failed**.

## The blocker: the plan revision changed every second

`plan_revision` hashed the raw netlist source, and `kicad-cli sch export netlist` stamps the export time into every file:

```
5c5
<     (date "2026-08-15T14:48:16")
---
>     (date "2026-08-15T14:48:18")
```

So the revision moved on the clock, not on the design. Two consecutive dry runs on an untouched design produced **identical plans — 68 additions each — and different revisions**. Since apply requires the revision a dry run returned, apply could only ever succeed if both calls landed inside the same wall-clock second. Your live test failed for me at `status: conflict / stale_plan_revision` 1.2 s after its own dry run.

That is why it passed for you and failed for me: it is a race, and the losing case is the one that matters — a human reading the plan before applying it, which is the entire point of having a plan.

**Fixed on your branch** (`a3c0233`): the revision now covers what the plan *reads* — components and nets — with the volatile export header excluded. A nested `(date …)` inside a component's fields is design content and still counts. The tool version is excluded for the same reason: a KiCad upgrade shouldn't invalidate every outstanding plan for an unchanged design. New unit test asserts all four cases, including that a footprint swap still moves the revision.

## Verification after the fix

Disposable copy of KiCad's `complex_hierarchy` demo with its 68 board footprints stripped, schematic untouched:

```
dry run  -> ready,   footprints_added {planned: 68, applied: 0}
apply    -> applied, footprints_added {planned: 68, applied: 68}
dry run  -> noop
```

Your acceptance criterion, passing. Then, independently of the tool under test, I asked KiCad to save its own live state: **the board file went 0 → 68 footprints across 68 distinct references**, and `kicad-cli pcb drc` parses the result with a live ratsnest (164 unconnected items, i.e. the pads carry nets).

## Two notes, neither blocking

- The counts here will need one more rebase — main moved a lot today (#205, #206, #178, #175, #173 landed, #170 pending), so `sch_export` 6 → 7 sits on top of a bigger total than when you branched. I resolved to 196 + 6 = 202 on the branch; `doc_tool_counts.rs` will catch it if #170 lands first.
- `get_component_pads` couldn't see the synced footprints during my verification because it reads the board *file*, not live IPC — which is exactly what your #207 fixes. Unrelated to this PR, but it means anyone verifying a sync by hand today will get a confusing answer. Worth landing #207 close behind this.

Excellent work, and thank you for holding implementation until the contract was settled — it made this reviewable in a way a 1,800-line PR usually isn't. Merging once CI confirms.

neusse:
Follow-up fix pushed in `1bb5a62` after auditing the timestamp correction.

The previous `find("(date ")` / `find("(tool ")` scanner removed every matching substring, including nested component nodes and tool-like text inside quoted values. That contradicted the intended header-only behavior and could make genuinely different design content share a plan revision.

The revision identity now structurally encodes only the parsed top-level `components` and `nets` trees. Export timestamp/tool metadata remains excluded, while all nested nodes and quoted content remain revision-sensitive. A regression test covers both cases; it failed against the scanner and passes with the structural encoding.

Local verification:

mixelpixx:
Took your structural hash over my string-stripping — yours is right and mine had a real hole: `find("(date ")` isn't string-aware, so a component field whose *value* contained that literal would have been mangled. Hashing the parsed `components` and `nets` trees can't have that problem. Merged, and my revision-stability test now runs against your implementation and passes, which is a nice cross-check: my test, your code.

Re-verified live against KiCAD 10.0.5 on a fresh fixture (a second disposable copy of `complex_hierarchy` with its 68 footprints stripped): dry run `ready` → apply `applied` → dry run `noop`, and KiCad saving its own state takes the board file 0 → 68 footprints.

Also rebased the counts — main moved again while this was in review (#170 added the `sch_bus` toolset, #199 landed), so it's 200 registered + 6 meta = 206 across 19 toolsets now.

PR #229, opened 2026-08-16
feat(pcb): support closed-board move and rotation

### LINKED ISSUE #228: Support closed-board move and rotation fallbacks
## Problem

`place_component` can safely fall back to a revision-aware closed-board file edit when KiCad IPC is unreachable, but `move_component` and `rotate_component` still require a live PCB editor.

That leaves headless layout workflows asymmetric: they can place a footprint into a closed board, but cannot adjust its position or orientation without opening KiCad.

## Proposed behavior

Extend the existing `move_component(board, reference, x, y)` and `rotate_component(board, reference, rotation)` tools without changing their schemas:

- use live KiCad IPC when reachable and verify the requested board is active;
- fall back to a closed-board S-expression edit only for `IpcFailure::Unreachable`;
- fail closed with zero file writes for any response classified as `Rejected`;
- locate exactly one footprint by reference;
- preserve the existing angle on move;
- set an absolute root angle on rotate and apply the delta to pad/text child angles;
- validate the complete resulting board and persist with `write_atomic_if_unchanged`.

Missing/duplicate references, malformed boards, and stale-source conflicts must leave the board unchanged.

## Compatibility

No tool names, arguments, or existing IPC behavior change. Successful responses gain a `source` field (`ipc` or `file`), matching `place_component`.

TWT233:
## Summary

Allow `move_component` and `rotate_component` to update a closed board when KiCad IPC is unreachable, matching the existing safe fallback behavior of `place_component`.

Closes #228

## Approach

The handlers now use the typed IPC failure classification instead of the unconditional IPC macro:

- live IPC still targets the requested board and remains the preferred path;
- a reachable KiCad rejection fails closed and never edits the file behind the editor;
- only an unreachable transport invokes the closed-board fallback.

The fallback reads one consistent board revision, finds exactly one footprint by reference, updates only its root `(at ...)` block, validates the complete board, and persists through `write_atomic_if_unchanged`. Moving preserves the current angle. Rotation applies the absolute target angle and propagates the delta to pad/text child angles so KiCad does not report a library-footprint mismatch.

## Compatibility and safety

No public tool names or arguments change. Successful move/rotate responses add `source: "ipc" | "file"`, consistent with `place_component`.

The file fallback is unavailable whenever KiCad answers the request. Missing references, duplicate references, malformed board structure, and stale-source conflicts leave the board untouched. Rollback is a normal revert of this commit; the existing live-IPC path remains unchanged.

## Validation

- [x] `python3 -m unittest discover -s plugin/tests -v`
- [x] `cargo build --release --locked -p konnect`
- [x] Focused closed-board move/rotate tests cover move, rotation, missing reference, live rejection, and stale-source conflict

Environment-dependent real-KiCad checks were not required for the fallback itself because the new path is deliberately gated on IPC being unreachable. Existing IPC tests cover the reachable path and board targeting.

## Review checklist

- [x] No public names or schemas were added or renamed.
- [x] No tools were added or removed, so catalogue counts are unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,8 @@ the main workspace — see [DEV.md](DEV.md) for build steps.
   tracked on the [roadmap](ROADMAP.md))
 - `kicad-cli` (ships with KiCAD — used for exports, ERC, DRC)
 - For most PCB tools: KiCAD running with the target board open (IPC API).
-  `place_component` can safely fall back to a closed board file when IPC is unreachable.
+  `place_component`, `move_component`, and `rotate_component` can safely fall
+  back to a closed board file when IPC is unreachable.
 
 ## License: free for the little guys
 
@@ -302,8 +303,8 @@ manual copy), then restart KiCAD.
 with that board first; most PCB tools talk to the running PCB editor. Only an
 unreachable KiCAD produces that message: if KiCAD is running and the tool
 refuses anyway, the error is the tool's own reason for refusing.
-`place_component` alone can fall back to a closed board file when no KiCAD
-process is reachable.
+`place_component`, `move_component`, and `rotate_component` can fall back to a
+closed board file when no KiCAD process is reachable.
 
 See [docs/TROUBLESHOOTING.md](docs/TROUBLESHOOTING.md) for socket setup, tools
 that don't appear after `load_toolset`, and transaction recovery.
--- a/crates/konnect-core/src/tools/pcb_components.rs
+++ b/crates/konnect-core/src/tools/pcb_components.rs
@@ -2733,6 +2733,81 @@ mod tests {
         assert!(moved.contains("(at 1 2 247.5)"), "{moved}");
     }
 
+    /// A reference naming two footprints identifies none of them, and a board
+    /// Konnect cannot parse is not one it will write to. Both refuse, both
+    /// with a structured reason, and both leave the file byte-identical.
+    ///
+    /// Added because neutering each guard found nothing: the PR describes all
+    /// three behaviours — missing reference, duplicate reference, unusable
+    /// board — and only the missing one was exercised, for move or rotate.
+    #[tokio::test]
+    async fn a_duplicate_reference_or_an_unparseable_board_refuses_without_writing() {
+        let tmp = tempfile::tempdir().unwrap();
+
+        let footprint = "  (footprint \"Test:R\"\n    (layer \"F.Cu\")\n    (at 10 20 30)\n    \
+             (property \"Reference\" \"R1\" (at 0 -1.65 30) (layer \"F.SilkS\"))\n    \
+             (pad \"1\" smd roundrect (at -0.9125 0 30) (size 1 1) (layers \"F.Cu\"))\n  )";
+        let duplicated = tmp.path().join("duplicate.kicad_pcb");
+        std::fs::write(
+            &duplicated,
+            format!(
+                "(kicad_pcb (version 20241229) (generator \"pcbnew\")\n{footprint}\n{footprint}\n)"
+            ),
+        )
+        .unwrap();
+
+        let truncated = tmp.path().join("truncated.kicad_pcb");
+        std::fs::write(
+            &truncated,
+            "(kicad_pcb (version 20241229)\n  (footprint \"Test:R\"",
+        )
+        .unwrap();
+
+        for (label, board, structured_reference) in [
+            ("duplicate", &duplicated, true),
+            ("truncated", &truncated, false),
+        ] {
+            let before = std::fs::read_to_string(board).unwrap();
+            for tool in ["move", "rotate"] {
+                let mut arguments = json!({
+                    "board": board.to_string_lossy(),
+                    "reference": "R1",
+                });
+                let result = if tool == "move" {
+                    arguments["x"] = json!(40.0);
+                    arguments["y"] = json!(50.0);
+                    handle_move_component(&arguments, &test_ctx())
+                        .await
+                        .unwrap()
+                } else {
+                    arguments["rotation"] = json!(90.0);
+                    handle_rotate_component(&arguments, &test_ctx())
+                        .await
+                        .unwrap()
+                };
+
+                assert!(result.is_error, "{label}/{tool} must refuse");
+                let text = result_text(&result);
+                if structured_reference {
+                    assert!(
+                        text.contains("invalid_argument") && text.contains("reference"),
+                        "{label}/{tool}: {text}"
+                    );
+                } else {
+                    assert!(
+                        text.contains("Refusing to edit the board"),
+                        "{label}/{tool}: {text}"
+                    );
+                }
+                assert_eq!(
+                    std::fs::read_to_string(board).unwrap(),
+                    before,
+                    "{label}/{tool} must leave the board byte-identical"
+                );
+            }
+        }
+    }
+
     /// The same board must survive a move, which does not rewrite children at
     /// all — pinned so a future refactor cannot reintroduce the coupling by
     /// making the move path share the rotate path's ordering.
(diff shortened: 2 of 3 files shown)


mixelpixx:
Reviewed, and the approach is right: `place_component` already had the
revision-aware closed-board fallback, and leaving `move_component` /
`rotate_component` without it made headless layout asymmetric for no good reason.
Gating on typed `IpcFailure::Unreachable` rather than a message match is the
correct gate.

I rebased your branch onto current `main` (v0.6.0) and ran the full gate: one
README conflict, **zero code conflicts**, and green — fmt, `clippy --workspace
--locked --all-targets -- -D warnings`, 840 tests.

Three things to fix before it lands. Chris asked me to take them rather than
round-trip, so I'll push to `TWT233:feat/closed-board-footprint-placement` with
your authorship intact, the way #160 went.

## 1. The rotate splice writes at stale offsets — reproduced

`update_footprint_placement` computes the root `(at …)` byte range against the
original footprint text, then for `Rotate` runs `apply_rotation_to_children` —
which rewrites child `(at …)` blocks and **changes their length** — and only then
splices the root `at` at the now-stale offsets.

Any child `(at …)` that precedes the root one and whose rendered angle changes
width misaligns it. With a property before the root `at`:

```
(footprint "Test:R" (layer "F.Cu")
  (property "Reference" "R1" (at 0 -1.65 -45) (layer "F.SilkS") (effects (font (size 1 1))))
  (at 10 20 30)
  (pad "1" smd roundrect (at -0.9125 0 30) (size 1 1) (layers "F.Cu")))
```

`rotate_component(reference="R1", rotation=75)` takes the property angle −45 → 0,
which `format_at` drops entirely (−4 bytes), and the splice lands four bytes
early:

```
Parse error at offset 0: ... "(at (at 10 20 75) (pad \"1\" smd roundrect ..."
```

A +4 variant — property at `(at 0 -1.65)` gaining an angle — fails with
`placement update changed the footprint root`.

Both fail closed: `persist_board_replacement` is never reached and the board is
untouched. So this is not corruption. But it's a spurious refusal of a file KiCad
accepts, and whether the misaligned splice happens to trip the root check is luck
rather than design.

KiCad's own writer emits `at` before properties and pads, so files KiCad wrote
are safe — but we read boards we didn't write, and token order isn't part of the
format.

**Fix:** splice the new root `(at …)` *first*, on the original string where the
offsets are valid, then run `apply_rotation_to_children` on the result. The two
passes touch disjoint blocks — `apply_rotation_to_children` only rewrites the
first `(at …)` inside `pad`/`property`/`fp_text`, and the root `at` is a direct
child of `footprint` — so neither can disturb the other. Both fixtures above go in
as regression tests.

Worth noting your own `flip_footprint_block` in #230 already gets this right: it
collects every `SexpEdit` against the original string and applies one batch.

## 2. Classify failures by type, not by message text

```rust
Err(error) if error.to_string().contains("not found on the board") => …,
Err(error) => Err(error),
```

`with_ipc_classified`'s doc comment asks for the opposite in as many words —
"never by matching message text" — and the practical cost is that every *other*
caller-facing refusal (duplicate reference, malformed board, "footprint must
contain exactly one root placement") bubbles as `anyhow` and reaches the caller as
`handler_error` with raw text. Missing and duplicate references are
`ToolErrorKind::InvalidArgument { field: "reference" }`; a board that can't be
parsed is a `CallToolResult::error` saying the file was left untouched.
`parse_points` in the same file is the pattern.

## 3. Use `attempt_ipc_write` instead of re-implementing it

`pcb_board::attempt_ipc_write` + `BoardWrite` exists for exactly this shape — a
board-mutating tool with an S-expression fallback to gate — and the hand-rolled
copy here has already drifted from its wording ("KiCAD rejected the move over
IPC" vs the helper's "the {what}"). Two more copies of that paragraph is how it
goes stale.

To be fair: `place_component` on `main` hand-rolls it too. That's ours, not
yours, and I'll migrate it in the same commit so all three closed-board tools sit
on one gate.

## One thing I'd push back on

> Environment-dependent real-KiCad checks were not required for the fallback
> itself because the new path is deliberately gated on IPC being unreachable.

That's *when* the path runs, not *whether its output is correct*. It writes a
`.kicad_pcb` that KiCad has to load afterwards, and a rotated footprint whose
child pad and text angles don't match what KiCad computes shows up as
`lib_footprint_mismatch` in DRC. You did exactly this check on #230 and reported
it — same standard here.

I'll do it as a differential: rotate with Konnect on a closed board, rotate the
same footprint to the same angle in the pcbnew GUI, and diff the two footprint
blocks. That answers "does our transform match KiCad's" directly, rather than
only "does KiCad accept ours".

Sequencing: two crash-class bugs in shipped v0.6.0 go first (#237, #238), then
this, then #230 rebased on top.

mixelpixx:
Rebased onto current `main` (v0.6.1) and pushed the fixes to this branch — your commit is unchanged at the base, mine sit on top.

## What I changed

**The rotate splice ordering** (`7d9f19f`). Root `(at …)` is now replaced first, while its offsets still index the string they were measured against; children rotate afterwards. The two passes touch disjoint blocks, so neither can disturb the other. Both reproductions from my earlier comment are pinned as tests — the shrinking case (`-45` → `0`, four bytes lost, splice lands mid-block) and the growing one.

**Typed classification** (`7d9f19f`). `ClosedBoardError` replaces the `contains("not found on the board")` match. A missing or duplicated reference is `invalid_argument` naming the `reference` field; a board Konnect will not edit says so and states the file was untouched. Nothing bubbles as `anyhow` any more.

**`attempt_ipc_write`** (`7d9f19f`). Both handlers now use the helper instead of re-implementing it. I left `place_component` alone — it hand-rolls the same thing on `main`, which is ours to clean up, not something to expand your diff with.

## What the verification found — one extra fix

I checked the transform by letting KiCad adjudicate rather than reasoning about it: rotate `R1` on the ecc83 demo to a deliberately awkward **247.5°** through the closed-board path, open the result in pcbnew, have KiCad save it, then compare.

First run **disagreed**:

```
Konnect wrote : root [136.271, 107.95, 247.5]
KiCad re-saved: root (136.271, 107.95, -112.5)
```

Same angle — KiCad normalises a footprint's orientation to (-180, 180] — but it means a footprint nobody touched shows a diff after the user's next save, and it makes this path disagree with the IPC path, which comes back normalised because KiCad does it there.

So rotation now writes KiCad's spelling (`ccc312c`). Children are left alone, because KiCad left **both pad angles at 247.5 exactly as written** — it normalises the root and not the children, which I would not have guessed. A move preserves the existing spelling untouched; renormalising an angle the caller did not ask to change is not a move.

Second run:

```
Konnect wrote : root [136.271, 107.95, -112.5]
KiCad re-saved: root (136.271, 107.95, -112.5)
2 pads compared: KiCad agrees with Konnect's transform
```

DRC against a baseline of the unrotated board: **`lib_footprint_issues` 15 → 15, `lib_footprint_mismatch` 0**. The 11 new violations are all physical — clearance, courtyard overlap, shorting, dangling track — because I spun a resistor 337° into its neighbours on a routed board. That is the fixture, not the transform.

## Coverage

Seven guards neutered one at a time. Three were **not caught**, which is why there are two test commits on top:

- a duplicate reference no longer refusing — the PR describes this behaviour and nothing exercised it for move or rotate;
- an unparseable board no longer refusing — same;
- the missing-reference refusal losing its structured field — the existing test only checked that the reference appeared somewhere in the error text, so swapping `field: "reference"` for `field: "board"` left the suite green.

All three now have tests. The seventh, the post-edit board-form re-check, has **no reachable trigger** — `update_footprint_placement` already refuses a non-`footprint` root and `apply_edits` swaps one balanced block for another — so I recorded that in a comment rather than pretending a test covers it.

## Gate

`cargo fmt --all -- --check`, `cargo clippy --workspace --locked --all-targets -- -D warnings`, `cargo test --workspace --locked --lib --tests` (**855**), doctests. Counts unchanged — no tool added.

## One thing for the release notes

`source` on `move_component` and `rotate_component` is a response-shape change, so whatever carries this is a minor.

Thanks for the feature — the closed-board asymmetry was worth closing, and the gate on typed `IpcFailure::Unreachable` was the right call from the start.

PR #230, opened 2026-08-16
feat(pcb): add safe closed-board footprint flipping

### LINKED ISSUE #115: place_component refuses B.Cu (back-side) placement outright — no flip/mirror tool exists to work around it
## Summary

`place_component` rejects any request with `layer: "B.Cu"` outright:

```
{"error":{"field":"layer","kind":"invalid_argument","reason":"back-side placement on 'B.Cu' is not yet supported"},
 "message":"Cannot place on 'B.Cu': back-side placement is not yet supported, because a correct flip must mirror the footprint geometry rather than just rename its layers. Place the footprint on F.Cu and flip it to the back in KiCAD (select it and press F)."}
```

That message is at least honest about the limitation (better than the silent no-ops in #61/#99), but there's no way to complete the workaround it suggests *through Konnect* — there is no `flip_component`/`mirror_component` tool, and none of `edit_component`, `move_component`, or `rotate_component` accept a layer or mirror parameter. The only way to get a footprint onto the back side is to place it on `F.Cu` via Konnect, then switch to the KiCad GUI, select it, and press <kbd>F</kbd> manually.

For any board with mixed front/back placement — which is common (e.g. a through-board button, EMI shielding component, or anything replacing an existing back-side part) — this makes full front-to-back automation impossible without a manual GUI step in the middle.

## Environment

- Konnect v0.2.1 (installed via KiCad PCM)
- KiCad 10.0.5, macOS
- Live IPC session, PCB Editor open with the board loaded

## Repro

1. Board has an existing `B.Cu` footprint (e.g. a button/switch replacing a back-mounted part that was just removed via `delete_component`).
2. Call:
   ```
   place_component(board="...", footprint="Button_Switch_SMD:SW_SPST_EVQP0", reference="SW1", x=124.25, y=89.925, rotation=-90, layer="B.Cu")
   ```
3. Immediate `invalid_argument` error as above — no attempt is made, nothing is placed on either side.
4. Placing the identical call with `layer="F.Cu"` succeeds normally and is confirmed via `get_component_list`.
5. No available tool (`edit_component`, `move_component`, `rotate_component`, or anything else in `pcb_components`) can subsequently move that same footprint to `B.Cu` — a human has to open the GUI, select it, and press F.

## Expected

Either:
- `place_component` accepts `layer: "B.Cu"` and performs a correct mirrored placement (flipping the footprint geometry, not just tagging it with the back layer name), or
- A dedicated `flip_component`/`mirror_component` tool exists so the documented GUI workaround ("place on F.Cu, then flip") can be completed entirely through Konnect, without requiring a manual step in the middle of an otherwise-automated session.

## Actual

Neither exists. `place_component` hard-refuses `B.Cu`, and there's no tool to flip a footprint after placement — the front-to-back workflow is a dead end without manual GUI intervention.

## Notes

Given the reason cited in the existing error message ("a correct flip must mirror the footprint geometry rather than just rename its layers"), this looks like a conscious decision to refuse rather than do it wrong — appreciated over a silent no-op, but it leaves no supported path to finish the job via MCP. A `mirror_component(board, reference)` tool (toggle F.Cu↔B.Cu with correct geometry mirroring, independent of `place_component`) would close the gap and is probably the smaller change of the two options above, since it only needs to get the mirroring math right once and can then be reused by a fixed `place_component` internally too.

## Affected versions

Confirmed on Konnect v0.2.1 / KiCAD 10.0.5.

TWT233:
## Summary

**Implements the safe closed-board `flip_component` path requested by #115.** This is a stacked draft on #229; until #229 merges, review commit `3003e48` for the flip-specific change. After #229 lands, this branch will be rebased onto `main` so the PR contains only the flip commit.

`flip_component(board, reference, layer)` moves an already placed footprint between `F.Cu` and `B.Cu` with KiCad-equivalent geometry mirroring. It is idempotent when the footprint is already on the requested side.

Part of #115. This draft intentionally does **not** close #115 yet: it delivers the revision-aware closed-board path, while live-IPC mirrored-instance construction remains a follow-up unless the maintainer asks to include it here.

## Why `flip_component`

I recommend the public name `flip_component` rather than `mirror_component`:

- **It matches KiCad's own domain language.** The GUI action is “Flip” and the implementation this mirrors is `FOOTPRINT::Flip` / `PAD::Flip`.
- **It groups naturally with the existing PCB operations** `place_component`, `move_component`, and `rotate_component`.
- **It describes the complete board-side operation.** This tool changes the footprint side, root orientation, child geometry, side-specific layers, and text mirroring together.
- **`mirror_component` is more ambiguous.** It can imply choosing an arbitrary X/Y mirror axis, or mirroring geometry without changing F.Cu/B.Cu, neither of which is this contract.
- **KiCad explicitly does not mirror placed footprints.** PCB Editor offers separate horizontal/vertical Mirror actions for shapes, text, zones, tracks, pads, and groups, but skips `PCB_FOOTPRINT_T` and tells users to use Flip to change board side.
- **The operation takes a target side rather than toggling blindly.** `layer: "F.Cu" | "B.Cu"` makes retries idempotent while retaining KiCad's established “flip” term.

This naming decision is explicitly called out for the #115 review. If the maintainer prefers `mirror_component`, the draft can be renamed before merge without shipping two aliases.

## Approach

The closed-board transform follows KiCad's current footprint library-frame behavior:

- mirror child Y coordinates in the footprint-local frame;
- negate and normalize the footprint root orientation;
- swap front/back copper, paste, mask, silkscreen, courtyard, fab, and adhesive layers;
- transform footprint text as `180 - local_angle` and toggle mirroring only on side-specific layers;
- negate pad-local angles;
- mirror primitive points and reverse mirrored arc start/end order;
- preserve 3D model transforms, metadata, groups, and unrelated content.

The handler uses the typed IPC classification established by #229. KiCad 10.0.5 and Konnect's currently vendored protocol do not expose a native footprint-flip command, so a reachable KiCad fails closed with zero file writes and only `IpcFailure::Unreachable` may use the closed-board path. KiCad upstream added `FlipItems` to the 10.0 branch on 2026-08-06; once Konnect raises its supported API baseline and vendors that command, a live-IPC path can reuse this tool contract instead of introducing a second name.

The file path reads one consistent board revision, finds exactly one footprint by reference, validates the complete resulting `kicad_pcb` form, and persists through `write_atomic_if_unchanged`.

## Compatibility and safety

This adds one public tool and updates the catalogue from 202 to 203 registered tools (209 including meta-tools).

Supported closed-board content includes standard SMD/THT pads, properties and legacy `fp_text` fields, line/rect/circle/arc/poly graphics, models, groups, and metadata. Complex padstack geometry that Konnect cannot mirror completely is rejected before writing: nested offsets, trapezoid deltas, chamfers, custom pads/primitives, footprint zones, and unknown `fp_*` items fail closed.

Safety properties covered by tests:

- reachable KiCad never falls back to file editing;
- missing or duplicate references leave the board unchanged;
- malformed/unsupported geometry leaves the board unchanged;
- stale revisions conflict instead of overwriting newer content;
- CRLF is preserved;
- F→B→F semantic round trips restore supported footprints, including non-cardinal angles, THT layer sets, hidden fields, and legacy references.

Rollback is a normal revert of `3003e48`; #229's move/rotate behavior remains independent.

## Validation

- [x] `python3 -m unittest discover -s plugin/tests -v`
- [x] `python packaging/validate-pcm.py --metadata packaging/metadata.json` (isolated venv with CI's `jsonschema` dependency)
- [x] `cargo build --release --locked -p konnect`
- [x] `pcb_components` focused suite: 59 passed
- [x] Tool count and asset-reference gates: 4 + 4 passed

### Real KiCad 10 coupon

Using the built MCP server (no direct KiCad-file edits), a temporary project registered the LH60 libraries, placed three representative footprints on F.Cu, then called `flip_component(..., layer="B.Cu")`:

- 23-pad RP2040 SMD module at a non-cardinal 37.5° rotation;
- SOD-323 diode;
- SMD test point.

All three returned `source: "file"`, `changed: true`, and `layer: "B.Cu"`. Read-only inspection confirmed B.Cu/B.Paste/B.Mask padsets where applicable, B-side graphics, mirrored text, preserved models and attributes.

KiCad 10 DRC reported **zero flip-specific findings**: no `lib_footprint_mismatch`, invalid-layer, or malformed-footprint errors. The six remaining coupon findings were expected and unrelated: one `invalid_outline` (the coupon intentionally has no board outline) and five `silk_over_copper` findings.

## Review checklist

- [x] The diff is focused on #115's footprint-side workflow and contains no local paths, generated output, or project-specific geometry.
- [x] The recommended public name and alternative are explicitly documented for maintainer/review-agent attention.
- [x] File mutations are revision-aware and atomic.
- [x] Reachable-IPC behavior fails closed and cannot race a live editor.
- [x] Registry, `tool-directory.md`, DEV.md, README, bundled skill, PCM metadata, plugin manifest, and troubleshooting counts moved together.

## Stack

- Depends on #229 / #228 for the shared closed-board placement-update seam.
- Flip-specific commit: `3003e485797eda8c973652ec8ce17f1184f23dcc`.
- Issue context and design discussion: #115.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -88,7 +88,7 @@ Konnect/
 │   │           ├── pcb_sync.rs       # update_pcb_from_schematic: pure planner + one-commit IPC apply
 │   │           ├── sch_hierarchy.rs  # 12 tools (typed Sheet model, sheet CRUD + hierarchy/page queries + pin lifecycle)
 │   │           ├── pcb_board.rs      # 11 tools (S-expr file editing, IPC fallback, SVG logo import)
-│   │           ├── pcb_components.rs # 15 tools (IPC real-time + safe headless single-placement fallback)
+│   │           ├── pcb_components.rs # 16 tools (IPC real-time + safe headless single-placement fallback)
 │   │           ├── pcb_routing.rs    # 12 tools (traces, vias, nets, netclasses)
 │   │           ├── pcb_export.rs     # 13 tools (Gerber, PDF, 3D, DRC, DXF/GenCAD/IPC-2581/ODB++)
 │   │           ├── library.rs        # 17 tools (symbol/footprint library management)
@@ -290,7 +290,7 @@ Source: [`crates/konnect-core/src/observability.rs`](crates/konnect-core/src/obs
 
 ## Tool Routing (Starter Kit + On-Demand Loading)
 
-The server does NOT expose all 202 tools (208 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
+The server does NOT expose all 203 tools (209 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
 
 - **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 6 meta-tools, baseline `tools/list` is 20 tools ≈ 2K tokens.
 - **On demand**: the LLM reads `list_toolboxes` → calls `load_toolset(name)` to expose a toolset's tools in subsequent `tools/list` responses. `unload_toolset(name)` prunes them when the task shifts.
@@ -365,9 +365,9 @@ convention for other `kicad-cli`-calling code.
 
 ## Current Stats
 
-- **19 toolsets, 202 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
+- **19 toolsets, 203 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
 - Baseline `tools/list`: 20 tools / ~2K tokens (starter kit + meta-tools)
-- Full-catalog `tools/list` (all loaded): 208 tools (202 registered + 6 meta) / ~25K tokens
+- Full-catalog `tools/list` (all loaded): 209 tools (203 registered + 6 meta) / ~25K tokens
 - **0 IPC stubs** (all protobuf methods implemented)
 - **0 unimplemented tools**
 - **3 CLI commands removed in KiCAD v10** (specctra DSN/SES, pcb sync — return clear errors)
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 Rust binary — that lets Claude and other AI assistants design schematics and PCBs
 through the [Model Context Protocol](https://modelcontextprotocol.io) (MCP).
 
-**202 tools across 19 on-demand toolsets.** Schematic capture, PCB layout and
+**203 tools across 19 on-demand toolsets.** Schematic capture, PCB layout and
 routing, ERC/DRC, design-review audits, JLCPCB part search, Freerouting, reference
 circuits, and a full manufacturing export pipeline — with bundled skills and agents
 that teach Claude KiCAD conventions out of the box.
@@ -70,7 +70,7 @@ through its own S-expression engine with atomic writes (write, fsync, rename), U
 preservation, and round-trip tests — no third-party schematic library with known
 gaps, no text-manipulation workarounds.
 
-**Context economy is a feature.** Exposing all 202 tools to an LLM costs roughly 23K
+**Context economy is a feature.** Exposing all 203 tools to an LLM costs roughly 23K
 tokens of context on every listing. Konnect's router loads a starter kit (~2K
 tokens) and lets the model pull in toolsets on demand — plus built-in observability
 (`get_recent_calls`, `server_stats`, JSONL call logs) so the model can diagnose its
@@ -264,7 +264,9 @@ the main workspace — see [DEV.md](DEV.md) for build steps.
 - `kicad-cli` (ships with KiCAD — used for exports, ERC, DRC)
 - For most PCB tools: KiCAD running with the target board open (IPC API).
   `place_component`, `move_component`, and `rotate_component` can safely fall
-  back to a closed board file when IPC is unreachable.
+  back to a closed board file when IPC is unreachable. `flip_component`
+  intentionally requires a closed board because KiCAD IPC has no native
+  footprint-flip command.
 
 ## License: free for the little guys
 
@@ -304,7 +306,8 @@ with that board first; most PCB tools talk to the running PCB editor. Only an
 unreachable KiCAD produces that message: if KiCAD is running and the tool
 refuses anyway, the error is the tool's own reason for refusing.
 `place_component`, `move_component`, and `rotate_component` can fall back to a
-closed board file when no KiCAD process is reachable.
+closed board file when no KiCAD process is reachable; `flip_component`
+requires one, and refuses while KiCAD holds that board open.
 
 See [docs/TROUBLESHOOTING.md](docs/TROUBLESHOOTING.md) for socket setup, tools
 that don't appear after `load_toolset`, and transaction recovery.
--- a/crates/konnect-core/src/router/registry.rs
+++ b/crates/konnect-core/src/router/registry.rs
@@ -74,9 +74,9 @@ pub static ALL_TOOLSETS: &[ToolsetMeta] = &[
     },
     ToolsetMeta {
         name: "pcb_components",
-        description: "Place, move, rotate, align and duplicate PCB footprints; inspect pads; inspect and edit a placed footprint's graphics",
+        description: "Place, move, rotate, flip, align and duplicate PCB footprints; inspect pads; inspect and edit a placed footprint's graphics",
         category: "pcb",
-        tool_count: 15,
+        tool_count: 16,
     },
     ToolsetMeta {
         name: "pcb_routing",
--- a/crates/konnect-core/src/tools/pcb_components.rs
+++ b/crates/konnect-core/src/tools/pcb_components.rs
@@ -1135,6 +1135,14 @@ fn direct_children_with_tags(
         .collect()
 }
 
+/// Whether `source` has at least one direct `(child_tag …)`, without caring
+/// how many or failing on a block that has none.
+fn has_direct_child(source: &str, parent_tag: &str, child_tag: &str) -> bool {
+    direct_children_with_tags(source, parent_tag)
+        .map(|children| children.iter().any(|(_, _, tag)| tag == child_tag))
+        .unwrap_or(false)
+}
+
 fn exactly_one_direct_child(
     source: &str,
     parent_tag: &str,
@@ -1515,6 +1523,20 @@ fn flip_footprint_block(footprint: &str) -> anyhow::Result<String> {
     for (start, end, tag) in direct_children_with_tags(footprint, "footprint")? {
         let block = &footprint[start..end];
         let replacement = match tag.as_str() {
+            // A property with no `(at …)` carries no geometry, so there is
+            // nothing to mirror and it passes through untouched.
+            //
+            // This is not an edge case. KiCad writes
+            // `(property ki_fp_filters "R_* Resistor_*")` — a bare token, no
+            // position, no layer — into every footprint it places from a
+            // library: 779 of them across the 19 boards shipped in
+            // `share/kicad/demos`. Requiring exactly one `(at …)` on every
+            // property therefore refused practically every real board with
+            // "property must contain exactly one direct (at ...) block",
+            // which is what the first live run of this tool hit. The
+            // synthetic fixture has only positioned properties, so nothing
+            // offline could have caught it.
+            "property" if !has_direct_child(block, "property", "at") => None,
             "property" | "fp_text" => Some(flip_text_block(block, &tag)?),
             "fp_line" | "fp_rect" | "fp_circle" | "fp_arc" => {
                 Some(flip_graphic_block(block, &tag)?)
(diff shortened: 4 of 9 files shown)


mixelpixx:
Reviewed. Answering the two questions you're holding on first, then the findings.

## The name: `flip_component`. Confirmed.

Your reasoning is the right reasoning, and the decisive part is that KiCad itself
draws this distinction: PCB Editor's Mirror actions skip `PCB_FOOTPRINT_T` and
tell the user to use Flip. Naming ours `mirror_component` would invent a
disagreement with the tool we're driving. It also groups with `place_component` /
`move_component` / `rotate_component`, which is what an LLM reading the toolset
listing needs.

Taking a target side rather than toggling is right too — retries stay idempotent.

## Closed-board-only first stage: accepted.

Your API-timeline correction settles it. `FlipItems` is not in KiCad 10.0.5 and
not in our vendored proto, so a live path today would have to synthesise a
mirrored instance and hope — which is exactly the class of bug that put
`place_component`'s `B.Cu` refusal in #115 in the first place. Ship the honest
closed-board path now; when we raise the API baseline and vendor `350f27419af5`,
the live path reuses this tool's contract instead of adding a second name.

So #115 stays open after this lands, tracking the live half. That's correct
scoping, not a shortfall.

## Verification I ran

Merged your branch with current `main` (v0.6.0) and ran the full gate. Conflicts
are bookkeeping only — tool counts and doc prose — with **zero code conflicts**.
Green: fmt, `clippy --workspace --locked --all-targets -- -D warnings`, 856 tests.

The mirroring itself is careful work. Batching every `SexpEdit` against the
original string is the right shape (and #229 does not do this — see below).
Explicit refusal on custom pads, nested offsets, `rect_delta`, chamfers,
primitives, unknown `fp_*` and footprint zones is the right default, and the
F→B→F semantic round trips at non-cardinal angles are the tests that matter.

## Findings

**1. The IPC gate refuses whenever *any* KiCAD is running, not just one holding
this board.** The handler runs `ensure_board_is_active` and then unconditionally
`bail!`s. `IpcFailure::from_error` classifies a plain `anyhow` bail as
`Rejected`, so a KiCAD with an unrelated project open blocks the flip — even
though that board file is demonstrably free and the edit is safe.

`pcb_board::refuse_if_board_open_in_kicad` is the house helper for a tool with no
IPC path, and its final arm is precisely the semantics you want:

```rust
Ok(())  => Some(refusal),   // KiCAD holds THIS board
Err(_)  => None,            // a different board, or unreachable — proceed
```

Call sites to follow: `pcb_board.rs:1065` and `pcb_routing.rs:553`.

This changes one test's expectation: `reachable_rejection_prevents_…` uses
`spawn_rejecting_kicad`, which refuses everything — under the correct gate that
becomes `Err(_)` and *proceeds*. It needs a mock that answers `GetOpenDocuments`
naming this board instead. Your `spawn_planner_mock` in #232 is the model.

**2. The refusal is attributed to KiCAD, which didn't say it.** It renders as
*"KiCAD rejected the footprint flip over IPC: live KiCAD IPC does not expose a
native footprint flip command…"*. KiCAD rejected nothing; Konnect did. This is
the class we fixed in v0.5.0, where `ipc!` announced "KiCAD must be running" for
every failure including refusals from a KiCAD that was running. Fixed for free by
finding 1.

**3. Dead `Ok` arm with a wrong response shape.** `attempt` is
`Result<(), IpcFailure>`, so `Ok(changed)` binds `()` and
`json!({"changed": changed})` serialises to `null`, next to `"source": "ipc"` on
a path that has no IPC implementation. Unreachable today, but it's a trap for
whoever adds the live path later.

**4. 3D model transforms — I want to measure this rather than argue it.**
`flip_footprint_block` matches `property|fp_text|fp_line|fp_rect|fp_circle|
fp_arc|fp_poly|pad` and falls through `_ => None` for `(model …)`. My reading of
`FOOTPRINT::Flip` is that KiCad negates the model's Y offset and its X/Y
rotation — which would make the pass-through a silent divergence for any model
with a non-zero offset, and your fixture carries
`(offset (xyz 1 2 3)) (rotate (xyz 4 5 6))` and asserts it survives untouched.

But that's from reading source, not from a run, so I'm not asserting it. I'll
settle it the way we settled `(pin_numbers hide)` in Round 8: place a footprint
with a non-zero model offset, flip it in the pcbnew GUI, and read what KiCad
actually wrote. If KiCad transforms it, we implement the transform; if not, the
pass-through stays and the claim comes out of the PR body. Either way it
shouldn't be listed under safety until it's measured.

**5. Inherited from #229**: `error.to_string().contains("not found on the board")`
to decide between a structured refusal and a bubbled `anyhow`. Fixing that in
#229 and here.

## What happens next

Chris asked me to take these myself rather than round-trip them, so I'll push the
fixes to `TWT233:feat/closed-board-footprint-flip` with your authorship intact,
as I did for #160. Order:

1. Two crash-class bugs in the shipped v0.6.0 land first — #237 (a `Dwgs.User`
   graphic can terminate a running KiCAD) and #238.
2. #229 with its splice fix.
3. This, rebased onto that, with findings 1–4 and a differential live check:
   flip with Konnect, flip the same footprint in the pcbnew GUI, diff the two
   footprint blocks. That's a stronger witness than the DRC coupon alone, which
   proves KiCad *accepts* the result rather than that it *matches* KiCad's own.

Your coupon run was the right instinct and I'm building on it, not replacing it.

One heads-up on #232: it's the strongest of your three and I'm taking it right
after these, but its `layer_from_name` rewrite is also the fix for #237, and a
crash that can take a user's unsaved board can't wait behind a 3139-line feature.
So I'm extracting that mapping into a focused fix first — with your arithmetic,
credited — and #232 will need it reconciled. That's on me, not you.

mixelpixx:
Correcting the scale of finding 4 before you act on it — I measured the population and it is far smaller than I implied.

I said the model pass-through was "a silent divergence for any model with a non-zero offset". Scanning every `.kicad_mod` in KiCad 10's standard footprint libraries — **14,818 with a `(model …)` block**:

| field | non-zero | flip would negate it |
|---|---:|---|
| `offset.x` | 7 | no |
| `offset.y` | **3** | yes |
| `offset.z` | 0 | no |
| `rotate.x` | **0** | yes |
| `rotate.y` | **0** | yes |
| `rotate.z` | 84 | no |

So the fields a flip would touch are non-zero in **three footprints out of 14,818**, all via `offset.y`, and the 84 `rotate.z` cases — the ones I would have guessed were the problem — are untouched by a flip either way.

The three:

```
RaspberryPi_Pico_Common_THT                 offset (8.89, -24.13, 0)
Bosch_LGA-8_2.5x2.5mm_P0.65mm_Clockwise…    offset (0.015, -0.035, 0)
Relay_SPST_Finder_32.21-x300                offset (-0.0208, -0.0044, 0)
```

Two are sub-0.04 mm and invisible. The Pico is **-24.13 mm**, so a flip that ignores it puts the 3D model roughly 48 mm from where it belongs — rare, but not subtle when it happens.

**And I still have not verified that KiCad transforms it at all.** I tried to measure it the way I measured the rotation: KiCad's demo boards contain exactly one back-side footprint whose library original I can compare against, and its offset is `(0, 0, 0)`, so it proves nothing — negating zero is zero. Settling it properly needs a footprint with a real offset flipped in the pcbnew GUI, since KiCad 10.0.5 has no `FlipItems` to drive it over IPC.

Given that — a claim I cannot yet verify, over a population of three — the proportionate answer is your own fail-closed policy rather than my guessed transform: **refuse a footprint whose model has a non-zero `offset.y`, `rotate.x` or `rotate.y`**, with a message naming the field. It is three lines, it is right whichever way KiCad behaves, and it fires for essentially nobody. If someone hits it, that is the signal to go measure properly.

What should not stay is the current combination: silently preserving the value *and* listing it under safety properties. Whichever way we go, the PR body should say the model transform is out of scope rather than claim it as preserved-by-design.

Sorry for the overstatement in the first pass — "any model with a non-zero offset" was doing a lot of work that three footprints do not support.

mixelpixx:
Rebased onto post-#229 `main` and pushed the fixes here — your flip commit is unchanged at the base, mine sit on top. #229 is merged (`f0fa08c`), so this PR is now just the flip.

## The one that only a real board could find

The first live run against the **stock ecc83 demo** refused every flip:

```
Refusing to edit the board: property must contain exactly one direct (at ...) block.
```

KiCad writes `(property ki_fp_filters "R_* Resistor_*")` — a bare token, no position, no layer — into **every footprint it places from a library**. There are **779** of them across the 19 boards in `share/kicad/demos`. Requiring exactly one `(at …)` on every property refused practically every real board.

Nothing offline could have caught it. All 67 tests passed, because `FLIP_FOOTPRINT` has only positioned properties. A property with no `(at …)` carries no geometry, so it now passes through untouched.

This is the strongest argument for the live step you were already doing on the coupon — your three LH60 footprints were placed by Konnect, so they never had a `ki_fp_filters` to trip over.

## The review findings

**The gate** now uses `refuse_if_board_open_in_kicad` (`aa3598d`). The previous shape ran `ensure_board_is_active` then bailed unconditionally, which `IpcFailure::from_error` classifies as `Rejected` — so *any* reachable KiCAD refused, including one holding an unrelated project. The dead `Ok` arm that would have emitted `"changed": null, "source": "ipc"` went with it, and the closed-board flip now shares #229's `ClosedBoardError`, so a missing or duplicate reference is `invalid_argument` naming the field.

One test changed meaning as a result: `reachable_rejection_prevents_flip_file_fallback` used the rejecting mock, which under the correct gate *proceeds*. It now asserts that — a KiCAD holding a different board must not block a safe edit.

**3D models** — I measured the population before deciding, and it is three footprints, not "any with a non-zero offset" as I first said. `offset.y` is non-zero in 3 of 14,818 library footprints and `rotate.x`/`rotate.y` in none. Since I still cannot verify what KiCad's flip does to them, the guard **refuses** a model a flip would have to move, naming the field — your own fail-closed policy rather than my guessed transform. The fixture's `(offset (xyz 1 2 3)) (rotate (xyz 4 5 6))` is not a shape any standard footprint has, so it now carries the common one (`rotate.z` only, which a flip does not touch).

## Verification

Same differential as #229: flip with Konnect on a closed board, open in pcbnew, let KiCad save, compare.

```
--- R1 ---   layer agrees (B.Cu), root agrees, pads agree, texts agree
--- C1 ---   layer agrees (B.Cu), root agrees, pads agree, texts agree
```

And the graphics, as a set rather than pairwise — **R1's 16 and C1's 197 are each exactly the Y-mirror of the original on the swapped layer**. (My first pass compared sorted lists index-by-index and reported dozens of false mismatches; the ordering changes when the coordinates mirror. The set comparison is the honest one.)

DRC against a baseline of the untouched board:

```
lib_footprint_mismatch      0  ->   0
lib_footprint_issues       15  ->  15
silk_overlap                0  ->  85
```

The 85 are worth explaining rather than waving away: they are all R1's silkscreen against C1's. Flipping negates the root orientation, so an axial part's body swings to the other side of its anchor — R1 at −90° became +90° and now points at C1, ~10 mm away on a dense demo board. Physical, not geometric: **zero `lib_footprint_mismatch`** is KiCad's own check that the placed geometry matches the library for that side and orientation, and it stayed at zero.

## Coverage

Seven guards neutered. Five caught; two were not, and both are now dealt with:

- the **neither-side root layer** refusal had no test — added (reachable from any hand-edited board);
- the **board-open gate** cannot be tested with what this crate has. The only IPC mock rejects everything, so it always exercises the *proceed* side. That is equally true for `add_zone` and the copper-pour path, this helper's two other callers, so I filed #241 with the shape of the mock that would close it and left a comment in the code rather than pretending a test covers it.

## Counts and gate

`pcb_components` 15 → 16, catalogue 203 registered / 209 total, all locations moved together. `cargo fmt --all -- --check`, `cargo clippy --workspace --locked --all-targets -- -D warnings`, **874 tests**, doctests.

`flip_component` and the closed-board-only staging are confirmed as discussed on #115, which stays open for the live-IPC half once `FlipItems` is in our supported baseline.

PR #232, opened 2026-08-16
feat(pcb): update placed footprints from libraries

### LINKED ISSUE #231: Add live Update Footprints from Library support
## Problem

Konnect can update schematic-owned PCB state and edit footprint libraries, but it cannot perform KiCad's **Tools → Update Footprints from Library** operation for already placed footprints. Boards therefore keep stale embedded pads, graphics, attributes, and models after a linked library footprint changes.

## Proposed scope

Add `update_footprints_from_library` to `pcb_components` with:

- live-KiCad-IPC-only operation against the explicitly requested active board;
- dry-run by default and an `expected_plan_revision` gate before apply;
- exact reference/library-ID filters with intersection semantics;
- one atomic `UpdateItems` batch in one KiCad undo transaction;
- preservation of KIID, placement, side, rotation, lock state, schematic metadata, instance fields/overrides, and pad nets by logical pad number;
- refresh of losslessly supported library-owned pads, graphics, fields, attributes, metadata, groups, zones, and 3D models;
- fail-closed conflict reporting for unresolved libraries, removed connected/routed pads, unsupported content, or stale board/library state;
- no closed-board text-editing fallback.

The implementation will include planner/merge tests, mock IPC transaction tests, an ignored disposable live-KiCad test, public tool docs/count updates, and a real 76-footprint LH60 acceptance pass after the isolated tests pass.

A detailed implementation handoff already exists on the contributor branch and will be referenced by the PR.

TWT233:
## Summary

Add `update_footprints_from_library` to the `pcb_components` toolset so an agent can refresh already placed footprints after their linked KiCad library definitions change.

The tool supports exact reference and library-ID filters, defaults to a non-mutating dry run, and reports changed domains before applying. It refreshes losslessly supported library-owned pads, graphics, fields, attributes, metadata, and 3D models while preserving board-owned instance state and pad nets.

Closes #231

## Approach

KiCad 10's IPC API does not expose a native "Update Footprints from Library" command, so this implements the operation as a typed planner and merge:

- resolve the explicitly requested live board and selected footprint libraries;
- parse supported `.kicad_mod` content into IPC footprint definitions and fail closed on unsupported/lossy clauses;
- merge library-owned content with preserved KIID, placement, side, rotation, lock state, schematic metadata, instance fields/overrides, and nets by logical pad number;
- transform back-side definitions with KiCad's exchange-footprint semantics;
- canonicalize KiCad protobuf round-trip defaults, ordering, polygon cycles, and equivalent enum values before deciding whether a footprint changed;
- apply all selected updates in one `UpdateItems` batch inside one KiCad undo commit.

A reviewed dry run returns `plan_revision`; apply requires that exact revision. Any selected conflict or stale revision prevents the batch from being sent.

## Compatibility and safety

- Adds one public MCP tool; existing tools and arguments are unchanged.
- Operation is live-KiCad-IPC-only and validates the requested active board. There is no closed-board text-editing fallback.
- `dry_run` defaults to `true`; `dry_run=false` requires `expected_plan_revision` from a current plan.
- Missing libraries, duplicate/missing selected references, unsupported footprint content, removed connected pads, and stale board/library inputs return conflicts without mutation.
- Apply uses one batch and one undo transaction; an IPC update failure drops the commit rather than leaving a partial refresh.
- Layer handling now composes with #237: footprint refresh routes through `build_footprint_item`, so root, pad, and graphic layers use `try_layer_from_name` and an unrepresentable layer is refused before anything reaches KiCad.
- Tool counts and bundled library/PCB workflow guidance are updated to 204 registered tools / 210 including meta-tools.
- Rollback is a normal KiCad undo for an applied operation, or reverting this PR for the public tool addition.

## Current branch status

Rebased-equivalent merge commit `5f330fe` brings the branch onto post-v0.6.1 `main`, including #229, #230, and #237. The PR is `CLEAN` / `MERGEABLE` at head `5f330fe90e6ec513c76dc38044b679bdb8b6ee38`.

## Validation

Local validation on this branch:

- [x] `cargo test -p konnect-core --locked pcb_footprint_update::tests` (21 passed)
- [x] `cargo test --workspace --locked --lib --tests -- --skip update_symbols_from_library_refreshes_a_stale_embedded_copy --skip update_symbols_from_library_refuses_a_moved_pin_unless_allowed --skip update_symbols_from_library_refuses_a_removed_pin`
- [x] `git diff --check`

The three skipped `update_symbols_from_library_*` tests fail identically on clean `upstream/main@3827cdd` on this host, so they are not introduced by this PR.

GitHub CI for head `5f330fe` is green:

- [x] Check & Test: ubuntu, macOS, Windows
- [x] Clippy
- [x] Format
- [x] Nix flake
- [x] Schematic viewer
- [x] Plugin (Python)
- [x] PCM packaging validation

Live/board acceptance from the original feature work:

- [x] Disposable KiCad 10 live test: `cargo test -p konnect --test live_kicad_tools footprint_library_update_apply_then_dry_run_is_noop --locked -- --ignored --nocapture` (1 passed). It created and placed a fixture, changed its library pads/graphics/metadata/model, completed dry-run -> apply -> noop, then ran DRC.
## Review checklist

- [x] New names follow `docs/NAMING_CONVENTIONS.md`; no existing public name changes require migration handling.
- [x] File/library reads fail closed; live PCB mutation preserves unrelated content.
- [x] Registry, `tool-directory.md`, DEV.md, README, plugin, and packaging counts are synchronized.


mixelpixx:
Status at Round 12: your main-merge resolved the layer_from_name reconciliation cleanly — thank you for taking that on before I got to it. The branch is conflict-free against main as of v0.7.0. What remains on my side before merge: the three review nits (the module-wide `#[allow(dead_code)]`, the `tcp://never-dialed` client built to reach a pure builder, and `preserved` defaulting rather than deriving) plus a live run of `update_footprints_from_library` against a KiCAD session — this tool rewrites placed footprints, so it gets the same live gate flip_component got. That's scheduled next round; nothing needed from you.

mixelpixx:
Merged, with the promised fixups landed on this branch. What I did on top of your work:

**The merge with main** was bookkeeping only — both branches had independently taken `pcb_components` 16→17 (yours for this tool, main's for `repair_corrupted_footprints`), so the merged truth is 18 tools, 205 registered / 211 total.

**The three review nits:**
- `build_footprint_item` lost its `&self` — it was already pure, and this branch's planner was constructing a client on `tcp://never-dialed` just to reach it. It's an associated function now; the phantom clients in the tests went with it.
- `preserved` is now derived by comparing the rebuilt instance against the board's, not asserted from an all-true `Default`. Position, rotation, layer, locked, KIID, symbol path, pad nets, and overrides are each compared; a test pins that a genuinely diverged instance reports `false`.
- The module-wide `#[allow(dead_code)]` is gone — nothing in the module was dead, so it cost nothing.

**One real bug the fixture rule caught** (fixtures-must-come-from-KiCad is a house rule for exactly this reason): KiCad 10 stamps `(embedded_fonts no)` and `(duplicate_pad_numbers_are_jumpers no)` into every footprint it saves — all 15,428 official library files carry them — and `validate_supported_children` refused both, so the tool refused **every** real KiCad-10-saved library file. The hand-written fixture couldn't see it because it shares the module's own format assumptions. Fix: the universal defaults are accepted; the semantic-bearing `yes` values still refuse (`remove_unused_layers yes` exists on 40 official-library pads and the typed rebuild hardcodes `UlrKeep`, so those must not be silently rewritten). The new fixture is your own fixture re-serialized by `kicad-cli fp upgrade` — i.e. what KiCad actually writes — and the flipped-flag refusals are pinned.

**Live verification** (KiCAD 10.0.5, ecc83 demo, results read from the file KiCad saved, not the tool's reply):
- Unfiltered dry run against the stock project library → honest `conflict` naming `fp_text` (see below), zero changes staged.
- After removing the fab text from the R_Axial library entry and widening its silkscreen 0.12→0.3: filtered dry run over R1–R4 → `ready`, 4 changes, `preserved` all-true (derived), domains `[pads, graphics]`.
- Stale `expected_plan_revision` → refused without mutating.
- Apply → one commit; `save_project`; the saved file shows all 4 footprints with 6 widened strokes each, zero old widths, both pad nets preserved by name, and the control footprint byte-untouched.
- Post-apply dry run → `noop` (your convergence property, confirmed live).
- DRC delta vs pristine baseline is exactly `+2 silk_overlap` — the physical consequence of the deliberately widened silkscreen. No `lib_footprint_mismatch`. pcbnew survived throughout.

**The known limitation, measured:** 393 of 400 sampled official library footprints carry `fp_text user "${REFERENCE}"` on F.Fab, which the validator (correctly, given the current builder) refuses — so today the tool conflicts on most official-library footprints and is fully usable on libraries without fab text. Refusing is the right call versus silently deleting fab text, but the capability gap deserves its own issue; filing it with these numbers. `IpcGraphicDefinition::Text` already exists, so the path is teaching `parse_library_footprint` the standard fab-text shape and verifying the round-trip against KiCad's own re-save.

Closes #231. Thanks for the careful design — the plan-revision gate and the mock-commit tests made this a pleasure to verify.

PR #236, opened 2026-08-16
test(sch): resolve the stub Device library from the stub project

pauliuszaleckas:
Three `update_symbols_from_library` tests fail on `main` on any machine with
KiCad installed:

```
assertion `left == right` failed:
  {"pins_moved":[],"unchanged":["Device:R"],"updated":[],"updated_count":0}
```

They put their schematic in a second tempdir, outside the stub project
`stub_symbol_dir()` builds. With no project `sym-lib-table` beside it,
`Device:R` resolves through the global table, which points at KiCad's own
`Device` library — so the test edits a stub the tool never reads, and the
embedded copy is correctly reported unchanged.

The fixture's doc comment already asks for this ("The returned tempdir doubles
as the project directory — put the test's schematic in it"); these three did
not. Each now creates its schematic in the stub project dir. Test-only change.

CI has no KiCad installed, which is why these pass there and fail only for a
developer running the gate locally.

### DIFF
--- a/crates/konnect-core/src/tools/sch_components.rs
+++ b/crates/konnect-core/src/tools/sch_components.rs
@@ -2911,9 +2911,11 @@ mod tests {
     /// "doesn't match copy in library".
     #[tokio::test]
     async fn update_symbols_from_library_refreshes_a_stale_embedded_copy() {
+        // In the stub project dir, so its sym-lib-table shadows the global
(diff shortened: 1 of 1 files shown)


anyn99:
reviewed and approved - without this i get these tests failed (having a local KiCAD installation here).
please merge!

anyn99 (approved):

PR #264, opened 2026-08-17
feat(pcb): batch component placement atomically

nordic-style:
## Summary

Add `set_component_placements`, an additive PCB tool that sets the X/Y position and absolute rotation of multiple existing footprints atomically.

No linked issue.

## Approach

- Validate the complete batch, including duplicate and missing references, before any mutation.
- For live KiCad, read one board snapshot, transform each footprint and all of its pads, graphics, and fields with the existing geometry primitives, and publish one `UpdateItems` request inside one KiCad commit.
- For a closed board, transform every requested footprint in memory and perform one revision-checked atomic file write.
- Keep single-item move and rotate behavior unchanged.

## Compatibility and safety

This is an additive MCP tool and schema. The `pcb_components` count and all quoted total/toolset counts are updated.

Both paths are all-or-nothing. Live IPC verifies that the requested board is active and creates one undo step. The file fallback is used only when IPC is unreachable, rejects stale source content, validates every reference before writing, and preserves pads, graphics, fields, line endings, and unrelated board content.

## Validation

- [ ] `cargo test --workspace --locked --lib --tests` — the full local core run reaches the same three `update_symbols_from_library_*` failures reproduced on `origin/main`; focused suites below pass
- [ ] `cargo test --workspace --locked --doc` — not run
- [x] `cargo clippy -p konnect-core --locked --all-targets -- -D warnings`
- [x] `cargo test -p konnect-core --locked --lib pcb_components::tests` — 57 passed
- [x] `cargo test -p konnect-ipc --locked --test footprint_transform_test` — 3 passed on this branch
- [x] `cargo test -p konnect --locked --test doc_tool_counts` — 6 passed
- [x] Real KiCad 10 check: two existing footprints were submitted at their current placements; the response reported `source: ipc`, one undo step, unchanged live pad positions, and the board saved successfully

## Review checklist

- [x] Tool counts and docs are updated.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -92,7 +92,7 @@ Konnect/
 │   │           ├── pcb_sync.rs       # update_pcb_from_schematic: pure planner + one-commit IPC apply
 │   │           ├── sch_hierarchy.rs  # 12 tools (typed Sheet model, sheet CRUD + hierarchy/page queries + pin lifecycle)
 │   │           ├── pcb_board.rs      # 11 tools (S-expr file editing, IPC fallback, SVG logo import)
-│   │           ├── pcb_components.rs # 18 tools (IPC real-time + safe headless single-placement fallback)
+│   │           ├── pcb_components.rs # 19 tools (IPC real-time + safe headless single-placement fallback)
 │   │           ├── pcb_footprint_update.rs # library refresh planner + one-commit IPC apply
 │   │           ├── pcb_routing.rs    # 12 tools (traces, vias, nets, netclasses)
 │   │           ├── pcb_export.rs     # 13 tools (Gerber, PDF, 3D, DRC, DXF/GenCAD/IPC-2581/ODB++)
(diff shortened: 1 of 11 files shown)


mixelpixx:
Rebased onto main (~45 merges of drift), fixed up, live-verified — merging once CI is green. The design held up completely: nothing that landed since supersedes an atomic full-placement batch for existing footprints, and your all-or-nothing construction (validate → one snapshot → one `UpdateItems` inside one commit, bail-drops-the-commit) is exactly the house shape.

**The rebase** (`43bb6b6` base commit): counts regenerate to 206 registered / 212 total, `pcb_components` 19 (of max 20 — you took the second-to-last slot). `types.rs` keeps both sides' additions (#265's `IpcPad`/`IpcTitleBlock` and your `IpcFootprintPlacement`). The one real hand-merge was `spawn_footprints_mock`: #265 had made the footprint mock *stateful* (a later `GetItems` observes what `UpdateItems` wrote — its pad-readback test depends on it) while you made it multi-footprint. It's now a `Vec` whose matching entries are replaced per incoming update, keeping your `BeginCommit`/`EndCommit` arms; both sides' tests pass against it.

**One substantive fixup, on-theme for this repo:** both success paths echoed the request back as `placements`. The IPC path now derives its response from a **post-commit read-back**, and the file path re-parses what it just wrote. The live run made the case better than any argument could: I requested `rotation: 270` for R1 against a live KiCAD 10.0.5 session, and the response reported **−90** — because that's what KiCad actually stored (it normalizes to (−180, 180]), and the read-back caught KiCad's own normalization. An echoed response would have said 270 while the saved file says −90: two answers for one state. A test pins the same case on the file path.

**Live verification** (ecc83 demo, results read from the file KiCad saved):
- Batch move+rotate of R1 and C1 over IPC: `source: "ipc"`, response values match the saved file exactly (positions and normalized rotations).
- A batch naming a nonexistent reference (`ZZ99`) alongside a real one: refused atomically — "KiCad batch failed; changes dropped" — and the real footprint in that batch was untouched in the file.
- One incidental find, not yours: a freshly-launched KiCAD answers `AS_NOT_READY` for a while, and the tool correctly refused without falling back to the file — the #265-era fail-closed gating doing its job.

Full gate green at 1142 tests. Thanks for the careful atomicity work — the byte-identical-after-refusal tests are exactly the property that matters.

Issue #237, opened 2026-08-16
[BUG] Dwgs.User footprint graphics can crash KiCad during live placement


neusse:
# KiCad 10.0.5 IPC footprint-creation crash research

Research date: 2026-08-16  
Observed KiCad build: Windows x64 `10.0.5.50609`  
Scope: official KiCad documentation, source, libraries, issue tracker, and official `kicad-python` tracker; local Konnect source and reproduced tests

## Incident request

Konnect v0.6.0 can terminate the running KiCad 10.0.5 process while placing an
official library footprint through live IPC. The immediate trigger is a valid
footprint graphic on `Dwgs.User`: Konnect maps that layer to `BL_UNDEFINED`, and
KiCad faults while consuming the invalid scalar layer instead of returning an
API error. This can discard unsaved work in the active KiCad session.

### Minimal reproduction

1. Create an empty KiCad project and open its PCB in KiCad 10.0.5 with the API
   enabled.
2. Load Konnect's `pcb_components` toolset.
3. Call `place_component` against the active board with either exact footprint:

   ```json
   {
     "board": "<active-board>.kicad_pcb",
     "footprint": "Connector_USB:USB_C_Receptacle_GCT_USB4105-xx-A_16P_TopMnt_Horizontal",
     "reference": "J1",
     "x": 100,
     "y": 100
   }
   ```

   or `Connector:BJB_Pico_46.110.1001_Receptacle_Horizontal`.
4. KiCad exits during `CreateItems`. Konnect reports an NNG receive timeout;
   Windows records exception `0xc0000005` in `kicommon.dll` at offset `0x87e70`.

The USB-C footprint also reproduces through `update_pcb_from_schematic`; simple
footprints and multiple targeted pad-construction controls succeed.

### Expected

The footprint is created with its `Dwgs.User` graphics intact. If a future layer
cannot be represented, Konnect returns a structured error before sending any
mutation to KiCad. KiCad must not be terminated.

### Actual

Konnect serializes `Dwgs.User` as `BL_UNDEFINED`. KiCad terminates with a native
access violation, the IPC call times out, and unsaved session work is at risk.

### Relationship to #232

#232 is a practical prerequisite or overlapping dependency: it already adds the
missing `Dwgs.User`, `Cmts.User`, `F.Adhes`, and `B.Adhes` mappings. However, it
is a much larger feature PR, is currently conflicting, and does not include the
two exact crash regressions or a fail-closed unknown-layer boundary. This bug
[...]

PR #265, opened 2026-08-17
fix(pcb): harden live footprint pad readback

nordic-style:
## Summary

This is now a stacked follow-up to #207, not a competing implementation.

The branch contains #207's commit unchanged, followed by `edcc9b4`. Review only the second commit. Once #207 lands, the shared base commit will disappear from this PR's diff.

Depends on #207.

## Approach

Keep #207's broader live-board reader and its explicit document targeting, then add only the remaining hardening:

- require exact protobuf `Any` types before decoding footprints and pads;
- report malformed pads and missing positions instead of silently dropping them or fabricating `0,0`;
- return live pad-stack layer names, with the same `layers` field in the saved-file fallback;
- distinguish transport-unreachable from KiCad-rejected IPC failures, so a reachable KiCad can never trigger a stale-file fallback;
- verify that move → pad readback observes the updated live board.

## Compatibility and safety

The existing tools remain source-compatible. Responses gain the additive `layers` field; `source` remains unchanged.

This is read-only. It keeps #207's named-document selection, absent-footprint handling, and pad-less-footprint consistency guard.

## Validation

- [x] `cargo test -p konnect-ipc --locked --test mock_server_test --test footprint_transform_test` — 29 passed, 1 intentionally ignored timeout test
- [x] `cargo test -p konnect-core --locked tools::pcb_components::tests` — 62 passed
- [ ] `cargo test --workspace --locked --lib --tests` — 551 core tests passed; the same three unchanged `update_symbols_from_library_*` / `Device:R` failures reproduced on the local baseline remain
- [x] The retained layer/filtering behavior was previously checked against real KiCad 10 on a 100-contact connector: 100 numbered pads plus 4 real unnamed mechanical pads, no footprint artwork, and the expected B.Cu/B.Mask/B.Paste layers

## Review checklist

- [x] The follow-up diff is limited to behavior not already supplied by #207.
- [x] New failure paths have regression coverage.
- [x] Reachable IPC failures fail closed instead of returning stale file data.
- [x] Tool counts are unchanged.

### DIFF
--- a/crates/konnect-core/src/tools/pcb_components.rs
+++ b/crates/konnect-core/src/tools/pcb_components.rs
@@ -1883,8 +1883,9 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "get_component_pads",
-            "Return the pad positions and net assignments for a footprint. \
-             Reads the board open in KiCad when it is reachable, else the file — \
+            "Return live board-space pad positions, layers and net assignments for a footprint. \
+             Reads the board open in KiCad when it is reachable and falls back to the file only \
+             when KiCad IPC is unreachable — \
              'source' says which, so unsaved placements are visible without a save. \
              A pad's 'net' is its net name, \"\" if the pad carries no net \
              (unconnected), or — reading the file — null if the net node is present \
@@ -1901,7 +1902,7 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "get_pad_position",
-            "Return the schematic-space position of a specific pad number on a footprint.",
+            "Return the live board-space position, layers and net of a specific pad number on a footprint.",
             json!({
                 "type": "object",
                 "properties": {
@@ -2977,7 +2978,7 @@ async fn handle_get_component_pads(
     // live one. The file stays the fallback for an offline session.
     let ipc_board = board_path.clone();
     let ipc_reference = reference.clone();
-    let live = with_ipc(ctx.config.ipc_address.clone(), move |c| {
+    let live = with_ipc_classified(ctx.config.ipc_address.clone(), move |c| {
         let document = c.find_open_board(&ipc_board)?;
         c.get_footprint_pads_in(document, &ipc_reference)
     })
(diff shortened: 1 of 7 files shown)


pauliuszaleckas:
Looks like this duplicates some of my PR #207

nordic-style:
You're right — thanks for pointing me to #207. I missed it when I opened the original version of this PR.

I've now rewritten #265 as a stacked follow-up:

- your `af5eb5a` commit is the unchanged base;
- the independent duplicate implementation is gone;
- the only follow-up commit is `edcc9b4`, which adds exact `Any` type checks and strict decode errors, pad-stack layer names, transport-vs-rejection fallback classification, and a move → live-readback regression test;
- #207's named-document/multi-board handling and pad-less-footprint guard remain intact.

I've also converted #265 to draft. Once #207 lands, its shared commit will disappear from this diff and #265 will contain only the hardening delta. I'll keep it rebased if #207 changes. Thanks again.

mixelpixx:
Holding review on this until #207 (its base) is decided — reviewing it standalone would double-count #207's diff. It stays in the queue; nothing needed from you now.

mixelpixx:
Rebased onto main and marked ready — the stacked #207 commit disappeared from the diff exactly as your description predicted, which made this the easiest review of the round. What remains is your hardening commit plus one fixup of mine.

**Verification (four-step, per house practice):**
- All three new guards neutered individually and caught by their tests: restoring the silent `filter_map(...ok())` skip fails `an_unreadable_live_pad_is_reported_instead_of_silently_dropped`; restoring the fabricated `(0,0)` fails `a_live_pad_without_a_position_is_reported_instead_of_fabricated_at_zero`; restoring the `Rejected → stale file` fallback fails `pad_reads_do_not_fall_back_when_reachable_kicad_rejects`.
- Live against KiCAD 10.0.5 (ecc83 demo): `get_component_pads` with pcbnew holding the board returns `source: "ipc"` with correct positions and net names; with KiCAD closed, `source: "file"` returns identical positions and nets. Both paths carry `layers`.
- One vocabulary note, deliberate and left as-is: the live path reports the pad stack's expanded layer names (a THT pad lists all 34 entries KiCad holds), while the file path reports the file's own `*.Cu` / `*.Mask` wildcards. Each is faithful to its source and `source` tells the caller which vocabulary applies; expanding the file's wildcards would be fabrication in the exact sense this PR removes.

**The fixup commit (`04b14da`):** your `layers` field made a latent gap user-visible. The live run reported a through-hole pad's stack as `F.Cu, F.Mask, B.Cu, B.Mask, In1.Cu, In2.Cu` followed by **28 copies of `"Unknown"`** — `layer_enum_to_name` was still the pre-#237 narrow map; only the forward direction was widened when the crash was fixed. There is now a `builders::layer_name` that is the exact inverse of `layer_from_name` (computed `In1–In30` and `User.1–45` runs included, stepping the `Rescue` hole), pinned by a round-trip test over all 96 representable layers, and the client delegates to it. Tracks and shapes benefit for free.

This continues the pattern your benchmark discussions established: the response derives from the result, and a value the code cannot derive is an error, not a plausible number. Merging once CI is green.

PR #268, opened 2026-08-17
fix(validation): reject malformed nested tool inputs

### LINKED ISSUE #234: Array-item schemas declare required fields and default them: batch_add_wire adds a (0,0)-(0,0) wire
#218 fixed required arguments at the top level of a tool's schema, and added a
dispatch-level check that enforces them before any handler runs. That check is
deliberately shallow: it validates presence of the names in the tool's own
`required` list and does not descend into array items.

Three tools declare `required` fields inside an array's `items` schema and then
default every one of them.

### The one that writes

`batch_add_wire` — items require `["x1","y1","x2","y2"]` (`sch_wiring.rs:45`),
and the loop reads each with `.as_f64().unwrap_or(0.0)`. A wire element missing
its coordinates becomes a wire from **(0,0) to (0,0)**, which is then snapped to
grid, **added to the schematic**, counted in `added_wires`, and run through
T-junction detection. `{"wires": [{}]}` adds a degenerate wire and reports
success.

### The two that build library content

- `create_footprint` — pad items require
  `["number","type","shape","x","y","width","height"]` (`library.rs:58`) and
  default all seven (`library.rs:685-692`): `"1"`, `"smd"`, `"rect"`, `0.0`,
  `0.0`, `1.0`, `1.0`. A `{}` pad becomes a 1x1 mm SMD rect pad numbered 1 at
  the origin. Since #218's fix `pads` itself can no longer be omitted, but its
  *contents* are still unchecked — and a footprint whose pads all landed on the
  origin is worse than one with no pads, because it looks plausible.
- `create_symbol` — unit items require `["pins"]` (`library.rs:182`), defaulted
  to an empty vec, producing a pin-less unit.

Ten other tools with nested `required` enforce it correctly — via
`let Some(..) else`, an explicit `match … None => error`, `read_xy_pair`,
`parse_points`, or `serde_json::from_value` failing on a `Null`. So the
convention exists; these three predate or missed it.

### The decision this needs

Unlike the top-level case there is no single obvious behaviour, which is why
this is not folded into #218:

- **Refuse the whole call** when any element is malformed — consistent with how
  a missing top-level argument now behaves, and keeps a batch atomic.
- **Skip the element and report it** — matches how several batch tools already
  handle a per-element failure (`errors: [...]` alongside a count), and lets a
  20-wire batch with one bad entry still do 19 useful things.

I lean to refusing for `create_footprint`/`create_symbol`, where the output is
a single artefact that is either right or not, and to skip-and-report for
`batch_add_wire`, where the batch shape already implies partial progress and
the tool already returns an `errors` array. But that is a judgement, not an
obvious answer.

Whichever is chosen, the error should name the **index** as well as the field —
"each point needs a numeric 'x'" tells you nothing about a twelve-element
batch. `parse_points` (`pcb_components.rs:113`) is the existing pattern for
that.

nordic-style:
## Summary

Reject malformed nested items instead of silently inventing data:

- `batch_add_wire` no longer turns missing coordinates into a `(0,0)-(0,0)` wire.
- `create_footprint` no longer turns an incomplete pad object into a plausible origin pad.
- `create_symbol` no longer turns an incomplete unit into a pin-less unit, and also validates the required fields of nested pin items.

Closes #234

## Approach

Each handler validates every nested item before it loads, creates, or replaces its target file. Errors are structured `invalid_argument` results whose field includes the precise array index, for example `wires[1].x2` or `units[0].pins[1].name`.

For wire batches this deliberately refuses the entire call rather than applying the valid prefix and reporting a partial result. That keeps the operation atomic and matches the failure behavior of missing top-level required arguments.

## Compatibility and safety

Valid payloads and schemas are unchanged. Payloads that violate the existing nested `required` declarations now fail instead of receiving hidden defaults.

Regression tests verify that a malformed item leaves an existing schematic, footprint, or symbol library byte-identical and does not create a new target file.

Risk is limited to callers that relied on the undocumented fallback values for malformed items. Rollback is a single commit revert; no generated data or migration is involved.

## Validation

- [ ] `cargo test --workspace --locked --lib --tests` (left to CI; local full `konnect-core` run described below)
- [ ] `cargo test --workspace --locked --doc` (left to CI)
- [ ] `cargo clippy --workspace --locked -- -D warnings` (left to CI; package-level clippy passed)
- [x] `cargo test -p konnect-core batch_add_wire`
- [x] `cargo test -p konnect-core create_footprint`
- [x] `cargo test -p konnect-core create_symbol`
- [x] `cargo test -p konnect-core required_argument_tests`
- [x] `cargo clippy -p konnect-core --all-targets -- -D warnings`
- [x] `cargo test -p konnect-core`: 546 passed; the same three environment-dependent `update_symbols_from_library_*` tests already fail on unmodified `main` because the locally installed KiCad library shadows their fixture.

## Review checklist

- [x] New names follow `docs/NAMING_CONVENTIONS.md`; no public names were changed.
- [x] IPC mutations verify the requested board and do not leave partial batches (no IPC mutations in this change).
- [x] If tools were added/removed: counts and docs updated per CONTRIBUTING.md (no tools added or removed).

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -692,16 +692,21 @@ async fn handle_create_footprint(
         Ok(v) => v.clone(),
         Err(e) => return Ok(e),
     };
+    if let Err(error) = validate_footprint_pad_items(&pads_val) {
+        return Ok(error);
+    }
     let mut pad_geoms: Vec<PadGeom> = Vec::new();
     let mut pad_sexp = String::new();
     for pad in &pads_val {
-        let number = pad["number"].as_str().unwrap_or("1").to_string();
-        let pad_type = pad["type"].as_str().unwrap_or("smd").to_string();
-        let shape = pad["shape"].as_str().unwrap_or("rect");
-        let x = pad["x"].as_f64().unwrap_or(0.0);
-        let y = pad["y"].as_f64().unwrap_or(0.0);
-        let w = pad["width"].as_f64().unwrap_or(1.0);
-        let h = pad["height"].as_f64().unwrap_or(1.0);
+        // Safe after validate_footprint_pad_items: every schema-required field
+        // has the declared type, so no plausible-looking pad is invented.
+        let number = pad["number"].as_str().expect("validated").to_string();
+        let pad_type = pad["type"].as_str().expect("validated").to_string();
+        let shape = pad["shape"].as_str().expect("validated");
+        let x = pad["x"].as_f64().expect("validated");
+        let y = pad["y"].as_f64().expect("validated");
+        let w = pad["width"].as_f64().expect("validated");
+        let h = pad["height"].as_f64().expect("validated");
 
         let layer_names = if let Some(layers) = pad["layers"].as_array() {
             let mut names = Vec::with_capacity(layers.len());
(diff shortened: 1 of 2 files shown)


mixelpixx:
#318 just landed the wire-batch half of #234 (same lines of `handle_batch_add_wire`, smaller diff). Please rebase this PR to drop the now-redundant wire hunk and keep only the `create_footprint`/`create_symbol` nested-item validation — that half is still wanted and still yours. Same consolidation move as the #285 series.

neusse:
Maintainer status check: the remaining `create_footprint` / `create_symbol` validation is still needed on current `main`, but this branch has been idle since the August 25 reconstruction request and now conflicts.

I reopened #234 because #318 completed only the `batch_add_wire` criterion. The current implementation still defaults malformed nested pad/unit fields into plausible library content. The issue-to-PR accounting is now:

- #318: completed the atomic `batch_add_wire` validation.
- #268: terminal PR for indexed, atomic nested-item validation in `create_footprint` and `create_symbol`; it may retain `Closes #234` once that remaining acceptance is satisfied.

Please let us know by September 12 whether you intend to continue. If so, reconstruct this PR from current `main` with only its unique library-validation change and ping us. If there is no response, we will close the stale PR without prejudice and leave #234 available for a fresh implementation.

neusse:
Reconstructed on current main as #482. The replacement preserves nordic-style's authorship, drops the `batch_add_wire` hunk already delivered by #318, keeps only the remaining library-validation work, includes the negative control, and passes the complete local gate. Closing this stale branch in favor of the focused replacement.

Issue #251, opened 2026-08-17
Tools accept parameters they never apply and report them as honoured


mixelpixx:
Found while verifying @neusse's discussion #239 findings on the JLCPCB CPL and Gerber outputs. The specific defects are confirmed; so is the general shape.

## The pattern

A tool's schema declares a parameter, the handler never reads it, and the response reports it back as though it had been applied. The caller has no way to tell the difference between "honoured" and "ignored".

## Confirmed instances

**1. `export_position_file` echoes `units` and `side` it never applies.**

`crates/konnect-core/src/tools/pcb_export.rs:491-514` reads both, then:

```rust
cli::export_position_file(cli, &board, &output, format).await?;   // units/side not passed
...
"side": side, "units": units,   // reported as if honoured
```

`cli::export_position_file` (`cli.rs:546-564`) forwards only `--format`. No `--units` flag ever reaches kicad-cli, so the file gets kicad-cli's own default — which is what produced the report's observation that the CPL carries what look like **inch** coordinates with no unit declaration while the response said `mm`. `export_manufacturing_package` (`manufacturing.rs:185-197`) has the same gap and does not even offer the parameters.

**2. `export_gerber`'s `layers` parameter is dead.**

The schema advertises it (`pcb_export.rs:53-57`, "empty = all fabrication layers"), but `handle_export_gerber` (`pcb_export.rs:316-357`) never reads `args["layers"]`, and `cli::export_gerber` (`cli.rs:413-425`) passes no `--layers` at all. The exported layer set is entirely governed by the board file's saved plot settings, which Konnect never inspects or overrides — consistent with the report's finding that the JLCPCB package included Adhesive, Courtyard, Fab, Margin, User_Comments and User_Drawings. A caller cannot restrict it even though the documented parameter says they can.

Also unfiltered: nothing in Konnect honours KiCad's `exclude_from_pos_files` footprint property, so mounting holes appear in the CPL. Whether that is Konnect's job or kicad-cli's is worth deciding explicitly, but right now nobody does it.

## Why this is worth a durable guard, not just three fixes

This is the same defect as #218 (arguments declared required, then defaulted, then reported as supplied) one step further along: **a response field derived from the request instead of the result.** The same shape produced #190 (netclass echoing schema defaults back as if asked for), #211 (register reporting success for a no-op), and the review verdicts filed alongside this.

[...]

PR #269, opened 2026-08-17
fix(manufacturing): apply Gerber and position options

nordic-style:
## Summary

- Apply `units` and `side` to position-file exports instead of only echoing them in the response.
- Apply the requested Gerber `layers` as KiCad's single comma-separated `--layers` value.
- Give manufacturing packages explicit `gerber_layers`, `position_units`, and `position_side` controls.
- When Gerber layers are omitted, derive a fabrication-only set from the real board layer table: enabled copper, masks, silkscreens, and `Edge.Cuts`.
- Document that KiCad itself honors `exclude_from_pos_files`.

Refs #251

## Approach

The CLI wrappers now have pure argument builders with regression tests. The handlers validate the public values, forward them to `kicad-cli`, and only report values that apply to the selected format. Gerber position output therefore omits `units`, because KiCad does not accept `--units` for that format.

The default Gerber selector parses KiCad's real layer table and filters by canonical layer role. It excludes adhesive, paste, courtyard, fab, margin, comments, and user drawings instead of inheriting an opaque saved plot selection.

The cross-cutting schema/handler guard requested by #251 is intentionally a separate follow-up PR: the first implementation found pre-existing ignored parameters in unrelated schematic, routing, review, and integration tools. Keeping that audit separate follows CONTRIBUTING.md's one-reviewable-outcome rule and avoids hiding those compatibility decisions inside a manufacturing fix.

## Compatibility and safety

The existing standalone `format`, `units`, `side`, and `layers` fields keep their names and now affect the generated files as documented. The manufacturing-package fields are additive.

An omitted/empty Gerber layer list now means the documented fabrication-only selection rather than the board's saved plot settings. Callers can preserve any custom set by passing its exact layer names. `format: "gerber"` with `side: "both"` now returns a structured argument error because KiCad supports only one side per Gerber position file.

Risk is limited to export selection and response metadata; source boards are only read. Rollback is a single commit revert. No project migration or generated artifact is committed.

## Validation

- [ ] `cargo test --workspace --locked --lib --tests`: all changed and unrelated suites passed except the same three local `update_symbols_from_library_*` tests that fail on unmodified `main` because the installed KiCad `Device:R` library shadows their fixture (551/554 `konnect-core` tests passed).
- [x] `cargo test -p konnect-core gerber_export_tests`
- [x] `cargo test -p konnect-core position_export_tests`
- [x] `cargo test -p konnect-core fabrication_option_tests`
- [x] `cargo test -p konnect-core package_export_option_tests`
- [x] Real KiCad 10 check on a temporary copy of a populated board: `--units mm` and `--units in` differed by exactly 25.4, `--side front` produced only top-side rows, and the explicit Gerber selection produced only the seven requested layers plus the job file.
- [x] Real KiCad 10 check on the repository's saved-board fixture: all four footprints marked `exclude_from_pos_files` were omitted from the CSV.

## Review checklist

- [x] New public fields follow `docs/NAMING_CONVENTIONS.md`; existing names remain compatible.
- [x] Source board files are read-only; exports go only to the requested output directory.
- [x] No IPC mutations are involved.
- [x] No tools were added or removed, so registry and catalogue counts are unchanged.

### DIFF
--- a/crates/konnect-core/src/tools/cli.rs
+++ b/crates/konnect-core/src/tools/cli.rs
@@ -486,16 +486,32 @@ pub async fn export_netlist(
 
 // ─── PCB Export ──────────────────────────────────────────────────────────────
 
-/// KiCAD 10: `pcb export gerbers --output <dir> <input>` (PLURAL!)
-pub async fn export_gerber(cli: &str, pcb: &Path, output_dir: &Path) -> Result<()> {
-    let args = [
-        "pcb",
-        "export",
-        "gerbers",
-        "--output",
-        output_dir.to_str().unwrap(),
-        pcb.to_str().unwrap(),
-    ];
+/// Argument vector for Gerber export. KiCad's plural `gerbers` subcommand
+/// accepts the complete selection as one comma-separated `--layers` value.
+fn gerber_args<'a>(output_dir: &'a str, pcb: &'a str, layers_csv: &'a str) -> Vec<&'a str> {
+    let mut args = vec!["pcb", "export", "gerbers", "--output", output_dir];
+    if !layers_csv.is_empty() {
+        args.push("--layers");
+        args.push(layers_csv);
+    }
+    args.push(pcb);
+    args
+}
+
+/// KiCad 10: `pcb export gerbers --output <dir> [--layers <csv>] <input>`
+/// (PLURAL!)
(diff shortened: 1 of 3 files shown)


nordic-style:

PR #271, opened 2026-08-18
fix(sch): resolve pins in find_orphan_items

### LINKED ISSUE #249: [BUG] find_orphan_items is pin-blind and contradicts ERC on labels placed on pins
Reported by @neusse in discussions #239 (section 3) and #165. Confirmed.

## Symptom

On a schematic where labels are placed directly on component pins — a legal and common KiCad connection method — `find_orphan_items` reported **all 35 labels as floating**, while ERC, the short detector, wire validation, component-connection validation, the connection audit and the single-pin-net check all passed with zero findings.

The reporter's attempted workaround made things worse: converting them to 2.54 mm labelled stubs caused two automatically oriented vertical stubs to meet at their midpoint and briefly short CC2 to GND. That is a second issue (automatic stub placement does not collision-check), but the trigger was this false positive.

## Root cause

`handle_find_orphan_items`, `crates/konnect-core/src/tools/sch_analysis.rs:583-611`. It builds a set of wire endpoints and a set of label positions, then:

```rust
let floating = labels.iter().filter(|l| !endpoint_counts.contains_key(&pt_key(l.x, l.y)))
```

"Floating" is defined purely as *a label whose coordinate does not coincide with a wire endpoint*. The function never calls `extract_symbol_instances`, `extract_lib_pins` or `pin_endpoint` — it has no notion of a pin position at all. A label sitting on a pin with no wire has no wire endpoint under it, so it is unconditionally reported as floating.

Every other connectivity path in the same file builds a union-find `NetGraph` via `build_net_graph` (`sch_analysis.rs:266-294`) and resolves pins through `pin_endpoint(p, inst.pin_transform())` — see `handle_get_pin_connections` (`:415`), `handle_get_component_nets` (`:462`), `handle_get_connected_items` (`:668`). Those resolve label-on-pin correctly. `run_erc` shells out to kicad-cli, KiCad's own authoritative engine, which also treats it as connected.

**The divergence, precisely:** `find_orphan_items` tests connectivity as *wire-endpoint coordinate membership*; every other validator tests it as *net-graph membership seeded by pins, wires and labels together*. It is the only one that omits pins.

## The tool description also overstates it

`sch_analysis.rs:130` describes the tool as finding "dangling wire ends, floating labels, and **unconnected pin endpoints**". It never inspects pin endpoints — it only mislabels connected ones.

## Why it matters beyond a false positive

An agent needs one reliable definition of connectivity. When two Konnect tools disagree about whether a design is connected, the agent either stops on a design that is fine or, worse, "fixes" it — which is exactly what happened here and what produced a real short.

## Direction

Reuse `build_net_graph` and resolve pins the way the other four callers already do, so orphan detection shares one connectivity model with ERC and the validators. Either implement the "unconnected pin endpoints" the description promises, or correct the description.

nordic-style:
## Summary

- stop treating labels placed directly on component pins as floating
- recognize labels anywhere on a wire and wire ends terminating on component or sheet pins
- report genuinely unconnected symbol pins, while respecting no-connect flags and intrinsic no-connect pins
- resolve placed pins through the existing unit-aware and `lib_name`-aware path
- validate the configurable coincidence tolerance before building spatial indexes

Closes #249

## Approach

`find_orphan_items` now indexes wire segments, wire endpoints, labels, junctions, hierarchical sheet pins, no-connect markers, and every actually placed symbol pin. It evaluates each orphan category against that shared physical connection set instead of equating connectivity with membership in the wire-endpoint map.

Pin extraction remains unit-aware, so an unplaced unit cannot create phantom findings. The shared S-expression helpers now expose no-connect and sheet-pin positions, and the batch connection validator reuses the same no-connect extractor.

This is a focused port of @pauliuszaleckas' original commit `1f11486fc54858d2087db19bcb5b594ffd50c432` from his stacked development branch onto current `main`. His Git authorship and the original co-author trailer are preserved. Thanks also to @anyn99 for the independent KiCad 10.0.4 validation recorded on #249.

## Compatibility and rollback

- The response now includes `tolerance` and may include `unconnected_pin` findings that the tool description already promised but the implementation never produced.
- The optional `tolerance` input defaults to 0.05 mm and rejects non-positive values.
- `LibPin` now retains KiCad's electrical type so intrinsic `no_connect` pins can be excluded correctly.
- The operation remains read-only; no `.kicad_sch` source is modified.
- Rollback is a single revert of commit `d5a8869`.

## Validation

- 9 focused regression tests pass: label on bare pin, label mid-wire, wire end on component pin, genuinely unconnected pin, no-connect flag, intrinsic no-connect pin, multi-unit isolation, hierarchical sheet pin, and invalid tolerance
- `cargo test -p konnect-sexp --locked`: 133 unit tests + 8 property tests pass; 6 doc tests pass and 1 is intentionally ignored
- `cargo test --workspace --locked --lib --tests`: all suites pass except the same three pre-existing local `update_symbols_from_library_*` failures caused by the installed KiCad `Device:R` library shadowing the test fixture; `konnect-core` reaches 551/554 passing tests including all 9 new tests
- independent real-world results already recorded on #249: two ERC-clean KiCad 10.0.4 schematics dropped from 103/14 false dangling ends and 3/48 false floating labels to zero; a deliberate-defect control still reported exactly the expected five findings


pauliuszaleckas:
OK, this works for me too :D

pauliuszaleckas (approved):

anyn99 (approved):
Working with AI is so damn fast, half a day is nothing, phew! You didn't PR this - a few ours later someone else does it and even improved! :-D

Since it's basically pauliuszaleckas' 1f11486 with improvements I would prever this one now ;-)

PR #267, opened 2026-08-17
fix(schematic): respect intentional unconnected pins

nordic-style:
## Summary

Stop `validate_component_connections` from reporting intentionally unconnected symbol pins and pins belonging to other units, and make its existing `ignore_power_pins` option effective.

Refs #182

## Root cause and approach

The validator used the unit-agnostic pin extractor, discarded each library pin's electrical type, and never read `ignore_power_pins`. As a result it superimposed pins from other units and treated declared no-connect and optionally ignored power pins as missing connections.

- Preserve each embedded library pin's KiCad electrical type.
- Exempt pins typed `no_connect` and ordinary pins carrying a schematic no-connect marker.
- Honor `ignore_power_pins` for `power_in` and `power_out` pins.
- Use the existing unit-aware extractor so a placed unit does not inherit another unit's pins.
- Add `pin_type` to each finding and clarify the public tool description.

This keeps the smaller #267 framing selected by the maintainer. #280 is now closed as its duplicate. #272 continues to cover the other analysis/export/review paths.

## Compatibility and safety

The request schema is unchanged. Findings gain an additive `pin_type` field. Results intentionally stop containing library-declared NC pins, ignored power pins, and pins from unplaced units.

The validator is read-only and does not alter the schematic.

## Validation

After rebasing this one-commit PR onto current `main` (`2183267`), all commands required by `CONTRIBUTING.md` pass on this exact branch:

- `cargo test --workspace --locked --lib --tests`
- `cargo test --workspace --locked --doc`
- `cargo clippy --workspace --locked --all-targets -- -D warnings`
- `cargo fmt --all -- --check`

Live-GUI tests remain intentionally ignored by the upstream suite unless a running KiCad GUI/socket is supplied.

## Risk and rollback

The main behavioral risk is filtering a pin that should still be reported. Regression tests cover declared NC pins, schematic no-connect markers, power filtering, selected units, and mid-wire junction behavior. Rollback is a single focused commit.

PR #272, opened 2026-08-18
fix(schematic): resolve pins through their placed unit

nordic-style:
## Summary

- resolve pins through the unit instance that actually places them in schematic validators and exports
- search every placed unit for reference-level pin and connectivity queries
- prevent `fix_connectivity` from snapping wires to phantom pins belonging to another unit
- make every design-review audit unit-aware and retire the temporary partial-coverage verdict
- identify the resolved unit in additive analysis and netlist-summary response fields

Refs #182

## Approach

Every per-instance path now calls `extract_lib_pins_for_unit(symbol, instance.unit)` before applying that instance's transform. Reference-level queries either use the established `resolve_placed_pin` helper or explicitly visit every placement sharing the reference. This preserves common pins while preventing another unit's local coordinates from being projected through the wrong placement.

Regression fixtures deliberately give both units a pin at the same local coordinate, then place those units on different labelled nets. That shape makes a first-instance lookup return a plausible but demonstrably wrong answer instead of merely failing to find a pin.

The sole remaining `extract_lib_pins` call in these files is intentional and documented: design-review coverage compares the complete definition with the selected unit only to classify a component as multi-unit. All electrical analysis uses the unit-filtered set.

This PR covers the read/analysis/export half of #182. The component mutation paths are being submitted separately, and #267 contains the additional intentional-NC/power-pin behavior for `validate_component_connections`; the one shared unit-filtering line may need a trivial rebase depending on merge order.

## Compatibility and safety

- Existing schemas are unchanged.
- `get_component_nets`, `get_net_components`, and `get_connected_items` gain additive `unit` fields.
- `export_netlist_summary` gains `unit`; each entry's pins now describe that placed unit rather than every unit superimposed at its coordinates.
- Multi-unit design review is now complete when all other coverage is available instead of being forcibly marked partial.
- `fix_connectivity` keeps its existing atomic write path but will no longer propose or apply a snap to an unplaced unit's phantom endpoint.
- Rollback is a single revert of commit `1693251`.

## Validation

- 18 focused Multi-Unit tests pass, including 8 new call-site regressions across analysis, both validators, netlist summary, connectivity repair, and design review
- fixtures cover equal local pin coordinates, distinct placed coordinates/nets, a phantom near-snap target, an unplaced power pin, and a real unwired unit pin
- `cargo test --workspace --locked --lib --tests`: all suites pass except the same three pre-existing local `update_symbols_from_library_*` failures caused by the installed KiCad `Device:R` library shadowing the test fixture; `konnect-core` reaches 550/553 passing tests
- HTTP and stdio protocol suites pass

### DIFF
--- a/crates/konnect-core/src/tools/design_review.rs
+++ b/crates/konnect-core/src/tools/design_review.rs
@@ -171,7 +171,7 @@ async fn handle_audit_decoupling(
             None => continue,
         };
 
-        let pins = extract_lib_pins(lib_sym);
+        let pins = extract_lib_pins_for_unit(lib_sym, inst.unit);
         let is_passive = inst.lib_id.contains("R_")
             || inst.lib_id.contains("C_")
             || inst.lib_id.contains("L_")
@@ -268,7 +268,7 @@ async fn handle_audit_connections(
             None => continue,
         };
 
-        let pins = extract_lib_pins(lib_sym);
+        let pins = extract_lib_pins_for_unit(lib_sym, inst.unit);
 
         // Check for I2C pull-ups
         if has_i2c_pins(&pins) {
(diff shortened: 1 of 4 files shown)


pauliuszaleckas:
Heads-up on overlap from #323 (`refactor(sch): give the connectivity tools one shared index`), which rewrites `validate_wire_connections` and `validate_component_connections` — the two handlers your `sch_batch` hunks edit.

It is on a different axis: it unifies what those tools count as *connected* (four tolerances, four item sets), not how pins are resolved. But because it sources pins from `placed_pins_by_reference` — unit-aware since #35/#143 — the two `sch_batch` validators inherit unit-awareness as a side effect, so your two `extract_lib_pins` → `extract_lib_pins_for_unit` lines there are reached by a different route.

**Your `design_review` and `sch_export` hunks are untouched by it and still needed** — those are the bulk of this PR and #323 does not go near them.

This PR is currently `CONFLICTING` against main independently of #323. If it lands first, #323 rebases cleanly on top (the affected lines get replaced wholesale). No action needed from you either way — flagging it so the `sch_batch` overlap is not a surprise at merge time.

mixelpixx:
Superseding notice: #323's shared ConnectivityIndex reaches the same unit-aware pin resolution across all the analysis call sites, and its author's overlap analysis (which I verified) shows this branch's remaining hunks become redundant once it lands — GitHub already marks this CONFLICTING against main. Closing in favor of #323 with thanks: this PR's framing is what established the read-half/mutation-half split for #182, and #273 (the mutation half) remains open and wanted after a rebase onto the index.

PR #273, opened 2026-08-18
fix(schematic): update every placed component unit

nordic-style:
## Summary

- delete every placed unit instead of leaving orphan units behind
- translate and rotate all units by one shared delta, preserving their relative layout
- update shared fields, reference paths, annotations, and groups across every unit in one revision-aware command
- target `move_region` placements by UUID so selecting unit 2 never moves unit 1
- replace every unit's library id atomically while preserving and validating unit numbers
- aggregate component and pin-location reads across all placements

Refs #182

## Approach

A reference designator identifies one logical component, but KiCad stores a multi-unit component as one top-level symbol block per placed unit. Every reference-based component operation now resolves the complete block/placement set. Component-wide moves use the lowest-numbered unit as the requested anchor and apply its snapped delta to the other units; rotation follows the same rule so relative orientation is preserved.

Shared properties are built as one edit set against the original document revision. Existing copies are updated everywhere, missing custom-field copies are repaired per unit, and all user text is escaped before insertion or replacement. Reference renames update both rendered properties and every netlist `instances` path.

Component replacement validates all preserved unit numbers against the new symbol, updates every `lib_id`, and writes once. Supplying a single `unit` for a reference with multiple placements is rejected before any write because the request is ambiguous.

This is the mutation half of #182. #272 covers analysis/export/validation/review call sites; #267 covers the additional intentional-NC and power-pin behavior in `validate_component_connections`.

## Compatibility and safety

- Single-unit behavior remains compatible; response objects gain additive unit counts and placement arrays.
- `move_schematic_component` and `rotate_schematic_component` now operate on the logical component instead of silently changing only the first block.
- `get_schematic_component` and pin-location tools now include every placement; the original top-level position remains the lowest-numbered unit's anchor.
- `replace_component.unit` is intentionally rejected for multi-placement references; omit it to preserve every existing unit.
- All text edits use one revision-aware command; typed move/delete paths still overwrite once after the complete in-memory mutation.
- An ambiguous replacement regression verifies the schematic remains byte-identical on error.
- Rollback is a single revert of commit `51c8f60`.

## Validation

- 12 focused end-to-end Multi-Unit tests pass: delete, move, rotate, value/custom-field edit with escaping, reference rename, partial annotation repair, grouping, component summary, pin locations, full replacement, ambiguous replacement rollback, and unit-specific region move
- `cargo test --workspace --locked --lib --tests`: all suites pass except the same three pre-existing local `update_symbols_from_library_*` failures caused by the installed KiCad `Device:R` library shadowing the test fixture; `konnect-core` reaches 554/557 passing tests
- HTTP and stdio protocol suites pass

### DIFF
--- a/crates/konnect-core/src/tools/sch_components.rs
+++ b/crates/konnect-core/src/tools/sch_components.rs
@@ -7,9 +7,8 @@
 use crate::mcp::protocol::CallToolResult;
 use crate::tool;
 use crate::tools::{
-    find_all_symbol_instance_blocks, find_symbol_instance_block, get_path, opt_f64, opt_str,
-    reembed_lib_symbols, require_array, require_f64, require_str, ReembedOutcome, ToolContext,
-    ToolDef,
+    find_all_symbol_instance_blocks, get_path, opt_f64, opt_str, reembed_lib_symbols,
+    require_array, require_f64, require_str, ReembedOutcome, ToolContext, ToolDef,
 };
 use konnect_schematic_editor as cse;
 use konnect_sexp::{
@@ -101,7 +100,7 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "delete_schematic_component",
-            "Remove a symbol instance from the schematic by its reference designator.",
+            "Remove a component by reference designator, including every placed unit of a multi-unit symbol.",
             json!({
                 "type": "object",
                 "properties": {
@@ -114,7 +113,7 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "edit_schematic_component",
-            "Update fields (Reference, Value, Footprint, custom properties) of a symbol instance.",
+            "Update fields (Reference, Value, Footprint, custom properties) consistently across every placed unit of a component.",
             json!({
                 "type": "object",
                 "properties": {
@@ -135,7 +134,7 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "get_schematic_component",
-            "Get all properties, position, and pin locations for a symbol instance.",
+            "Get a component's shared properties and every placed unit's position. Use get_schematic_pin_locations for pins.",
             json!({
                 "type": "object",
                 "properties": {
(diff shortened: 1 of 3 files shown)


mixelpixx:
Rebased onto main (post-#323) and verified — merging once CI is green. The port was far cleaner than the diff size suggested: main's connectivity refactor never touched the handlers you rewrote, so everything applied verbatim except the classic both-branches-append-tests conflict, resolved by keeping both modules.

**Two fixups on your branch:**

1. **Rotation normalization** (`ddf8be0` → amended in place): the delta arithmetic could write an angle eeschema never stores — a unit at 270° following a +90° turn computed 360°, which went into the file and the response as-is. eeschema keeps 0/90/180/270 and re-saves anything else, so the unnormalized value survived exactly until KiCad touched the file and then silently diverged from what the tool had reported. Angles now normalize with `rem_euclid(360.0)` before writing; a two-turn test pins that 270°+90° stores 0°, and the neutered check fails it.

2. **A KiCad-authored fixture** (`7aa9a1b`): your test schematic is hand-written in Konnect's own serialization style (2-space, LF, single-line properties), which leaves `property_insert_edit`'s indentation-matching branch untested against what eeschema actually writes. Checked in the ecc83 demo schematic verbatim (tabs, CRLF, U1 placed as **three** units of the embedded dual triode — both triodes plus the heater), with tests that the whole component is seen (`unit_count: 3`) and that an added annotation lands tab-indented in every unit block. Both passed against your code unmodified — the dialect handling was already right; now it stays that way.

**Verification:** full gate green (1135 tests). E2E through the real server binary against the demo file, results read from the file: `move_schematic_component` applies one shared delta to all 3 units (25.4, 12.7 exactly, per-unit); `edit_schematic_component` lands the Value in all 3 blocks; `rotate_schematic_component` reports and stores identical normalized angles. The by-UUID `move_region` fix and the atomic multi-unit replace are covered by your own tests, which are thorough — the byte-identical-file assertion on the refused ambiguous `unit` override is exactly the right property.

This closes the mutation half of #182 that #323's index left open. One heads-up: #328 (the connectivity model is bus-blind) is adjacent territory if you're looking for the next one.

PR #274, opened 2026-08-18
fix(library): inspect footprint structure

### LINKED ISSUE #248: [BUG] get_footprint_info reports pad_count 0 for every real KiCad footprint
Reported by @neusse in discussions #239 and #165, and by @woozer in #152 section 4. Confirmed, with a concrete root cause.

## Symptom

`get_footprint_info` finds a footprint and returns a normal success response with `pad_count: 0`, for footprints that unambiguously have pads. Reported for `Package_SO:SOIC-8_3.9x4.9mm_P1.27mm` and `Connector_USB:USB_C_Receptacle_HRO_TYPE-C-31-M-12`; in #152, for project-local footprints with alphanumeric and duplicate pad numbers.

Library resolution is not the problem — `resolve_footprint_path` (`library.rs:1418-1508`) returns a real `Err` when it cannot find the file, so a resolution failure surfaces as a tool error, not as a zero.

## Root cause

`crates/konnect-core/src/tools/library.rs:3580`:

```rust
let pad_count = content.matches("\n  (pad ").count();
```

A literal substring match requiring a newline followed by **exactly two spaces**. That is the indentation Konnect's own `handle_create_footprint` writer emits (`library.rs:767-770`), and exactly what the unit fixture at `library.rs:5216-5226` uses.

Real KiCad footprints are **tab-indented and CRLF**. From the file on disk, `C:\KiCad\10.0\share\kicad\footprints\Package_SO.pretty\SOIC-8_3.9x4.9mm_P1.27mm.kicad_mod` (`cat -A`):

```
(footprint "SOIC-8_3.9x4.9mm_P1.27mm"^M$
^I(version 20260206)^M$
...
^I(pad "1" smd roundrect^M$
```

`"\n  (pad "` matches neither `"\r\n\t(pad "` nor anything else in that file. `pad_count` is 0 for every stock KiCad footprint, and the response reports it as a successful result.

## Why the test suite could not catch it

The fixture is generated in Konnect's own 2-space, LF style. The parser and the fixture share the same wrong assumption, so the test passes and proves nothing about real input. This is the same self-referential-fixture failure as the `flip_component` / `ki_fp_filters` bug found in #230, and CONTRIBUTING now carries a rule about it that has no corpus behind it yet.

## Related

This is the same defect class as #84 (indentation-dependent scans), which lists `sch_export`, `integration` and `schematic_builder` but not `library.rs`. Filing separately because `pad_count: 0` is the symptom a user would search for.

## Direction

Replace the substring scan with the structural S-expression helpers used elsewhere (`parse_sexp` then per-footprint `find_all("pad")`), which is indentation- and line-ending-agnostic. While there, check the neighbouring `content.contains(...)` scans in the same function (`has_courtyard`, `has_3d`) — those happen to be whitespace-insensitive, but they are the same style of guess.

Check in a real KiCad-authored `.kicad_mod` (tab-indented, CRLF) as a fixture so the next assumption of this kind has somewhere to fail.

nordic-style:
## Summary

`get_footprint_info` now reports pads from real KiCad footprint files instead of returning zero whenever their indentation or line endings differ from a hand-written fixture.

Closes #248

## Approach

The old implementation counted occurrences of a newline followed by exactly two spaces and `(pad `. Stock KiCad 10 libraries are tab-indented, and CRLF files invalidate that source-text assumption as well.

Parse the footprint once and inspect its direct S-expression children instead:

- direct `pad` nodes determine `pad_count`, independent of pad number syntax;
- child nodes whose `layer` is `F.CrtYd` or `B.CrtYd` determine courtyard presence;
- direct `model` nodes determine 3D-model presence.

The regression fixture is reduced from KiCad 10's stock `Connector_USB:USB_C_Receptacle_HRO_TYPE-C-31-M-12` output without rewriting the retained nodes, then stored with CRLF to cover the reported line-ending case. It includes tabs, unnumbered mechanical pads, alphanumeric pad numbers, and duplicate `SH` numbers. A second regression proves that pad/model/courtyard-looking words in metadata are not counted as structure.

## Compatibility and safety

No schema, tool name, argument, or response-shape changes. Correctly formed footprints now produce format-independent summary values. Malformed non-KiCad content is rejected by the existing parser rather than receiving a heuristic summary.

This is a read-only handler; it performs no file or IPC mutation.

## Validation

- [x] `cargo test -p konnect-core get_footprint_info_ -- --nocapture` — 3 passed
- [ ] `cargo test --workspace --locked --lib --tests` — 544 passed, 3 failed; the failures are the same `update_symbols_from_library_*` baseline failures reproduced on unmodified `main`, caused locally by the installed `Device:R` library shadowing the test fixture
- [x] `cargo test --workspace --locked --doc` — passed (7 passed, 2 intentionally ignored)
- [x] Relevant real-KiCad check — regression nodes copied from the installed KiCad 10 stock HRO USB-C footprint; no live GUI is involved in this read-only file handler

## Review checklist

- [x] New names follow `docs/NAMING_CONVENTIONS.md`; no public names changed.
- [x] New behavior and false-positive paths have regression coverage.
- [x] File mutation requirements are not applicable; the tool only reads the footprint.
- [x] IPC mutation requirements are not applicable.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -3566,6 +3566,7 @@ async fn handle_get_footprint_info(
     };
 
     let content = tokio::fs::read_to_string(&path).await?;
+    let footprint = parse_sexp(&content)?;
 
     // Parse basic info: description, pads
     let description = extract_sexp_string(&content, "descr").unwrap_or_default();
@@ -3576,12 +3577,16 @@ async fn handle_get_footprint_info(
             .to_string()
     });
 
-    // Count pads
-    let pad_count = content.matches("\n  (pad ").count();
-
-    // Extract courtyard bbox (gr_poly on B.CrtYd or F.CrtYd) — simplified
-    let has_courtyard = content.contains("B.CrtYd") || content.contains("F.CrtYd");
-    let has_3d = content.contains("(model ");
+    // KiCad controls indentation and line endings, so these properties must be
+    // read from the parsed footprint rather than inferred from source text.
+    // Pads and models are direct children of a `(footprint ...)` node.
+    let pad_count = footprint.find_all("pad").len();
+    let has_courtyard = footprint
+        .children()
+        .unwrap_or(&[])
+        .iter()
+        .any(|node| matches!(node.find_str("layer"), Some("B.CrtYd" | "F.CrtYd")));
+    let has_3d = !footprint.find_all("model").is_empty();
 
     let mut response = json!({
         "name": fp_name,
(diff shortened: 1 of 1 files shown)


pauliuszaleckas:
Commit message is malformed, contains "/n/n" instead of real new-lines. Otherwise looks good.

pauliuszaleckas (approved):

PR #275, opened 2026-08-18
fix(schematic): measure placed symbol geometry

nordic-style:
## Summary

Make schematic placement queries describe the visible symbol geometry they claim to measure:

- `get_schematic_layout.bounding_box` encloses transformed symbol drawings and pin extents instead of component origins;
- `check_schematic_overlaps` detects intersecting symbol envelopes even when their anchors differ.

No linked issue.

## Approach

Add one structural symbol-bounds implementation in `konnect-sexp` and share it between both handlers. It:

- reads rectangles, polylines, Bezier control hulls, circles, exact three-point arc extrema, explicit text boxes, and complete pin segments;
- selects common (`_0_M`) plus the placed unit's (`_N_M`) geometry for multi-unit symbols;
- applies KiCad's schematic Y-axis convention, absolute rotation, and both mirror modes through the existing canonical transform;
- deliberately excludes free library/property text because KiCad does not provide an exact font-layout box there.

`check_schematic_overlaps` reports positive-area intersections and leaves edge/pin contact valid. When an embedded library definition cannot be resolved, both tools preserve an origin fallback and name every affected reference in `bounds_unresolved` rather than presenting partial geometry as complete.

The regression uses retained nodes from KiCad 10's stock `Device:R` output and covers rotation, multi-unit selection, exact arcs, distinct-origin collisions, touching edges, invalid tolerance, and unresolved geometry.

## Compatibility and safety

No tool, argument, or schema removal.

`get_schematic_layout` keeps its existing fields and adds per-component `unit`/`bounds`, aggregate width/height, and resolution coverage. Its existing `bounding_box` values intentionally change from origin extrema to the actual geometry promised by the field name.

`check_schematic_overlaps` keeps its existing overlap list and adds unit, bounds, detection source, intersection depth, and resolution coverage. Conflicting-label detection remains coordinate based.

Both tools are read-only and perform no file or IPC mutation.

## Validation

- [x] `cargo test -p konnect-sexp --lib` — 137 passed
- [x] `cargo test -p konnect-core layout_bounds_tests -- --nocapture` — 2 passed
- [x] `cargo test -p konnect-core placement_overlap_tests -- --nocapture` — 3 passed
- [ ] `cargo test --workspace --locked --lib --tests` — 547 passed, 3 failed; the failures are the same `update_symbols_from_library_*` baseline failures reproduced on unmodified `main`, caused locally by the installed `Device:R` library shadowing the test fixture
- [x] `cargo test --workspace --locked --doc` — passed (7 passed, 2 intentionally ignored)
- [x] Real KiCad check through MCP — a 102-instance KiCad 10-authored schematic resolved 102/102 unit bounds with no fallback; its collision query completed with no false positives

## Review checklist

- [x] New names follow `docs/NAMING_CONVENTIONS.md`; no public name was removed or renamed.
- [x] New behavior, transformation cases, and fallback paths have regression coverage.
- [x] File mutation requirements are not applicable; both handlers only read.
- [x] IPC mutation requirements are not applicable.

### DIFF
--- a/crates/konnect-core/src/tools/sch_analysis.rs
+++ b/crates/konnect-core/src/tools/sch_analysis.rs
@@ -12,8 +12,7 @@ use konnect_sexp::{
     geometry::{point_on_segment, points_coincident},
     schematic::{
         extract_all_net_labels, extract_labels, extract_symbol_instances, extract_wires,
-        find_lib_symbol, read_schematic, symbol_bounds_for_instance, LabelKind, SymbolBounds,
-        Wire,
+        find_lib_symbol, read_schematic, symbol_bounds_for_instance, LabelKind, SymbolBounds, Wire,
     },
 };
 use serde_json::json;
--- a/crates/konnect-core/src/tools/sch_analysis.rs
+++ b/crates/konnect-core/src/tools/sch_analysis.rs
@@ -11,8 +11,9 @@ use konnect_schematic_editor as cse;
 use konnect_sexp::{
     geometry::{point_on_segment, points_coincident},
     schematic::{
-        extract_all_net_labels, extract_symbol_instances, extract_wires, find_lib_symbol,
-        read_schematic, LabelKind, Wire,
+        extract_all_net_labels, extract_labels, extract_symbol_instances, extract_wires,
+        find_lib_symbol, read_schematic, symbol_bounds_for_instance, LabelKind, SymbolBounds,
+        Wire,
     },
 };
 use serde_json::json;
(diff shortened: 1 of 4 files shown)


mixelpixx:
Rebased onto main and verified — merging once CI is green. This closes a quiet instance of the honest-response class: `get_schematic_layout` reported a "bounding_box" that was really a hull of component *origins*, and `check_schematic_overlaps` compared anchors, not envelopes. Your structural bounds (rectangles, polylines, Bézier hulls, exact arc extrema, full pin segments, through the canonical transform) replace a plausible number with a measured one, and the `bounds_unresolved` reporting instead of a silent origin fallback is exactly the house pattern.

**The rebase** (`a093ae3` on your branch): three conflicts, all mechanical — the `sch_analysis` import union (your `extract_junctions` import was unused and would have failed the `-D warnings` gate, so it's dropped), the classic both-sides-appended-test-modules in `sch_batch` (kept both), and the tool-directory rows (kept main's #308-era `add_schematic_text` description alongside your new `get_schematic_layout` one). `find_lib_symbol` needed restoring into `sch_batch`'s import list after the union.

**One test added on top, and it's the one I cared about most:** `get_schematic_layout` run over a full eeschema save (KiCad's ecc83 demo, now a checked-in fixture) where U1 is placed as **three units** of the embedded dual triode. All three placements resolve bounds, and the boxes come in at least two distinct sizes — the heater unit's geometry differs from the triodes' — which pins your `_0_M` + `_N_M` unit selection against what eeschema actually writes rather than a reduced fixture. It passed against your code unmodified: the multi-unit selection was right on first contact with real data, which given that #273 (update-every-unit semantics) landed only hours before this rebase is a genuinely good sign for the design.

Full gate green at 1156 tests.

PR #277, opened 2026-08-18
fix(export): apply every advertised output option

nordic-style:

Several export tools advertised output controls but silently used KiCad's defaults: schematic SVG/PDF monochrome settings, SVG themes, root-sheet-only PDF export, PCB PDF/SVG monochrome output, and the 3D unspecified-model policy. The BOM schema also suggested arbitrary formats although KiCad 10 produces CSV here.

This PR makes each advertised value affect the `kicad-cli` invocation and narrows BOM format validation to the supported result.

Refs #251

Depends on #266, which supplies the corrected single-file PCB PDF/SVG argument builder. Once #266 merges, this PR reduces to one focused follow-up commit.

## Root cause and design

Handlers read or defaulted the schema values, then called CLI wrappers whose signatures had no corresponding options. The responses could therefore echo values that had not affected the artifact.

Typed option structs and pure argument builders now carry the values to KiCad. PCB PDF/SVG add `--black-and-white`; schematic exports add `--black-and-white`, `--theme`, or `--pages 1`; 3D export adds `--no-unspecified` when exclusion is requested. An unsupported BOM format returns a structured `invalid_argument` instead of producing a differently formatted file.

Existing field names and defaults remain unchanged, but they now do what the public schema says. `export_bom.format` is explicitly limited to `csv`; callers requesting another value now get a typed error rather than a misleading success.

No source schematic or board is modified.

## Validation

- `cargo test -p konnect-core schematic_export_option_tests --locked`
- `cargo test -p konnect-core pcb_plot_export_tests --locked`
- `cargo test -p konnect-core three_d_export_option_tests --locked`
- `cargo test -p konnect-core new_export_format_tests --locked`

Live GUI export was not required; the changed behavior is covered at the exact CLI argument boundary.

## Risk and rollback

Risk is limited to export flags and response metadata. Older KiCad versions lacking these KiCad 10 flags will return their normal CLI error rather than silently ignoring the request. Rollback is a single commit revert after #266.

### DIFF
--- a/crates/konnect-core/src/tools/cli.rs
+++ b/crates/konnect-core/src/tools/cli.rs
@@ -364,35 +364,91 @@ pub async fn annotate_schematic(_cli: &str, schematic: &Path) -> Result<()> {
 
 // ─── Schematic Export ────────────────────────────────────────────────────────
 
-/// KiCAD 10: `sch export svg --output <dir> <input>`
+#[derive(Debug, Default, Clone)]
+pub struct SchematicSvgOptions<'a> {
+    pub black_and_white: bool,
+    pub theme: Option<&'a str>,
+}
+
+fn schematic_svg_args<'a>(
+    output_dir: &'a str,
+    schematic: &'a str,
+    options: &'a SchematicSvgOptions<'a>,
+) -> Vec<&'a str> {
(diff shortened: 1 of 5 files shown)


nordic-style:

PR #278, opened 2026-08-18
fix(api): remove ignored tool parameters

nordic-style:

Eight optional MCP inputs were advertised even though their handlers never read them. A caller could reasonably believe those values changed validation, costing, review, hierarchy, zone refill, or manufacturing output when they did nothing.

This PR removes only those dead inputs, corrects overstated descriptions, and records a supported replacement workflow for every removal.

Refs #251

## Root cause and design

Schemas and handlers evolved independently. The unused fields survived because existing tests checked registration and required arguments, but not whether every advertised property reached implementation behavior.

The schemas now match the operations Konnect can verify. `docs/API_MIGRATIONS.md` is the explicit compatibility record, and `schema_migrations.rs` prevents a removed field from returning without a migration note.

Removed fields:

- `import_sheet_pins.project_name`
- `refill_zones.zones`
- `run_drc.tests`
- `audit_decoupling.board`
- `audit_decoupling.max_distance_mm`
- `export_manufacturing_package.quantity`
- `validate_for_manufacturing.schematic`
- `estimate_cost.schematic`

All were optional and behaviorally inert, so generated files and analysis results are unchanged. The migration document points callers to project-instance workflows, complete-board zone refill, KiCad DRC configuration, PCB placement inspection, `estimate_cost.quantity`, `check_bom_health`, and board-footprint counting as appropriate.

## Validation

- `cargo test -p konnect --test schema_migrations --locked`

## Risk and rollback

The intentional API narrowing may expose clients that were sending inert fields. Migration is omission or the documented companion tool; no data format or file write changes. Rollback is one commit, including the migration and regression test.

### DIFF
--- a/crates/konnect-core/src/tools/design_review.rs
+++ b/crates/konnect-core/src/tools/design_review.rs
@@ -28,18 +28,13 @@ pub fn tools() -> Vec<ToolDef> {
     vec![
         tool!(
             "audit_decoupling",
-            "Check that all ICs have appropriate decoupling capacitors. Finds power pins \
-             without nearby capacitors and flags wrong values. The #1 most common PCB design mistake.",
+            "Audit schematic connectivity between IC power nets and decoupling capacitors. \
+             This does not inspect PCB placement distance; use PCB clearance/placement tools \
+             for a physical review.",
             json!({
                 "type": "object",
                 "properties": {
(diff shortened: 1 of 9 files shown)


nordic-style:

PR #279, opened 2026-08-18
fix(project): verify the requested open board

nordic-style:

`open_project.path` accepted either a project or board path, but the handler ignored the requested path and treated any successful PCB IPC ping as proof that the requested project was open. With multiple projects, it could report success for the wrong board.

This PR resolves `.kicad_pro` paths to their sibling `.kicad_pcb`, queries KiCad's open boards, and succeeds only when the requested board is actually active/open.

Refs #251

## Root cause and design

The old implementation used a liveness check without reading `path`. The IPC client now exposes normalized open-board paths, and the handler compares the requested canonical board against that list. Unsupported extensions, missing sibling boards, unreachable IPC, and a different open board return structured, actionable failures.

Both documented input forms remain supported. Correct calls keep the same success behavior; false-positive calls now fail and name the requested/open board state. This is a correctness tightening, not a schema change.

The tool remains a verifier: it does not launch KiCad or modify project files.

## Validation

- `KONNECT_STATE_DIR=/private/tmp/konnect-open-project cargo test -p konnect-core tools::project::tests --locked`

## Risk and rollback

Risk is confined to IPC path normalization and refusal behavior. No files are written. If a platform-specific path form is missed, the safe failure is a refusal rather than acting on the wrong project. Rollback is a single commit revert.

### DIFF
--- a/crates/konnect-core/src/tools/project.rs
+++ b/crates/konnect-core/src/tools/project.rs
@@ -4,7 +4,7 @@
 //!
 //! KiCAD interface:
 //!   - create_project   → file system (template)
-//!   - open_project     → IPC ping (check if project is open)
+//!   - open_project     → IPC open-document query
 //!   - save_project     → IPC board.save()
 //!   - get_project_info → file system read
 //!   - snapshot_project → kicad-cli export PDF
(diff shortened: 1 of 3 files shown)


nordic-style:

PR #280, opened 2026-08-18
fix(schematic): honor ignored power pins

nordic-style:

`validate_component_connections(ignore_power_pins: true)` advertised that power pins would be ignored, but the handler never read the flag. Designs could therefore report unwanted unconnected-power findings even when the caller explicitly excluded them.

This PR parses KiCad library pin electrical types and omits only `power_in` and `power_out` pins when requested.

Refs #251

Depends on #272, which makes placed-unit pin resolution correct for multi-unit symbols. Once #272 merges, this PR reduces to one focused follow-up commit.

## Root cause and design

The validator had pin coordinates and numbers but not the library electrical type, so the schema value was inert. `LibPin` now carries the parsed KiCad type. Validation applies the option after resolving the actual placed unit, reports how many power pins were ignored, and leaves all non-power pins under the existing rules.

The public field and its default remain unchanged. `false` preserves existing behavior. `true` now implements the documented behavior; no schematic is modified.

## Validation

- `KONNECT_STATE_DIR=/private/tmp/konnect-power-pins cargo test -p konnect-core power_pin_validation_tests --locked`

## Risk and rollback

Risk is limited to classification of KiCad's `power_in`/`power_out` pin tokens. Unknown types remain validated rather than being skipped. Rollback is one commit after #272.

### DIFF
--- a/crates/konnect-core/src/tools/design_review.rs
+++ b/crates/konnect-core/src/tools/design_review.rs
@@ -171,7 +171,7 @@ async fn handle_audit_decoupling(
             None => continue,
         };
 
-        let pins = extract_lib_pins(lib_sym);
+        let pins = extract_lib_pins_for_unit(lib_sym, inst.unit);
         let is_passive = inst.lib_id.contains("R_")
             || inst.lib_id.contains("C_")
             || inst.lib_id.contains("L_")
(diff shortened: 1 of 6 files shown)


mixelpixx:
Duplicate of #267's power-pin filter — same code in `sch_batch.rs`, filed under the #251 narrative instead of #182. See the series note on #285. Please drop this or reduce it to its non-duplicate remainder after #267 and #272 settle.

nordic-style:
Closing this duplicate in favor of #267, following the maintainer's guidance to keep the smaller implementation. The consolidated #251 branch is #285 and intentionally carries #267's fix until #267 lands; #280 itself is not included.

PR #281, opened 2026-08-18
fix(pcb): create rounded board outlines

nordic-style:

`add_board_outline.corner_radius` was documented and echoed in the response, but every request still produced a sharp four-line rectangle. Callers could not create the rounded circular boards/enclosures the API promised.

This PR generates a real closed KiCad outline made from shortened lines and circular corner arcs.

Refs #251

## Root cause and design

The handler read `corner_radius` only for response metadata. A shared geometry builder now normalizes the bounding box, validates the radius against half the shorter side, and produces four tangent arcs plus non-zero straight sections. Both IPC messages and the safe file fallback consume the same primitives. Capsule geometry omits zero-length sides instead of emitting degenerate lines.

`corner_radius: 0` preserves the existing rectangular result. Positive valid radii now implement the documented shape. Invalid/overlapping radii return a structured `invalid_argument` without writing.

The file fallback writes standard `gr_line`/`gr_arc` S-expressions and reparses the result in regression coverage.

## Validation

- `KONNECT_STATE_DIR=/private/tmp/konnect-rounded-outline cargo test -p konnect-core rounded_outline_tests --locked`
- `KONNECT_STATE_DIR=/private/tmp/konnect-rounded-outline-file cargo test -p konnect-core rounded_outline_file_fallback_writes_real_kicad_arcs --locked`

## Risk and rollback

This changes board-outline geometry only when a positive radius is requested. The main risk is arc orientation/continuity; tests assert a closed path, tangent endpoints, primitive counts, capsule handling, and a valid KiCad roundtrip. Rollback is a single commit revert.

### DIFF
--- a/crates/konnect-core/src/tools/pcb_board.rs
+++ b/crates/konnect-core/src/tools/pcb_board.rs
@@ -33,6 +33,146 @@ fn rect_outline_items(x1: f64, y1: f64, x2: f64, y2: f64, w: f64) -> Vec<prost_t
         .collect()
 }
 
+#[derive(Debug, Clone, PartialEq)]
+enum OutlinePrimitive {
+    Line {
+        start: (f64, f64),
+        end: (f64, f64),
+    },
+    Arc {
+        start: (f64, f64),
+        mid: (f64, f64),
+        end: (f64, f64),
+    },
+}
+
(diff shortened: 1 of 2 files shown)


nordic-style:

PR #282, opened 2026-08-18
fix(library): apply pad shape edits

nordic-style:

`edit_footprint_pad.shape` accepted a new pad shape but never changed the footprint. Width or height edits also took effect only when both were supplied. The response could therefore claim a successful edit while leaving some requested geometry untouched.

This PR applies supported shape changes and independent dimension updates to the actual pad S-expression.

Refs #251

## Root cause and design

The editor only rewrote position, size-as-a-pair, drill, layers, and pad number. It now locates the structural shape token, validates `circle`, `rect`, `oval`, and `roundrect`, removes children that are invalid for the destination shape, and adds a conservative default round-rectangle ratio when needed. Width and height each preserve the unspecified counterpart.

Existing calls without `shape` behave as before, except that a one-axis size update now works as advertised. Unsupported shapes return a structured error before reading or writing the footprint. Changing away from round/chamfer/custom forms intentionally removes incompatible shape-specific children so the result remains valid KiCad syntax.

## Validation

- `KONNECT_STATE_DIR=/private/tmp/konnect-pad-shape cargo test -p konnect-core pad_shape --locked`
- `KONNECT_STATE_DIR=/private/tmp/konnect-pad-persist cargo test -p konnect-core edit_footprint_pad_persists_a_valid_shape_change --locked`
- `KONNECT_STATE_DIR=/private/tmp/konnect-pad-dimension cargo test -p konnect-core changing_one_pad_dimension_preserves_the_other --locked`

## Risk and rollback

Risk is confined to the selected pad block and guarded by revision-aware atomic persistence. Tests cover shape-specific cleanup, unsupported values, independent dimensions, and reparsing the persisted footprint. Rollback is a single commit revert.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -131,7 +131,11 @@ pub fn tools() -> Vec<ToolDef> {
                     "y": { "type": "number", "description": "New Y position in mm (optional)" },
                     "width": { "type": "number", "description": "New pad width in mm (optional)" },
                     "height": { "type": "number", "description": "New pad height in mm (optional)" },
-                    "shape": { "type": "string", "description": "New pad shape (optional)" },
+                    "shape": {
+                        "type": "string",
+                        "enum": ["circle", "rect", "oval", "roundrect"],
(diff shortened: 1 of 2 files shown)


nordic-style:

PR #283, opened 2026-08-18
fix(verification): bound the KiCad UI health check

nordic-style:

`check_kicad_ui.timeout_seconds` was advertised but ignored. A caller could request a short health-check deadline and still wait on the fixed process/IPC behavior, while the response implied the timeout was meaningful.

This PR applies one bounded deadline across the complete KiCad process check and IPC ping. It also recognizes standalone `pcbnew`/`eeschema` processes and treats responsive IPC as authoritative running evidence.

Refs #251

## Root cause and design

The handler never read the schema field and performed its phases independently. It also checked only the top-level `kicad` process and skipped IPC entirely when that process was absent, so a standalone macOS `pcbnew` returned `running: false` while `open_project` was successfully using its socket.

The handler now validates an integer range of 1–300 seconds, wraps the complete health future in one Tokio timeout, checks `kicad`, `pcbnew`, and `eeschema`, and always probes IPC. A responsive endpoint is sufficient running evidence; `process_detected` remains separate diagnostic evidence. Timeout failures are typed and do not wait for the abandoned inner future.

The existing default remains 5 seconds. Valid callers now get the documented bound; invalid types/ranges get a structured `invalid_argument`. The check remains read-only.

## Validation

- `cargo test -p konnect-core health_ --locked`
- `cargo test -p konnect-core standalone_editors_count_as_kicad_ui_processes --locked`
- `cargo test -p konnect-core responsive_ipc_is_sufficient_running_evidence --locked`
- Live macOS test with standalone `pcbnew`: `running: true`, `process_detected: true`, and `ipc_responsive: true` in 39 ms.

## Risk and rollback

Risk is limited to health-check timing and cancellation. No project state is touched. Tests use a deliberately pending future to prove the outer deadline returns promptly. Rollback is one commit.

### DIFF
--- a/crates/konnect-core/src/tools/verification.rs
+++ b/crates/konnect-core/src/tools/verification.rs
@@ -433,27 +433,49 @@ async fn handle_get_design_rules(
 
 // ─── KiCAD UI management ──────────────────────────────────────────────────────
 
-/// Check if the KiCAD GUI is running by scanning the process list.
+const KICAD_GUI_PROCESS_NAMES: &[&str] = &["kicad", "pcbnew", "eeschema"];
+
+fn is_kicad_process_name(name: &str) -> bool {
+    let file_name = std::path::Path::new(name.trim_matches('"'))
+        .file_name()
+        .and_then(|name| name.to_str())
+        .unwrap_or(name)
+        .to_ascii_lowercase();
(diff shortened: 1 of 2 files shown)


nordic-style:

PR #284, opened 2026-08-18
fix(pcb): validate the requested routing board

nordic-style:

Seven IPC-backed routing operations required a `board` path but never used it. The related pad-to-pad router read the file for pad coordinates but still did not verify that the same board was active in KiCad. If a different board was active, a valid request could route, query, modify, or delete copper on that unintended board.

This PR makes every routing IPC call verify that the requested board is the active KiCad board before performing the operation.

Refs #251

## Root cause and design

The routing toolset's local `ipc!` helper established connectivity but, unlike the component helper, did not resolve or compare `args["board"]`. The helper now reads the path once and calls `ensure_board_is_active` inside the same blocking IPC operation before executing the requested mutation/query.

The affected tools are `route_trace`, `route_pad_to_pad`, `add_via`, `delete_trace`, `query_traces`, `get_nets_list`, `modify_trace`, and `route_differential_pair` (including both route segments where applicable).

No schema changes. Calls naming the actually active board behave as before. Calls naming another board now fail safely instead of operating on the wrong file.

## Validation

- `KONNECT_STATE_DIR=/private/tmp/konnect-routing-board cargo test -p konnect-core pcb_routing --locked` (16 passed)
- The catalogue-wide guard PR independently verifies that `board` is now reached through the registered handler/macro call graph.

## Risk and rollback

This intentionally adds a refusal path before IPC routing mutations. The safe failure mode is preferable to cross-board modification. No file fallback is introduced. Rollback is a single commit revert.

### DIFF
--- a/crates/konnect-core/src/tools/pcb_routing.rs
+++ b/crates/konnect-core/src/tools/pcb_routing.rs
@@ -25,9 +25,15 @@ where
 }
 
 macro_rules! ipc {
-    ($ctx:expr, |$c:ident| $body:expr) => {{
+    ($ctx:expr, $args:expr, |$c:ident| $body:expr) => {{
         let addr = $ctx.config.ipc_address.clone();
-        match with_ipc(addr, move |$c| $body).await? {
+        let requested_board = get_path($args, "board")?;
+        match with_ipc(addr, move |$c| {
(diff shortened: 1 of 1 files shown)


nordic-style:

Issue #217, opened 2026-08-15
[BUG] Correct stale and invalid guidance in bundled skills


neusse:
## Summary

The six bundled skills currently installed for Claude contain stale catalog metadata, invalid example signatures, and behavioral claims that do not match the registered tool schemas or handlers. This is a shared asset-correctness bug, independent of adding another client.

Fix the shared guidance first so Claude users receive accurate instructions today and future Codex support in #216 installs the corrected assets rather than copying the defects.

## Evidence on current `main`

The main Konnect skill says there are **187 tools across 18 toolsets**. The current registry contains **200 domain tools across 19 toolsets**, plus 6 always-available meta-tools. A hard-coded total will continue to drift; the skill should direct clients to `list_toolboxes` instead.

Several call examples and capability descriptions are also wrong:

| Bundled guidance | Current schema/implementation |
| --- | --- |
| `route_pad_to_pad(from_reference, from_pad, to_reference, to_pad, width, layer)` | Requires `board`, `net_name`, `ref1`, `pad1`, `ref2`, and `pad2`. |
| `route_pad_to_pad` “will use vias if pads are on different layers” | The handler creates one or two tracks on one supplied layer and does not add a via. |
| `route_differential_pair(net_positive, net_negative, width, spacing)` and “matches trace lengths automatically” | Requires `board`, `net_pos`, `net_neg`, explicit endpoints, and other routing parameters; it creates parallel traces but is not a general length-matching router. |
| `export_manufacturing_package(output_dir, format)` | Requires `board` and `output_dir`; the selector is `fab_house`, not `format`. |
| `estimate_cost(quantity, fab_options)` | Requires `board`; those argument names do not exist in the schema. |

The existing `asset_references` tests pass because they validate toolset and tool names, not the parameter names or required fields in examples. Issue #183 is the necessary regression guard and should land with or before this correction.

The asset prose also contains Claude-specific wording and prompt-conflict language such as “Mandatory,” “loaded when,” and “no exceptions.” Shared skills should describe the operational safety rule without claiming they override user or higher-priority instructions.

## Proposed direction

1. Audit every `assets/skills/**/SKILL.md`, reference file, and bundled agent prompt against the live registry and handler behavior.
2. Remove hard-coded tool and toolset totals; use runtime discovery.
3. Correct all call-shaped examples to use exact named parameters and include required fields.
4. Remove or qualify behavioral claims the implementation does not provide.
[...]

PR #285, opened 2026-08-18
fix(api): apply or remove ignored tool parameters

### LINKED ISSUE #251: Tools accept parameters they never apply and report them as honoured

nordic-style:

Konnect advertised public MCP inputs that some handlers never read. Clients could therefore receive a success response even though the requested behavior was not applied.

This PR rebases and consolidates the remaining #251 fixes on current `main`. Each advertised input is now either implemented or deliberately removed with migration guidance. A catalogue-wide regression test prevents the same bug class from returning.

Closes #251

## Consolidated changes

- Apply manufacturing Gerber layers and position format, units, and side (#269).
- Apply advertised schematic, PCB, and 3D export options (#277).
- Remove inputs that KiCad/Konnect cannot honor and document their migrations (#278).
- Verify the board requested by project and routing tools (#279, #284).
- Add rounded board outlines and pad-shape updates (#281, #282).
- Bound KiCad UI health checks and recognize standalone editors (#283).
- Add a schema-to-handler guard without a tool or parameter allowlist (#285).

The branch also includes #267's smaller `ignore_power_pins` implementation as a temporary stacked prerequisite: the guard correctly detects that existing ignored input. #280 is intentionally excluded as its duplicate. Once #267 lands, this branch can be rebased to remove the duplicate commit.

Already-merged #266, #271, #274, and #276 are inherited from `main` and are not duplicated here. #272 and #280 are not part of this branch.

## Root cause and design

The public schemas and their handlers previously evolved independently, with no catalogue-wide proof that every input reached executable code. The new test starts from the registered tool catalogue, locates each real handler, and follows reachable helpers and macros. Its lexer ignores comments and ordinary, raw, byte, and character literals so examples cannot masquerade as implementation. Same-file helper resolution wins when names are duplicated; cross-file helpers remain discoverable when no local definition exists.

Every schema property must appear in its handler's reachable implementation. There is no tool/parameter allowlist. The complete 202-tool catalogue is checked in about 0.24 seconds locally.

Implemented inputs keep their public names and defaults and now affect the requested operation. Eight previously inert inputs are removed: `import_sheet_pins.project_name`, `refill_zones.zones`, `run_drc.tests`, `audit_decoupling.board`, `audit_decoupling.max_distance_mm`, `export_manufacturing_package.quantity`, `validate_for_manufacturing.schematic`, and `estimate_cost.schematic`.

Each removal and its supported replacement workflow is documented in `docs/API_MIGRATIONS.md` and enforced by `schema_migrations.rs`.

## Validation

All commands required by `CONTRIBUTING.md` pass on this exact consolidated branch:

The focused catalogue guard also passes 5/5 with `cargo test -p konnect --test schema_parameter_usage --locked -- --nocapture`. Live-GUI tests remain intentionally ignored by the upstream suite unless a running KiCad GUI/socket is supplied.

## Risk and rollback

This is intentionally an integration PR, but the runtime fixes remain separated into focused commits. The main risks are incorrect KiCad CLI argument mapping, geometry conversion, board identity checks, and UI-process detection. Direct argv tests, real-format fixtures, and focused regressions cover those paths. The guard's static call-graph risk is covered by lexer, helper, macro, and duplicate-name tests. Any runtime area can be rolled back independently by reverting its focused commit.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -2,6 +2,10 @@
 
 Internal reference for developing and maintaining the Rust port.
 
+New to the codebase? Start with the map in
+[docs/DEVELOPER_OVERVIEW.md](docs/DEVELOPER_OVERVIEW.md), then return here for
+the detailed implementation reference and current statistics.
+
 Repository-wide naming, public API, branch, and pull-request rules live in
 [docs/NAMING_CONVENTIONS.md](docs/NAMING_CONVENTIONS.md).
 
--- a/crates/konnect-core/src/tools/cli.rs
+++ b/crates/konnect-core/src/tools/cli.rs
@@ -536,16 +536,32 @@ pub async fn export_netlist(
 
 // ─── PCB Export ──────────────────────────────────────────────────────────────
 
-/// KiCAD 10: `pcb export gerbers --output <dir> <input>` (PLURAL!)
-pub async fn export_gerber(cli: &str, pcb: &Path, output_dir: &Path) -> Result<()> {
-    let args = [
-        "pcb",
-        "export",
-        "gerbers",
-        "--output",
-        output_dir.to_str().unwrap(),
-        pcb.to_str().unwrap(),
-    ];
+/// Argument vector for Gerber export. KiCad's plural `gerbers` subcommand
+/// accepts the complete selection as one comma-separated `--layers` value.
+fn gerber_args<'a>(output_dir: &'a str, pcb: &'a str, layers_csv: &'a str) -> Vec<&'a str> {
+    let mut args = vec!["pcb", "export", "gerbers", "--output", output_dir];
+    if !layers_csv.is_empty() {
+        args.push("--layers");
+        args.push(layers_csv);
+    }
+    args.push(pcb);
+    args
+}
+
+/// KiCad 10: `pcb export gerbers --output <dir> [--layers <csv>] <input>`
+/// (PLURAL!)
+pub async fn export_gerber(
+    cli: &str,
+    pcb: &Path,
+    output_dir: &Path,
+    layers: &[&str],
+) -> Result<()> {
+    let layers_csv = layers.join(",");
+    let args = gerber_args(
+        output_dir.to_str().unwrap_or(""),
+        pcb.to_str().unwrap_or(""),
+        &layers_csv,
+    );
     run_cli(cli, &args, LONG_TIMEOUT).await?;
     Ok(())
 }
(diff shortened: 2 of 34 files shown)


mixelpixx:
A maintainer note on this whole series (#264–#285, #294), since this draft is its integration branch.

First: thank you. The individual work here is genuinely good — where I deep-reviewed (#266, #278, #294), the root-cause analysis was accurate, the tests assert on real argv/fixtures rather than circularly on the code's own behavior, and #294 anticipates the exact safety hazard a reviewer would raise. Four PRs from this series are merged as of tonight (#266, #271, #274, #276), and more will follow.

But the series was submitted as ~22 independent PRs and it is not internally coordinated, which creates real review hazards I need you to fix before the rest can land:

1. **#267 and #280 are the same fix filed twice.** Both add the identical `ignore_power_pins` filter to `validate_component_connections` in `sch_batch.rs` — #267 framed as part of #182, #280 framed as part of #251 and stacked on #272. Whichever lands first breaks the other. Please pick one (I'd keep #267's framing, it's the smaller diff) and drop or reduce the other to its non-duplicate remainder.

2. **#266 and #270 independently rewrote the same `export_pdf` function** with different helper names. #266 is merged (it also fixed SVG, which #270 did not). #270's artifact-verification core (issue #252) is still wanted — please rebase it over current main so it keeps only that.

3. **Eight PR bodies repeat the same sentence verbatim** — "Combined-series CI commands: … 585/588 tests pass" — describing the *combined* branch, not each PR's own diff. That means no individual PR in the #251 series has independently stated evidence. Going forward, per-PR claims need per-PR runs; CONTRIBUTING asks for exactly that.

4. **This draft (#285) is the right shape for the #251 remainder.** Rather than me merging #269/#277/#278/#279/#281/#282/#283/#284 one at a time with cascading rebases, let's land the series through this branch: rebase it over current main (which now carries #259, #262, #263, #266, #271, #274, #276, #297, #298 — several of your satellites will shrink or empty out), un-draft it, and I'll review it as one unit with its catalogue-wide schema-parameter guard as the proof. One request on that guard: derive its exemptions from the schemas themselves, not an allowlist — a previous allowlist-based guard here accumulated 22 phantom entries and suppressed the evidence for the bug it was meant to catch (#217).

On identity: the branch prefixes say Codex, and the volume says automation. That's fine here — the code quality carries it — but the coordination failures above are exactly what an automated pipeline doesn't catch on its own. Run the cross-check between your own open PRs before filing, and this series becomes easy to take.

nordic-style:
Consolidation requested in the maintainer review is complete:

- rebased onto current `main` (`2183267`)
- #269, #277, #278, #279, #281, #282, #283, and #284 are consolidated here and their satellite PRs are closed
- #280 is closed as the duplicate; #267's smaller implementation is stacked here temporarily until #267 lands
- the catalogue guard derives registrations and handlers from the schemas/source and has no tool or parameter allowlist
- every command required by `CONTRIBUTING.md` passes locally; GitHub CI is 9/9 green and the branch has no conflicts

This PR is ready for review as the single remaining #251 path.

Issue #244, opened 2026-08-17
[BUG] update_pcb_from_schematic rewrites every footprint graphic as a phantom pad


mixelpixx:
Reported by @neusse in discussion #239 (end-to-end Codex/KiCad benchmark). Confirmed by reading the source; the mechanism explains the reported numbers exactly.

## Symptom

`update_pcb_from_schematic` completes a revision-checked apply without error, reports success, and leaves every complex footprint semantically destroyed. From the report:

- U1 (`Package_SO:SOIC-8_3.9x4.9mm_P1.27mm`): **28 pads instead of 8**, the 20 extra unnamed and at `(0,0)`; **graphics 0 instead of 20**.
- J1 (`Connector_USB:USB_C_Receptacle_HRO_TYPE-C-31-M-12`): **35 pads instead of 22**, 13 extra unnamed; **graphics 0 instead of 13**.

The extra pad count equals the lost graphic count exactly, in both cases. Graphics are being rewritten as pads.

## Root cause

`apply_footprint_fields`, `crates/konnect-core/src/tools/pcb_sync.rs:926-945`:

```rust
for child in &mut definition.items {
    let Ok(mut pad) = kiapi::board::types::Pad::decode(child.value.as_slice()) else {
        continue;
    };
    ...
    *child = konnect_ipc::builders::pack_any(&pad, "kiapi.board.types.Pad");
}
```

`definition.items` is a single repeated `Any` field holding **pads, graphic shapes and text mixed together** (`konnect-ipc/src/client.rs:1582-1591`). The loop never inspects `child.type_url`; it relies on `Pad::decode` returning `Err` to filter out non-pads. proto3 decoding skips unknown field numbers rather than failing, so a `BoardGraphicShape` decodes as a mostly-empty `Pad` — and the last line then **overwrites the child with a `Pad`-typed `Any`**, discarding the graphic.

The empties match the report:

- `Pad.number` is tag 3; a graphic built by `board_shape()` leaves `net` (also tag 3) unset, so tag 3 is absent from the wire bytes and `number` stays `""` — the **unnamed** pads.
- `Pad.position` is tag 7; `BoardGraphicShape` has no field 7, so `position` stays `None` — read back as **(0,0)**.

Both call sites are inside this one tool, and both paths are affected:

- `update_footprint_item` (`pcb_sync.rs:870`) — the update path that hit U1/J1.
[...]

PR #294, opened 2026-08-18
feat(pcb): repair footprints corrupted by legacy sync

nordic-style:
## Summary

Adds a safe recovery path for boards already corrupted by the legacy schematic-to-PCB sync bug in #244. The new `repair_corrupted_footprints` tool detects the exact anonymous-pad signature, previews a revision-bound repair plan, and restores footprint drawing shapes from the registered libraries.

Refs #244.

## Approach

The original bug converted `BoardGraphicShape` children inside placed footprints into anonymous pads. Prevention exists on `main`, but affected board files remain damaged.

The repair tool:

- runs only against the exact board currently open over KiCad IPC;
- defaults to a non-mutating dry run;
- recognizes both the raw empty-layer signature and KiCad's saved/reloaded 1 nm PTH normalization;
- requires a one-for-one match between phantom pads and missing library drawing shapes;
- requires the legitimate pad-number multiset to match the library exactly;
- preserves placed-footprint identity, placement, symbol path, live pad KIID/net/lock state, and board-specific non-shape children;
- reports a per-reference `repair_merge_failed` diagnostic and continues scanning if the final footprint merge fails;
- applies all requested repairs in one KiCad undo commit;
- reads the board back after apply and verifies that no phantom pads remain and all expected drawing shapes are present.

Ambiguous or unresolvable footprints are reported as conflicts and are not changed.

## Compatibility and safety

This is an additive MCP tool in the existing `pcb_components` toolset; there is no breaking public API or config change.

Apply requires the exact `plan_revision` returned by a current dry run. The tool is live-IPC-only and refuses file fallback, stale plans, library mismatches, unknown references, and partial/ambiguous signatures. The complete mutation is one KiCad undo step, so Ctrl-Z rolls it back.

Tool counts and the public tool directory, README, development guide, plugin metadata, and troubleshooting guide are updated.

## Validation

- [x] `cargo test -p konnect-core tools::pcb_components --locked` — 71 passed
- [x] Signature, preservation, and per-footprint failure-isolation regression tests pass
- [x] Real KiCad IPC checks: four clean boards remained no-ops; two affected boards repaired 20 and 93 footprints atomically; save/reload scans became no-ops and corruption-specific DRC categories disappeared
- [ ] Independent maintainer live-session validation requested in review

## Review checklist

- [x] IPC mutations validate the requested board, apply atomically, verify live readback, and do not leave partial batches.
- [x] Counts and docs are updated per `CONTRIBUTING.md`.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -298,7 +298,7 @@ Source: [`crates/konnect-core/src/observability.rs`](crates/konnect-core/src/obs
 
 ## Tool Routing (Starter Kit + On-Demand Loading)
 
-The server does NOT expose all 203 tools (209 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
+The server does NOT expose all 204 tools (210 total with the 6 meta-tools) in `tools/list` by default — that would cost ~23K tokens of context on every listing. Instead:
 
 - **Startup**: only `STARTER_KIT` toolsets are pre-loaded (see `router/registry.rs::STARTER_KIT`). Currently: `project`, `config`. Combined with the 6 meta-tools, baseline `tools/list` is 20 tools ≈ 2K tokens.
 - **On demand**: the LLM reads `list_toolboxes` → calls `load_toolset(name)` to expose a toolset's tools in subsequent `tools/list` responses. `unload_toolset(name)` prunes them when the task shifts.
@@ -373,9 +373,9 @@ convention for other `kicad-cli`-calling code.
 
 ## Current Stats
 
-- **19 toolsets, 203 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
+- **19 toolsets, 204 tools** + 6 meta-tools (4 routing + 2 observability — see `tool-directory.md`)
 - Baseline `tools/list`: 20 tools / ~2K tokens (starter kit + meta-tools)
-- Full-catalog `tools/list` (all loaded): 209 tools (203 registered + 6 meta) / ~25K tokens
+- Full-catalog `tools/list` (all loaded): 210 tools (204 registered + 6 meta) / ~25K tokens
 - **0 IPC stubs** (all protobuf methods implemented)
 - **0 unimplemented tools**
 - **Specctra DSN/SES are PCB-editor operations**, not `kicad-cli` commands. Konnect
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 Rust binary — that lets Claude and other AI assistants design schematics and PCBs
 through the [Model Context Protocol](https://modelcontextprotocol.io) (MCP).
 
-**203 tools across 19 on-demand toolsets.** Schematic capture, PCB layout and
+**204 tools across 19 on-demand toolsets.** Schematic capture, PCB layout and
 routing, ERC/DRC, design-review audits, JLCPCB part search, reference
 circuits, and a full manufacturing export pipeline — with bundled skills and agents
 that teach Claude KiCAD conventions out of the box.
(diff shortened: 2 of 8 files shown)


mixelpixx:
Deep review done — this is the most careful diff in your series, and the safety question that matters most here is one you answered correctly in the code: the phantom-pad signature requires empty number AND empty/absent net AND (empty pad-stack layers OR a ≤1 nm PTH drill), and `issue_244_signature_distinguishes_real_unnumbered_pads` plants a genuine unnumbered mechanical pad and proves it survives. The exact-count gates (phantom count == missing-graphics count, legitimate pad-number multiset == library multiset), the fresh-revision requirement, live-IPC-only, single undo commit, and the post-apply read-back are all the right calls for a tool that rewrites footprint contents.

Two things before merge:

1. **One robustness gap**: inside the per-footprint scan loop, `merge_clean_footprint_children(...)?` uses `?` where every other failure path in the same function is caught into `diagnostics`. If it ever fails, the whole board's dry-run aborts instead of reporting that one reference and continuing. Given the preceding gates it may be unreachable — but "unreachable given the current gates" is precisely the kind of invariant that stops being true quietly. Please catch it per-reference.

2. **It needs a live session before it lands, and I want to run that myself.** This tool deletes and rewrites footprint children on real boards; your "20 and 93 footprints repaired" runs can't be independently confirmed from the diff. Now that v0.7.0 ships the corruption fix, I can produce a genuinely corrupted board with v0.6.1's binary, run this against it live, and diff the result against the library. That's scheduled for the next round with a KiCad session up.

Related context for your rebase: v0.7.0's release notes point affected users at this PR, and main gained `konnect_sexp::board` (pad/footprint counting helpers) plus `builders::any_is`/`any_type_name` for `Any` discrimination — your detection code may be able to shed some local duplication by reusing them.

nordic-style:
Thanks for the careful review — addressed in `694e736`:

- rebased the PR onto current `main`;
- replaced the final `merge_clean_footprint_children(...)?` propagation with a per-reference `repair_merge_failed` diagnostic and continue;
- added `repair_merge_failure_is_scoped_to_one_footprint`, which makes one malformed candidate fail and proves the next valid footprint is still processed;
- kept the new shared `builders::any_is` helper throughout the IPC `Any` discrimination. I also reviewed `konnect_sexp::board`; those helpers count parsed board S-expressions, while this repair path validates live IPC `FootprintInstance` protobuf children, so they do not replace the local signature/count logic here;
- updated the rebased catalogue total to 204 registered tools.

Local validation is green:

- `cargo test -p konnect-core tools::pcb_components --locked` — 71 passed

I have left the independent live KiCad session explicitly pending for you, as requested.

mixelpixx:
Live verification complete — with a board corrupted by the actual shipped bug, not a synthetic fixture.

Recipe: built the **v0.6.1 release binary** from its tag, stripped every footprint from a fresh `complex_hierarchy` demo copy, opened it in pcbnew 10.0.5, and ran that binary's `update_pcb_from_schematic`. Result: the genuine #244 damage at full scale — **1013 pads (165 real + 848 phantom nameless), 0 graphics** across 68 footprints, live and saved.

Then this branch (merged with current main, zero conflicts), against that board:

- dry run: **all 68 corrupted footprints detected**, per-reference `phantom_pads_removed == expected_drawing_shapes` in every candidate
- apply with the returned revision: `applied: true, repaired_count: 68`, one undo commit
- `save_project`, then audited **the file KiCad wrote**: **165 pads, 0 nameless, 848 drawing shapes, 68 footprints** — matching the pristine original demo board per reference

The robustness fix from the earlier review (`repair_merge_failed` scoped per-reference, commit `694e736`) is confirmed present with its regression test. Merging now. This closes the loop the v0.7.0 release notes promised to users holding boards damaged by v0.4.0–v0.6.1.

Genuinely good work — the adversarial test for real unnumbered mechanical pads is exactly the discipline this repo asks for.

Issue #192, opened 2026-08-14
add_zone and add_copper_pour write legacy net syntax — zone lands on net 0 (write-side counterpart to #142)


anyn99:
On KiCad 10 boards, `add_zone` and `add_copper_pour` write the zone with `(net 0)` regardless of the requested net, leaving the pour electrically orphaned. This is the write-side counterpart of the read-side problem fixed by #142 / reported in #133.

**Environment:** Konnect v0.3.0 (build 2026-08-14), KiCad 10.0.4, Windows 11.

## Repro

Against a copy of a real, fully routed KiCad 10 board that has a GND net:

```
add_zone(board=<board>, net_name="GND", layer="B.Cu", points=[4 pts])
  → {"layer":"B.Cu","net":"GND","net_id":0,"point_count":4}
  → file contains: (zone (net 0) (net_name "GND") …)

add_copper_pour(board=<same board>, net_name="GND", layer="F.Cu", points=[4 pts])
  → file contains: (zone (net 0) (net_name "GND") (layer "F.Cu") …)
```

`net 0` is the unconnected pseudo-net. The file still loads — KiCad is lenient here — but the pour is attached to nothing. Worth noting that #142 points users at `add_copper_pour` as the KiCad-10-safe alternative when `add_net` refuses, and it carries the same defect.

## Side by side with what KiCad 10 writes

Same file, one zone drawn in the GUI and one written by Konnect:

```
; KiCad 10                        ; Konnect
(zone                             (zone (net 0) (net_name "GND") (layer "F.Cu") …
    (net "GND")
    (layers "F.Cu" "B.Cu")
```

Three differences: the net is a **name, not an id**; there is **no `net_name` token**; and the key is **`layers`** (plural), which is also what lets a zone span both copper layers.

## Suggested fix

#142 already landed the hard part in `konnect-sexp`: format detection by shape (child index 1 is a quoted string → KiCad 10, bare atom → legacy) plus `collect_net_keys`. Reusing that detection to choose the *output* form — `(net "GND")` for KiCad 10 boards, the legacy pair for older ones — would fix both tools in one place and keep both formats supported, in line with #142's "read by shape, not version" principle.

## Second, related problem: no-op on an open board

[...]

PR #316, opened 2026-08-23
fix(add_zone): edit the live board over IPC, not just the file

jctrouble:
## The defect

`add_zone` never attempted IPC. It read the `.kicad_pcb`, `rfind(')')`, inserted a `(zone …)` S-expression and wrote the file back — regardless of whether KiCad had the board open. With pcbnew holding the board, that zone is invisible to the user and is discarded by pcbnew's next save: success reported, nothing kept. This bit a real board.

`add_copper_pour` was a second copy of the same handler with the same behaviour (its description at least admitted to being an "S-expression file insert").

An earlier fix bolted `refuse_if_board_open_in_kicad` onto both. That stopped the silent loss, but left the tool refusing outright in the one situation it is most wanted — KiCad open, board on screen — because it had no IPC path to offer instead.

## The fix

Both tools now follow the `attempt_ipc_write` pattern `import_svg_logo` and `add_mounting_hole` already use:

- **Live KiCad** → a `Zone` protobuf through `create_items`, followed by `RefillZones` so the user actually sees copper rather than a bare outline. The result reports `"source": "ipc"` and `zone_id`, the KIID KiCad assigned. The change lands in KiCad's undo stack.
- **Unreachable transport** → the file insert, as before, but the result now reports `"source": "file"` **and an explicit `warning`**: KiCad was not reachable, so if it does have this board open (IPC API switched off, say), use File → Revert before saving there.
- **KiCad answers and refuses** → fail closed, file untouched. That is `attempt_ipc_write`'s existing contract, and it is a behaviour change worth flagging: a reachable KiCad holding a *different* board now refuses rather than editing this board's file. That is the same rule every other board-mutating tool follows, and the alternative is the wrong-board edit its doc comment describes.

## New parameters

On both tools:

| Param | Type | Default | Notes |
|---|---|---|---|
| `name` | string | `""` | Zone name as shown in pcbnew's zone properties |
| `priority` | integer | `0` | Higher wins where zones overlap |
| `pad_connection` | `solid` \| `thermal` \| `none` | `thermal` | KiCad's own default |

Both the IPC path and the file fallback honour them. The file path omits `(name …)` and `(priority 0)` at their defaults, which is what pcbnew writes, and spells thermal as a bare `(connect_pads (clearance …))` for the same reason.

## Consolidation

`add_copper_pour` now delegates to `add_zone`'s implementation instead of duplicating it. The two copies had already drifted — `min_width` defaulted to `0.25` there and `0.2` here — and the #192 net-lookup bug had to be found and fixed in both. **Unified on 0.2**, KiCad's default; this is the one behaviour change to `add_copper_pour` beyond the IPC path. Their JSON schemas are now one function, so they cannot drift apart again. The tool keeps its name and arguments.

`create_items` gains a `create_items_returning` / `create_items_in_returning` sibling so a caller can learn the KIID KiCad assigned. The existing `Result<()>` signatures and the per-item success accounting are unchanged — `create_items_in` now delegates to the returning form and discards the result.

## Testing

`cargo fmt --all -- --check`, `cargo clippy -p konnect-ipc -p konnect-core --all-targets -- -D warnings`, and `cargo test --workspace` all pass. **973 passing / 0 failing / 9 ignored**, up from 961 / 0 / 8 on `main` — 12 new tests that CI runs, plus one new `#[ignore]`d live test.

*(A pre-existing `clippy::byte_char_slices` warning in `crates/konnect-sexp/src/writer.rs:1113` fires under Rust 1.97's clippy on unmodified `main` too. Untouched here — unrelated file.)*

### Offline (run by CI)

- `builders::build_zone` — mm→nm point conversion, layer, net name + code, zone name, priority, connection style, clearance/min-thickness, `ZfmSolid`/`IrmAlways`; that every copper layer name resolves rather than silently becoming `BL_UNDEFINED`; and that the zone goes over the wire **unfilled**, since the fill polygons are KiCad's to compute.
- Handlers — the fallback names itself and warns; `name` / `priority` / `pad_connection` reach the S-expression and the result JSON; the defaults write exactly what pcbnew writes (no `(priority 0)`, no empty `(name)`, bare `connect_pads`); `pad_connection: none` writes the `no` token; an unknown `pad_connection` is refused before anything is written; **a rejecting KiCad leaves the file byte-identical**; and the alias carries the shared defaults (`min_thickness 0.2`) plus the new arguments.
- The existing #192 net-format tests still pass unchanged on both tools.

### Live IPC — run and passing (two rounds)

Josh ran the suite against **KiCad 10.0.4, standalone pcbnew, macOS**, `--test-threads=1`:

**Round 1** (at the previous head, before the fixture fix):

| Test | Result |
|---|---|
| `moving_and_rotating_footprint_preserves_child_geometry` | **PASS** — as far as I can tell the first time anything in `live_kicad_test.rs` has been confirmed to run |
| `adding_a_via_actually_creates_it_on_the_board` (pre-existing) | **FAIL** — `board has no named net to attach a via to` |
| `adding_a_zone_creates_it_on_the_live_board` (new here) | **FAIL** — same precondition |

**Round 2** (fixed fixture, run against a disposable copy per the docs):

| Test | Result |
|---|---|
| `adding_a_via_actually_creates_it_on_the_board` | **PASS** |
| `adding_a_zone_creates_it_on_the_live_board` | **PASS** — including `zone.filled && !zone.filled_polygons.is_empty()`: KiCad computed and holds real fill after `RefillZones`, so the user is looking at copper |
| `moving_and_rotating_footprint_preserves_child_geometry` | **PASS** |

`3 passed; 0 failed` in 1.83 s — full output in the comment below.

Both failures are in the fixture, not the code: `fixtures/live_ipc.kicad_pcb` is the stock EuroCard template whose net table declares only the unnamed net 0, so `get_nets` returns nothing usable and both tests die on their own `expect` before reaching anything under test. **The pre-existing `add_via` test could therefore never have passed against this fixture in any environment** — independent evidence, on top of the CI gap below, that this suite has never actually run.

Fixed here, as an 18-line pure addition to the fixture (no format churn): it gains a `GND` and a `SIG1` net in the legacy table-and-id shape the board already uses, and both tests' coordinates move inside the board. The template does not start at the origin — its outline is (55, 45) to (215, 145) — so the via test's `(40, 40)` was **off the board entirely**, and my zone outline was half off it. The zone now straddles the GND segment deliberately: `add_zone` sets island removal to `IRM_ALWAYS` (KiCad's default), so a pour that reaches nothing else on its own net has its whole fill discarded as islands, and the `filled_polygons` assertion would have failed on a floating zone even once a net existed.

~~**Rerun pending** — the two net-attached tests have not yet been observed passing. That is the outstanding verification on this PR.~~ **Done — round 2 above.** Every code path in this PR has now been observed working against a real KiCad; what remains unautomated is only the CI wiring for the live suite, which was already out of scope.

### Two things worth knowing if you run it

**The socket is per *frame*, not per process.** On macOS with KiCad 10 the *manager* owns the well-known `/tmp/kicad/api.sock` and answers `GetOpenDocuments` with `AS_UNHANDLED`, because it holds no board. A standalone pcbnew registers its own `/tmp/kicad/api-<pid>.sock`, and that is the one to target — the suffix changes every launch, so it cannot be hard-coded. The module docs assumed a single well-known path; corrected.

**Point `KONNECT_LIVE_KICAD_BOARD` at a copy.** These tests mutate the board and call `save_board()`. Run against the tracked fixture, KiCad rewrites it in place — format-upgraded, with `MH1` left where the move test put it — and drops a `.kicad_pro`, `.kicad_prl`, lock file and `.history/` beside it. That happened during this run and `git add -A` swept the lot into a commit, including a lock file recording the developer's hostname and username. All four are now gitignored and the docs say to copy first. `.history/`, `*.lck` and `*.kicad_prl` are blanket-ignored everywhere, fixtures included — those must never be tracked even deliberately.

`.kicad_pro` takes three layers, because generated and intentional ones are both real and are distinguishable by name:

1. **ignored everywhere** by default — usually a machine dropping;
2. **re-opened under `crates/*/tests/fixtures/`** — `create_netclass` and `assign_net_to_class` write to the sibling project file by design, so a fixture exercising them needs one tracked on purpose, and a blanket ignore would make `git add` skip it *in silence*: the same failure mode inverted;
3. **re-closed by name** for `live_ipc.kicad_pro` and `test.kicad_pro` — KiCad writes `<board>.kicad_pro` beside any bare `<board>.kicad_pcb` it opens, `git ls-files '*.kicad_pcb'` returns exactly those two boards, and those exact filenames are therefore always droppings.

So an intentional project fixture just needs a name of its own (`netclass.kicad_pro`, say) and is trackable with no gitignore edit at all. Verified with plain `git check-ignore` and end-to-end `git add -A` probes.

```sh
cp crates/konnect-ipc/tests/fixtures/live_ipc.kicad_pcb /tmp/live/
export KONNECT_LIVE_KICAD_BOARD=/tmp/live/live_ipc.kicad_pcb
export KICAD_API_SOCKET="$(ls -t /tmp/kicad/api-*.sock | head -1)"
cargo test -p konnect-ipc --test live_kicad_test -- --ignored --test-threads=1
```

### A guard for the preconditions

Since no CI job runs the live suite, `the_live_fixture_satisfies_what_the_live_tests_assume` is deliberately **not** `#[ignore]`d. It checks the fixture still has `GND` plus a second named net, is still in its expected net shape, that both tests' coordinates lie inside the Edge.Cuts outline, and that a GND segment crosses the zone region. Verified by mutation — stripping the nets, rewriting the board into the name-only shape, and moving the GND segment off the zone region each fail it. This is the check that would have caught the defect above without a KiCad.

*`live_kicad_test.rs` and `live_kicad_tools.rs` have never been run by CI: `e2e-kicad.yml` installs KiCad but runs only `e2e_kicad` (kicad-cli, no IPC), `conformance_test` and `mock_server_test`, and sets none of the `KICAD_API_SOCKET` / `KONNECT_LIVE_KICAD_*` variables those files require. The Xvfb "CI live-GUI job" their doc comments described does not exist in the repo; that claim is corrected here. Wiring up such a job is out of scope for this PR, but it is why the offline guard is worth having.*

## Docs

`tool-directory.md` and `crates/konnect/assets/skills/kicad-pcb/SKILL.md` updated for the new arguments and the IPC-first behaviour. The SKILL.md note that "there is no priority argument" is now false and has been replaced.


jctrouble:
Live verification, round 2 — **3/3 pass** on the current head (`748a748`), against the fixed fixture, run as the module docs now instruct (disposable copy, per-frame socket):

**Environment:** macOS (Darwin 25.5.0), KiCad 10.0.4, standalone pcbnew (per-pid socket `/tmp/kicad/api-<pid>.sock`), `--test-threads=1`.

```
running 3 tests
test adding_a_via_actually_creates_it_on_the_board ... ok
test adding_a_zone_creates_it_on_the_live_board ... ok
test moving_and_rotating_footprint_preserves_child_geometry ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 1.83s
```

Notes from the run:
- The zone appeared filled on screen and was gone after the test's cleanup — the `filled_polygons` assertion (the refill-timing-dependent one flagged in the PR body) held on real hardware.
- Round 1 at the previous head is what surfaced the fixture defects the latest commits fix (no named nets; the pre-existing via test's coordinates off the board entirely) — details in the PR body's Testing section.

mixelpixx:
Live-verified at merge: `add_zone` on a board open in pcbnew 10.0.5 returned `source: "ipc"` with a real `zone_id`, and after `save_project` the zone is in the file KiCad wrote — B.Cu, GND, the exact uuid the tool reported. (One operational note for anyone scripting this: a save immediately after zone creation can catch KiCad mid-fill and return AS_BUSY; the retry succeeded cleanly.) The gate usage is textbook `attempt_ipc_write`/`BoardWrite`. Thanks @jctrouble — first contribution and it follows every house convention.

PR #322, opened 2026-08-24
fix(library): read footprint and lib-table fields from the parsed tree

pauliuszaleckas:
## Summary

`get_footprint_info` and `parse_lib_table` read their string fields by scanning
the raw file text instead of the S-expression they had already parsed. Two
user-visible consequences:

- An escaped quote truncates the value. KiCad writes a description of
  `0.1" pitch headers` as `(descr "0.1\" pitch headers")`; the scan stopped at
  the escaped quote and returned `0.1\`.
- A field written across a line break is missed entirely, and the entry comes
  back blank — an fp-lib-table entry loses its nickname, URI and type, so its
  libraries stop resolving.

Not linked to an issue. #84 tracks the same root cause elsewhere but explicitly
excludes `parse_lib_table`, whose *indentation* scan was already fixed; this is
the field scraping that remained inside each located block.

## Approach

`extract_sexp_string` looked for the literal `(key "` and read to the next `"`:

```rust
let pat = format!("({} \"", key);
let start = block.find(&pat)? + pat.len();
let end = block[start..].find('"')? + start;
```

It has no notion of escapes, of nesting, or of any whitespace other than one
space. Both call sites now read from the parsed tree — `find_str(tag)` for a
tagged field, and the root node's datum for a footprint's name — so escape
decoding and whitespace tolerance come from the parser. The function had no
callers left and is deleted.

Two `get_footprint_info` hunks are behaviour-preserving and included only for
consistency: `find_all("pad").len()` becomes an explicit direct-child filter
(`find_all` is already direct-children-only), and `!find_all("model").is_empty()`
becomes `find("model").is_some()`, which short-circuits and allocates nothing.

**Why lib-table entries are still located textually.** The obvious move is
`parse_sexp` over the whole file. `parse_lib_table` is infallible by design —
every caller path (`read_flat_lib_table` → `flatten_lib_table`) degrades to an
empty list and a `tracing::warn!` rather than surfacing an error — so a
whole-file parse would let one malformed entry silently discard *every* library
in the table, which for this path means footprint and symbol resolution failing
wholesale. Keeping `find_block_starts` + `find_balanced_block` (both already
string-aware and indentation-agnostic) preserves per-entry tolerance while still
removing the scraping. A malformed entry is now skipped with a warning naming
its offset.

## Compatibility and safety

No public compatibility impact. No tool, argument, schema, config key or path
changes. Read paths only — nothing here writes a file or touches IPC.

`get_footprint_info`'s `name` and `description` response fields can now return a
*correct* value where they previously returned a truncated or empty one; no key
is added, removed or renamed.

## Validation

```
$ cargo fmt --all -- --check
$ cargo clippy --workspace --locked -- -D warnings
```

All four clean, run against this branch on top of `main` (not merely on the
development branch this was cherry-picked from).

The two bug-fix tests were verified to fail against the old scraper before the
fix was restored:

```
parse_lib_table_keeps_an_escaped_quote_in_a_description ... FAILED
parse_lib_table_reads_a_field_split_across_lines ... FAILED
```

`parse_lib_table_skips_only_the_entry_that_does_not_parse` passes both before
and after — it is not a regression test for a fixed bug, but a guard on the
tolerance property against the whole-file-parse refactor described above. Its
comment says so.

Not run: viewer, plugin, packaging and real-KiCad checks. This touches one file
in `konnect-core` and none of those paths.

- [ ] Relevant viewer, plugin, packaging, and real-KiCad checks — not applicable

## Review checklist

- [x] File mutations are atomic and preserve unrelated content — read paths only, nothing written.
- [x] IPC mutations verify the requested board and do not leave partial batches — no IPC.
- [x] If tools were added/removed: counts and docs updated — no tools added or removed.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -1142,6 +1142,12 @@ impl konnect_schematic_editor::library::SymbolLibrarySource for KiCadSymbolSourc
 /// Indentation-agnostic: KiCad's own writers emit tab-indented, CRLF-terminated
 /// tables while this crate's writer uses two spaces, so a fixed literal such as
 /// `"\n  (lib "` silently matches nothing in a real `fp-lib-table`.
+///
+/// Each entry is located textually and then *parsed*, so a field is read from
+/// the tree rather than scraped out of the block: an escaped quote inside a
+/// `descr` no longer truncates the value, and a field KiCad wrote across a line
+/// break is still found. Locating entries textually keeps a malformed one from
+/// discarding the rest of the table.
 fn parse_lib_table(content: &str) -> Vec<serde_json::Value> {
     let mut libs = Vec::new();
     // Each entry: (lib (name "NICK") (type "...") (uri "...") (options "") (descr "..."))
@@ -1150,17 +1156,18 @@ fn parse_lib_table(content: &str) -> Vec<serde_json::Value> {
             continue;
         };
         let block = &content[block_start..block_end];
+        let Ok(entry) = parse_sexp(block) else {
+            tracing::warn!("lib-table entry at byte {block_start} does not parse — skipped");
+            continue;
+        };
 
-        let nickname = extract_sexp_string(block, "name").unwrap_or_default();
-        let uri = extract_sexp_string(block, "uri").unwrap_or_default();
-        let lib_type = extract_sexp_string(block, "type").unwrap_or_default();
-        let descr = extract_sexp_string(block, "descr").unwrap_or_default();
+        let field = |tag: &str| entry.find_str(tag).unwrap_or_default().to_string();
 
         libs.push(json!({
-            "nickname": nickname,
-            "uri": uri,
-            "type": lib_type,
-            "description": descr
+            "nickname": field("name"),
+            "uri": field("uri"),
+            "type": field("type"),
+            "description": field("descr")
         }));
     }
     libs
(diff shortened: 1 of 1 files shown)


mixelpixx:
Holding merge on the two Windows-only CI failures: `flatten_lib_table_follows_nested_table_entries` and `list_footprint_libraries_expands_a_nested_table_of_env_var_uris`. The core escape/whitespace fix reviews correct — it's specifically the nested Table-type / env-var URI expansion path that regressed on Windows. Everything else is MERGE-ready once those two are green.

pauliuszaleckas:
Both Windows failures were the test fixture, not the fix.

`kicad_style_table` interpolated a tempdir path into the table raw. On Windows that path holds backslashes, and the nested table is named `template-fp-lib-table`, so the file text contained `…\.tmpXXXX\template-fp-lib-table` — the parser reads `\t` as a TAB, exactly as KiCad does, so the URI pointed at a file that does not exist and the `(type "Table")` entry expanded to nothing. Not reachable through Konnect on a real table: `register_*_library` writes URIs through `portable_uri` (forward slashes) and `quote_lib_table_string` (escaped), and KiCad escapes them too.

The helper now escapes what it interpolates via `escape_library_string`, the same way `quote_lib_table_string` writes it, so the fixture is what KiCad would have written on either platform.

Verified rather than assumed: pointing `TMPDIR` at a directory containing `\t` reproduces both failures verbatim on Linux (`nested table not followed: []`, `nested table not expanded: {"count":0,"libraries":[]}`) and both pass with the fix. Added `parse_lib_table_reads_an_escaped_windows_uri` as a platform-independent guard on the property the helper now relies on.

Full gate green: 602 lib/tests, doctests, clippy `-D warnings`, fmt.

PR #323, opened 2026-08-24
refactor(sch): give the connectivity tools one shared index

pauliuszaleckas:
## Summary

Asking "is this point connected" had four answers in the tree, and they disagreed on real sheets:

| Where | Tolerance | Counts as connected |
|---|---|---|
| `sch_analysis::find_orphan_items` | argument, default 0.05 | wire end, wire interior, pin, label, sheet pin, junction, no-connect |
| `sch_batch::validate_wire_connections` | argument, default 0.01 | wire end, wire interior, pin, label |
| `sch_batch::validate_component_connections` | hardcoded 0.01 | wire end, junction on a wire, net-graph reachability |
| `build_net_graph` | hardcoded 0.01, whatever the caller asked | wires, labels, junctions — no sheet pins |

Reproduced against `main` before writing any tests:

```
wire ending on a sheet pin: find_orphan_items 0 orphans | validate_wire_connections 1 floating end
two pins placed on each other: find_orphan_items 0 orphans | validate_component_connections 2 unconnected
```

No linked issue — the tracker has nothing on the divergence itself.

## Approach

One `ConnectivityIndex` (`crates/konnect-core/src/tools/sch_connectivity.rs`) built once per tree, holding wires, wire endpoints, labels, unit-aware pins with the instance that placed them, sheet pins, junctions and no-connects **under a single tolerance**. Two predicates carry the policy the tools each used to invent:

- `terminates_wire_end` — a pin, label, sheet pin, junction, no-connect, second wire end, or a wire interior it lands mid-span on.
- `attaches_pin` — a wire end, a label, a sheet pin, a second stacked pin, or a junction dot on a wire.

`NetGraph` moved into the same module and `build_net_graph` became `net_graph_for(tree, wires, labels)`, which reads junctions *and sheet pins* off the tree itself — so all ten call sites gained sheet-pin awareness in a one-line swap. One `seed_net_graph` is the only definition of the graph, and its attach tolerance is a property of the `WireIndex` it is handed rather than a literal in the function body. Every production caller still passes `COINCIDENT_TOLERANCE`, so that part is structure, not a behaviour change.

`placed_pins_by_reference` now yields the whole `SymbolInstance`, so a caller reporting a pin reads the value from the instance that placed it rather than from a reference-keyed map — which collapses on a pre-annotation sheet where every part is `R?`.

## What changes for callers

All in the direction of fewer false faults:

- `validate_wire_connections` — sheet pins, junctions and no-connects terminate a wire end; its pins are unit-aware.
- `validate_component_connections` — stacked pins, a label at a pin, and sheet pins all connect; its pins are unit-aware. Its net-graph reachability fallback is gone (see the third trap below), so an unwired power symbol is now reported.
- `find_orphan_items` — a pin meeting a sheet pin now counts as connected. The only behaviour change it takes.

**Public tolerance defaults are unchanged and no tool schema or JSON key moved.** `find_orphan_items` keeps its `tolerance` argument defaulting to 0.05, `validate_wire_connections` keeps its 0.01, and `validate_component_connections` keeps its hardcoded 0.01 — now the named `COINCIDENT_TOLERANCE` rather than four scattered literals. "One tolerance" is one *per index*, not one global; collapsing the defaults is a separate, user-visible change.

## Three traps worth knowing about

All three were caught in review after an earlier version got them wrong, and all three are now pinned by tests:

- **A power symbol's pseudo-label sits on the symbol's own pin.** Index it as a label and every power pin sees a label at its own position and reports itself attached — which hides an unwired `GND`, one of the commonest real mistakes on a sheet, from `find_orphan_items` entirely. Power symbols are excluded from the index's *label points*; the graph still gets every label, because naming is what it needs them for.
- **A hierarchical sheet pin is a pin.** It is registered in the graph but never attached mid-span: a wire that *ends* on one connects, a wire that merely passes through its coordinates does not, and a junction dot bridges the difference — the same #104 rule `attaches_pin` states for symbol pins. Attaching mid-span would merge a wire routed along a sheet edge with whatever net that pin carries, inventing a short KiCad does not see.
- **The reachability fallback had to go, not stay.** An earlier version kept `graph.net_at(…)` after the coincidence test, on the theory that a fallback can only *remove* false positives. It cannot fire legitimately: every point the graph can name is a wire endpoint, label, junction or sheet pin, and `attaches_pin` covers all four *at tolerance* rather than at exact `pt_key` equality — so `net_at` returning `Some` implies `attaches_pin` already returned true. The single case it did fire on was the power-symbol pseudo-label the index deliberately excludes, so an unwired `GND` came back clean from `validate_component_connections` while `find_orphan_items` called it an orphan. That is this PR'"'"'s own divergence, surviving inside the fix for it. Removing it makes the two agree and costs nothing else.

### One invariant worth stating

Coincidence is a **superset** of the graph, and stays that way. A point is coincident within the index'"'"'s tolerance; the graph joins nodes on exact `pt_key` equality plus the attach step. A wire ending 5 µm off a sheet pin is therefore *terminated* for the validators yet does not reach that pin'"'"'s net. That is the safe direction: leniency suppresses a false "floating end" from a rounding artefact, while strictness stops the graph inventing a connection KiCad'"'"'s own netlister would not make. KiCad snaps to grid, so exact agreement is the normal case. Documented at `pt_key` and in the module header.

## Overlap with open PRs

Both touch handlers this PR rewrites, and both are on the **unit-blind pin resolution** axis (#182) rather than the definition of "connected" — a different axis, deliberately not addressed here:

- **#272 (`fix(schematic): resolve pins through their placed unit`)** — its two `sch_batch` hunks are each one `extract_lib_pins` → `extract_lib_pins_for_unit` line. This PR reaches the same result by a different route, because the index sources pins from `placed_pins_by_reference`, unit-aware since #35/#143. **Its `design_review` and `sch_export` hunks are untouched here and still needed.** It is currently `CONFLICTING` against main independently of this PR.
- **#267 (`fix(schematic): respect intentional unconnected pins`)** — its exemption work on `validate_component_connections` (`electrical_type == "no_connect"`, `ignore_power_pins`, a reported `pin_type`) is policy *over* the connectivity answer rather than the answer itself. Deliberately **not** included here; it stays that PR's to land, and rebasing it onto this index should be mechanical.

Happy to rebase this behind either if maintainers would rather land those first.

## Not in scope

`design_review::find_net_at_point` — the fourth row of the table — resolves nets by a 0.5 mm text scan for `(label "`, follows no wires and knows no power symbols. It never used the net graph, so it is untouched here; moving the audits onto this index is its own change, and the index now exists for it.

A fifth answer turned up while reviewing this one: `sch_export::handle_fix_connectivity` builds its own `snap_targets` at its own 0.01 with its own inline T-junction test, and no sheet pins, junctions or no-connects. It is the worst-placed duplicate because it is the one that *writes* — a wire ending on a sheet pin reads as a near miss and gets snapped. Migrating it changes what the tool edits, so it needs its own PR.

## Testing

Full local gate green: `cargo fmt --check`, `cargo clippy --workspace --locked --all-targets -- -D warnings`, `cargo test --workspace --locked --lib --tests` (609 in `konnect-core`), `cargo test --workspace --locked --doc`. `schematic-viewer` untouched; no MCP tools added or removed, so no count sync needed.

Eleven new tests in `sch_connectivity::agreement_tests`, most asking two tools the same question:

- `a_sheet_pin_terminates_a_wire_end_for_every_tool`
- `stacked_pins_are_connected_for_every_tool`
- `an_isolated_pin_is_unconnected_for_every_tool` — the agreement has to hold in the direction that still reports a fault, or the unification is just a blanket "everything is connected"
- `an_unwired_power_symbol_is_still_an_orphan`
- `an_unwired_power_symbol_is_unconnected_for_every_tool`
- `a_wired_power_symbol_is_connected_for_every_tool`
- `unannotated_components_keep_their_own_values`
- `a_wire_ending_on_a_sheet_pin_joins_its_net`
- `a_sheet_pin_mid_wire_joins_its_net_only_through_a_junction`
- `the_graph_attaches_at_the_index_tolerance`
- `an_unconnected_pin_still_reports_its_component_value`

### DIFF
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -19,6 +19,7 @@ pub mod sch_analysis;
 pub mod sch_batch;
 pub mod sch_bus;
 pub mod sch_components;
+pub(crate) mod sch_connectivity;
 pub mod sch_export;
 pub mod sch_hierarchy;
 pub mod sch_wiring;
@@ -421,12 +422,16 @@ pub(crate) fn placed_pins(
         .collect()
 }
 
-/// [`placed_pins`], grouped under the reference designator that owns each
-/// placed unit, for callers that report pins by name rather than position.
+/// [`placed_pins`], grouped under the instance that placed each unit, for
+/// callers that report pins by name rather than position. The whole instance
+/// is returned because a caller reporting a pin usually wants its reference
+/// *and* something else about the component — value, uuid, unit — and a
+/// reference alone collapses on a pre-annotation sheet where every part is
+/// `R?`.
 pub(crate) fn placed_pins_by_reference(
     tree: &konnect_sexp::SexpNode,
 ) -> Vec<(
-    String,
+    konnect_sexp::schematic::SymbolInstance,
     Vec<(
         konnect_sexp::schematic::LibPin,
         konnect_sexp::geometry::PinTransform,
@@ -452,7 +457,7 @@ pub(crate) fn placed_pins_by_reference(
             .into_iter()
             .map(|p| (p, t))
             .collect();
-        by_reference.push((inst.reference, pins));
+        by_reference.push((inst, pins));
     }
     by_reference
 }
--- a/crates/konnect-core/src/tools/sch_analysis.rs
+++ b/crates/konnect-core/src/tools/sch_analysis.rs
@@ -1,17 +1,18 @@
 //! `sch_analysis` toolset — net connectivity, pin queries, trace paths, overlap/orphan detection.
 //!
-//! All operations are read-only S-expression analysis.
-//! Net graph uses union-find (O(W+L+P)), matching net_analysis.py.
+//! All operations are read-only S-expression analysis. Connectivity — the net
+//! graph and what counts as attached at a point — lives in `sch_connectivity`.
 
 use crate::mcp::protocol::CallToolResult;
 use crate::tool;
+use crate::tools::sch_connectivity::{net_graph_for, pt_key, ConnectivityIndex};
 use crate::tools::{get_path, opt_f64, require_f64, require_str, ToolContext, ToolDef};
 use konnect_schematic_editor as cse;
 use konnect_sexp::{
     geometry::{point_on_segment, points_coincident},
     schematic::{
-        extract_all_net_labels, extract_junctions, extract_labels, extract_symbol_instances,
-        extract_wires, find_lib_symbol, read_schematic, Label, Wire,
+        extract_all_net_labels, extract_symbol_instances, extract_wires, find_lib_symbol,
+        read_schematic, LabelKind, Wire,
     },
 };
 use serde_json::json;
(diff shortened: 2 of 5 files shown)


pauliuszaleckas:
Pushed a revision. A review pass raised four points; three were non-blocking notes, one had a real bug behind it.

**The reachability fallback is gone.** Keeping `graph.net_at(…)` after the coincidence test left `validate_component_connections` strictly more permissive than `find_orphan_items`, and the agreement tests only covered cases where the two *must* agree. Probing that direction found a live disagreement: an unwired `power:GND` came back `unconnected_count: 0` from the validator while `find_orphan_items` reported it as an orphan.

The cause is the power-symbol trap this PR already guards against, one layer down — the pseudo-label is excluded from the index's label points, but the *graph* still holds it, so the fallback answered "connected" for a power symbol attached to nothing. Checking when it could fire legitimately: every point the graph can name is a wire endpoint, label, junction or sheet pin, and `attaches_pin` covers all four at tolerance rather than at exact `pt_key` equality, so `net_at` returning `Some` implies `attaches_pin` already returned true. It was redundant-or-wrong, never load-bearing. Removed; the two tools now agree, with `an_unwired_power_symbol_is_unconnected_for_every_tool` and `a_wired_power_symbol_is_connected_for_every_tool` pinning both directions.

Knock-on: `ConnectivityIndex::net_graph()` then had no production caller, so it and the two `Vec` fields backing it are deleted. `seed_net_graph` remains the single definition, reached through `net_graph_for`.

**Corrected an overclaim.** The earlier commit message and PR body said the graph "attaches at the caller's tolerance instead of a hardcoded 0.01". Only the index path honoured a caller tolerance, and that path is now gone — every production caller passes `COINCIDENT_TOLERANCE`. The accurate statement, now in both: the attach tolerance is a property of the `WireIndex` the seeder is handed rather than a literal in the function body, which is structure, not a behaviour change.

**Documented an invariant.** Coincidence is a superset of the graph, not equal to it: a wire ending 5 µm off a sheet pin is terminated for the validators yet does not reach that pin's net. That is the safe direction and it is deliberate — leniency suppresses a false floating end from a rounding artefact, while strictness stops the graph inventing a connection KiCad's netlister would not make. Noted at `pt_key` and in the module header.

Not addressed, deliberately: `NetGraph::net_at` cloning `point_nets` per call is pre-existing, and this change does not make it hotter — the one call site it added is gone entirely now.

Gate green on the rebuilt branch: fmt, clippy `--all-targets -D warnings`, 609 tests in `konnect-core`, doctests.

Issue #328, opened 2026-08-25
The shared connectivity model is bus-blind: wire ends on bus entries and labels on buses are reported floating


triglav-modular:
The shared connectivity model (#323) extracts wires, labels, pins, sheet pins, junctions and no-connects — **and not buses or bus entries**. On a sheet that uses buses, the consolidated consumers now agree with each other on the wrong answer: a wire terminated on a bus entry and a label naming a bus are both reported floating.

**Version:** v0.8.0 · KiCad 10.0.5 · macOS

### Repro — Konnect tools only

```python
create_project(name="probe", path=TMP)
add_hierarchical_sheet(schematic=root, sheet_file="child.kicad_sch", sheet_name="child", x=100, y=100)
add_bus(schematic=child, x1=100.33, y1=100.33, x2=150.11, y2=100.33)
add_bus_entry(schematic=child, x=120.65, y=100.33, direction="down_right")
add_wire(schematic=child, x1=123.19, y1=102.87, x2=140.97, y2=102.87)   # from the entry's free corner
add_schematic_net_label(schematic=child, net="D[0..3]", x=110.49, y=100.33)  # names the bus
add_schematic_net_label(schematic=child, net="D0",      x=140.97, y=102.87)  # terminates the far end
```

Everything on this sheet is connected: wire → bus entry → bus, bus named, far end labeled.

### What the validators say vs what KiCad says

| point | `validate_wire_connections` / `find_orphan_items` | `kicad-cli sch erc` |
|---|---|---|
| wire end at (123.19, 102.87) — on the bus entry | **floating** / **dangling_wire_end** | no complaint |
| label `D[0..3]` at (110.49, 100.33) — on the bus | **floating_label** | no complaint |

```
validate_wire_connections -> {"floating_count": 1, "floating_endpoints": [{"x": 123.19, "y": 102.87, ...}], "valid": false}
find_orphan_items         -> {"orphan_count": 2, "orphans": [
                                {"type": "dangling_wire_end", "x": 123.19, "y": 102.87, ...},
                                {"type": "floating_label", "net": "D[0..3]", "x": 110.49, "y": 100.33}]}
```

KiCad's ERC on the same file reports neither (its only finding is a `label_dangling` for `D0`, a separate matter).

### Why it matters

[...]

Issue #315, opened 2026-08-23
`move_connected` leaves every wire behind and reports success


komar3456:
# `move_connected` leaves every wire behind and reports success

**Version:** Konnect v0.7.0 (`konnect-pcm-v0.7.0-windows.zip`, installed via PCM)
**KiCad:** 10.0.4 · **OS:** Windows 11 · **Transport:** stdio

`move_connected` is documented as "Move a symbol and stretch/shrink connected wire
stubs to preserve connections". In v0.7.0 it moves the symbol and leaves the wiring
exactly where it was, so every connection the symbol had is silently broken. The
call returns `{"moved": ...}` with no warning, and nothing in the response
distinguishes this from a successful drag.

## Reproduction

Konnect tools only — the wiring is created by `connect_to_net`, so nothing here
depends on hand-placed coordinates.

```jsonc
create_schematic          { "path": "stub.kicad_sch" }
add_schematic_component   { "schematic": "stub.kicad_sch", "lib_id": "Device:R",
                            "reference": "R1", "x": 100.33, "y": 100.33 }
connect_to_net            { "schematic": "stub.kicad_sch", "reference": "R1",
                            "pin_number": "1", "net": "VBUS" }
move_connected            { "schematic": "stub.kicad_sch", "reference": "R1",
                            "x": 111.76, "y": 100.33 }
```

`connect_to_net` writes the stub at pin 1 exactly: wire `(100.33, 96.52) → (100.33, 93.98)`
with the `VBUS` label at the far end.

## Observed

`move_connected` returns:

```json
{"moved":"R1","x":111.76,"y":100.33}
```

In the file, the symbol moved and the stub did not:

```
(symbol (lib_id "Device:R") (at 111.76 100.33 0) ...   ← moved
(wire (pts (xy 100.33 96.52) (xy 100.33 93.98)))       ← unchanged
(label "VBUS" (at 100.33 93.98 0) ...)                 ← unchanged
```

Pin 1 is now at `(111.76, 96.52)`; the stub still starts at `(100.33, 96.52)`.

`kicad-cli sch erc` on the same file, before and after the move:

| | Errors | Warnings | Messages |
|---|---|---|---|
| before `move_connected` | 1 | 1 | `pin_not_connected` (pin 2, expected — nothing attached), `isolated_pin_label` |
| after `move_connected` | 3 | 1 | `pin_not_connected` ×2, `label_dangling`, `unconnected_wire_endpoint` |

[...]

Issue #329, opened 2026-08-25
add_bus_entry's response swaps bus_side and wire_side


triglav-modular:
`add_bus_entry`'s response swaps its two labels: the corner that touches the bus is reported as `wire_side`, and the free corner meant for a wire is reported as `bus_side`.

**Version:** v0.8.0 · KiCad 10.0.5 · macOS

### Repro

A horizontal bus along y=100.33, then an entry placed on it:

```python
add_bus(schematic=child, x1=100.33, y1=100.33, x2=150.11, y2=100.33)
add_bus_entry(schematic=child, x=120.65, y=100.33, direction="down_right")
```

returns

```json
{"added": "bus_entry",
 "bus_side":  {"x": 123.19, "y": 102.87},
 "wire_side": {"x": 120.65, "y": 100.33}}
```

### Why that is backwards

The written element is `(bus_entry (at 120.65 100.33) (size 2.54 2.54))`. The `at` corner (120.65, 100.33) lies **on the bus** — y matches the bus exactly. The far corner at `at + size` = (123.19, 102.87) hangs off it, which is where a wire attaches. So the labels are exactly inverted.

### Impact

Anyone wiring programmatically follows `wire_side` to place the connecting wire — and lands it on the bus itself, where it does nothing (a wire touching a bus without an entry is not connected). The file written is fine; only the response misleads. Found while building the repro for the bus-connectivity report, where the wire had to be placed at the coordinate labeled `bus_side` for the sheet to be correct.

mixelpixx:
Fixed in `1434c7e` (+ fmt follow-up `0a3d3e3`), on main now. Verified with your exact repro through the real binary: the response is

```json
{"added": "bus_entry",
 "bus_side":  {"x": 120.65, "y": 100.33},
 "wire_side": {"x": 123.19, "y": 102.87}}
```

— the on-bus corner is `bus_side`, and following `wire_side` now lands the wire where it actually connects.

[...]

PR #330, opened 2026-08-25
fix(sch): re-judge junction dots after a move

### LINKED ISSUE #120: Component and label moves don't add or prune junction dots

triglav-modular:
## Summary

Moving a component off a wire left its junction dot stranded; moving one onto a wire left the pin unconnected with no dot at all. The stranded case is the dangerous one — at a point where two wires cross, a leftover dot shorts them, which is a connection the user never asked for.

Scope is the **junction model** for the two component-move tools, not #315's wire-carrying move. Per your #315 comment that #120 is its prerequisite, this is that prerequisite and deliberately not the feature itself.

Closes #120

## Approach

`reconcile_junctions_at` re-applies the judgement `prune_orphaned_junctions` already encodes, driven by **pins** rather than removed wires, because a move changes no wires. Candidate points are the symmetric difference of pin sets before/after — nothing outside the moved geometry is considered, so a dot placed anywhere else cannot be touched. This composes with #273's multi-unit move: every unit's pins are in the diff, no extra code.

All connectivity answers come from the shared `ConnectivityIndex` — after #323 consolidated four disagreeing answers, this must not become a fifth. That caught two cases a hand-rolled check missed, both now tested: a **hierarchical sheet pin** justifies a dot exactly as a symbol pin does, and a **no-connect** forbids adding one. The index gains `wires_at()`, the count neither existing boolean could answer.

Guards, each with a regression test:

- **Prune** only a dot left justifying nothing — no pin, no sheet pin, fewer than two wires. Never when `lib_symbols` fails to resolve, which is indistinguishable from "no pin here" only if you guess.
- **Add** only where a pin lands mid-span on exactly one wire. An earlier revision added wherever the pin was justified, which **merged two nets** when it landed on a crossing; caught in self-review.
- **Buses are out of jurisdiction** — KiCad has bus junctions, invisible to any wire count, so a candidate point on a bus line is neither pruned nor added at. See #328; the guard collapses into the index once it learns buses.

Alternative considered and rejected: reusing `prune_orphaned_junctions` directly. Its candidate set comes from removed wires, and a move removes none.

## Compatibility and safety

Two new response fields on `move_schematic_component` and `bulk_move_schematic_components`: `junctions_added_count` and `junctions_pruned_count`, matching the `*_count` convention and the existing `junctions_pruned_count` on `batch_delete_schematic_wire`. Additive only — no field renamed or removed, no schema argument changed, no tool added or removed, so `tool_count` and the directory counts are untouched. Both tool descriptions now document the behaviour.

File mutation: the reconcile writes only when it changed something, through `write_atomic_if_unchanged` against the content it read, so a concurrent edit aborts rather than clobbers. On any parse failure it returns the content untouched.

Known scope limits, stated rather than hidden: `move_labels_by_offset` (named in #120) is **not** wired — it edits label anchors only, which neither justify nor need junctions; correct me if you disagree. `rotate_schematic_component` and the delete paths have the same defect and are left for a separate PR.

**Open question for you:** #120 closes by noting eeschema splits wires rather than leaving mid-segment junctions, and that settling split-vs-junction is "a good reason to settle that question before building the prune". This judges dots against the current file shape. Say the word and I'll park it until that's decided.

## Validation

- [x] `cargo test --workspace --locked --lib --tests` — 755 pass, 1 ignored

7 new tests: prune-when-vacated, keep-a-real-T, crossing-does-not-merge-nets, sheet-pin-keeps, no-connect-forbids, bus-tee-abstains, and an end-to-end `bulk_move` prune.

End-to-end against KiCad 10.0.5 on both wired paths, rebased onto current main: single move reports `junctions_pruned_count: 1` and leaves no dot; moving back reports `junctions_added_count: 1` at `(120.65 139.7)` — rounded, since `format_junction` interpolates raw f64 and pin arithmetic yields `139.70000000000002`; `kicad-cli sch export netlist` exits 0 on the result.

Not run: the `#[ignore]`d live-KiCAD tests (need a running editor); this change is closed-board only and touches no IPC path.

### DIFF
--- a/crates/konnect-core/src/tools/sch_batch.rs
+++ b/crates/konnect-core/src/tools/sch_batch.rs
@@ -139,7 +139,10 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "bulk_move_schematic_components",
-            "Move multiple components by a uniform dx/dy offset in a single atomic file write.",
+            "Move multiple components by a uniform dx/dy offset in a single atomic file \
+             write. Junction dots are re-judged where the pins moved: a dot the pins \
+             leave unjustified is removed and a pin landing mid-span on a wire gains \
+             one, reported as junctions_pruned_count and junctions_added_count.",
             json!({
                 "type": "object",
                 "properties": {
@@ -854,13 +857,53 @@ async fn handle_bulk_move(
         }
     }
 
+    // Pin positions before the shift, so a dot the pins vacate can be re-judged
+    // and a pin landing mid-span gets one (#120). A move changes no wires.
+    const TOL: f64 = 0.01;
+    let pins_of = |src: &str| -> Vec<(f64, f64)> {
+        konnect_sexp::parse_sexp(src)
+            .ok()
+            .map(|t| crate::tools::all_pin_endpoints(&t))
+            .unwrap_or_default()
+    };
+    // No wires means nothing can be justified and nothing can be landed on, so
+    // the whole pass — including two full symbol/lib_symbols walks — is skipped.
+    let has_wires = expected.contains("(wire");
+    let before_pins = if has_wires {
+        pins_of(&expected)
+    } else {
+        Vec::new()
+    };
+
     let new_content = apply_edits(content, edits);
+
+    let after_pins = if has_wires {
+        pins_of(&new_content)
+    } else {
+        Vec::new()
+    };
+    let differs = |a: &[(f64, f64)], b: &[(f64, f64)]| -> Vec<(f64, f64)> {
+        a.iter()
+            .copied()
+            .filter(|&(x, y)| {
+                !b.iter()
+                    .any(|&(ox, oy)| konnect_sexp::geometry::points_coincident(x, y, ox, oy, TOL))
+            })
+            .collect()
+    };
+    let mut points = differs(&before_pins, &after_pins);
+    points.extend(differs(&after_pins, &before_pins));
+    let (new_content, junctions_added, junctions_pruned) =
+        crate::tools::sch_wiring::reconcile_junctions_at(new_content, &points);
+
     write_atomic_if_unchanged(&sch_path, &expected, &new_content)?;
 
     Ok(CallToolResult::json(&json!({
         "moved_count": moved.len(),
         "moved": moved,
         "dx": dx, "dy": dy,
+        "junctions_added_count": junctions_added,
+        "junctions_pruned_count": junctions_pruned,
         "errors": errors
     })))
 }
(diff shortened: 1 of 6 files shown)


mixelpixx:
Verified and fixed up on your branch (rebased onto today's main — your commit needed no re-expression; writing it on top of the fresh #273 merge within hours of it landing made this the cleanest review of the day). Merging once CI is green.

**Verified before touching anything:** the bug is real on main — the move handlers had no junction pass at all, and `prune_orphaned_junctions`' candidates come only from removed wires, so a move could neither gain nor lose a dot. Your reconcile design holds up well under scrutiny: symmetric before/after pin diffing, the crossing-vs-tee distinction (adding a dot at a crossing would merge nets — the exact silent-connectivity change #120 exists to prevent), the no-connect guard, the bus-jurisdiction guard, and everything answered through the shared `ConnectivityIndex`. Multi-unit moves are covered by construction since `all_pin_endpoints` sweeps the whole sheet. The fixture is genuine eeschema output — thank you for that; it's the house rule and it keeps proving itself.

**One fixup commit (`3646941`) on top:**
- **Reverted the `junctions_added` → `junctions_added_count` rename** in `add_schematic_component`'s response. It was out of scope for this PR, breaking for existing consumers, a misnomer (the field holds an array of `{x,y}`, not a count), and left `add_power_symbol` reporting the same data under the old name. Your two new tools keep their `*_count` fields — those are genuinely counts and genuinely new.
- **Dedup before counting**: `pruned` was `ranges.len()` before the dedup, and `to_add` was never deduped — two coincident pin endpoints vacating one spot overcounted the response, and the add branch wrote the same dot twice. Both now dedup first; a test hands the reconciler the same candidate twice and asserts one dot, honest counts.
- **The add path had no positive test** (the PR body's test list mentions one more test than the diff contains — the sheet-pin case; the `has_sheet_pin` keep-branch is currently untested, noted here rather than blocking). Added: strip R1's dot from your fixture, reconcile, and watch it come back with the float-noise rounding pinned.
- **Convergence bonus**: your `extract_buses` + `geometry::point_on_segment` helpers replaced the local copies I'd written hours earlier for #329's bus-entry side adjudication — "is this point on a bus" now has one answer everywhere. Good helpers travel.

Full gate green at 1146 tests. On your open question about eeschema's split-vs-junction convention: judging dots against the file as it stands is the right call for now — that stays the house convention until #328's bus-aware index work forces the larger conversation.

Issue #337, opened 2026-08-26
Implement a native Rust/IPC Freerouting DSN/SES bridge


neusse:
## Problem

Konnect intentionally removed the old `autoroute` tool in #253 because it advertised a workflow that could never succeed. The cleanup was correct, but Konnect still has no supported end-to-end path from a live KiCad board through Freerouting and back.

This should remain a new implementation issue rather than reopening #253. That issue established the engine/plugin distinction and removed a false public capability; this issue proposes the actual bridge.

## Important discovery: Freerouting already provides the engine-side MCP server

Freerouting 2.3.0 includes a native Java MCP server inside the Freerouting JAR. In local/offline mode it runs headlessly over stdio and exposes the complete routing-job state machine:

`create_session -> enqueue_job -> upload_job_input_from_local_file -> update_job_settings -> start_job -> get_job_details -> download_job_output_to_local_file`

Official documentation: [Freerouting v2.3.0 MCP guide](https://github.com/freerouting/freerouting/blob/v2.3.0/docs/API/MCP.md).

This removes the need for Konnect to implement another error-prone Freerouting process protocol, job/session manager, settings transport, progress poller, log/cancellation layer, or artifact upload/download mechanism. Local mode needs neither the Node public-API bridge nor cloud access.

The native Freerouting MCP is an **engine interface**, not a KiCad bridge. It accepts an already-created Specctra DSN and returns a Specctra SES. It does not:

- bind the job to the exact live KiCad document and revision;
- convert a KiCad IPC board snapshot into DSN;
- parse and validate SES against the exported board;
- apply tracks, arcs, and vias through a KiCad transaction; or
- provide rollback, read-back, candidate-save, or DRC evidence.

Those remain Konnect's responsibility. The existing Freerouting KiCad ActionPlugin proves the workflow using legacy Python/SWIG `pcbnew.ExportSpecctraDSN` and `pcbnew.ImportSpecctraSES`, but it is suitable only as a parity oracle or manual fallback. It is not the production architecture for a Rust/IPC bridge and is not forward-compatible with KiCad 11's removal of SWIG.

## Revised feasibility finding

A native bridge is feasible against stock KiCad 10.0.5 without Python, SWIG, GUI automation, or a modified KiCad build:

1. bind and read the exact live board through KiCad IPC;
2. generate a deterministic Specctra DSN in Rust;
3. delegate the routing job to the discovered local Freerouting JAR through its native MCP server;
4. parse the returned SES in Rust; and
[...]

PR #338, opened 2026-08-26
feat(pcb): export revision-bound Specctra DSN

neusse:
## User-visible outcome

Adds `export_specctra_dsn` to the `pcb_export` toolset. A supported board that is open in KiCad PCB Editor can now produce a deterministic Specctra DSN and a revision-bound reverse manifest without Python, SWIG, GUI automation, or direct board-file editing.

Part of #337. This is the first adapter in the agreed three-PR series; it does not advertise end-to-end autorouting or SES import.

## Design

- Proves the requested board through `GetOpenDocuments`.
- Captures the exact live editor state through `SaveDocumentToString` before and after querying KiCad's effective netclass values.
- Refuses the export if those snapshots differ.
- Hashes the exact IPC snapshot and records board, layer, footprint, pad, net, class, transform, and padstack mappings in a deterministic JSON manifest.
- Uses `topola_specctra` for the DSN AST/parser/serializer while retaining Konnect-owned KiCad lowering and fail-closed validation.
- Writes both artifacts with create-new atomic writes; existing targets are structured conflicts and a failed second write removes only the DSN created by that call.

The deliberately narrow first profile supports two copper layers, front-side SMD or plated-through footprints, circle/rectangle pads, one closed straight-line outline, and no existing routing, zones, custom DRC rules, local clearance overrides, pad offsets, or removed-unused-layer pads. Unsupported input is refused rather than approximated.

This adds one public MCP tool and updates the synchronized catalogue from 210 to 211 registered tools (217 including meta-tools). No existing tool, schema, config key, path, or behavior changes. There is no persisted migration.

The manifest schema starts at version 1 so the SES importer can reject incompatible artifacts explicitly in PR 2.

## Tests

Passed locally:

- `FREEROUTING_JAR=/path/to/freerouting-2.3.0.jar cargo test -p konnect-core freerouting_accepts_exported_fixture -- --ignored`

The opt-in engine parity test ran against the installed Freerouting 2.3.0 JAR. Freerouting loaded the generated DSN, completed routing, exited successfully, and produced a non-empty SES. IPC envelope and value conversion are covered with a mock KiCad server. A live-GUI test is intentionally not part of CI because it requires an open disposable board and KiCad API socket.

## Risk and rollback

The main risk is incorrect interchange semantics. The first profile minimizes that surface, round-trips every generated DSN through the Specctra parser, and tests the artifact with Freerouting itself. The tool never mutates the live board and never replaces an output file.

Rollback is a straight revert of this commit; no board or config migration is required.


neusse:
Release-integration note: Konnect v0.10.0 added four placement tools while this branch was in review. That blind-side change conflicted with every generated tool-count surface, so the earlier green CI no longer represented the released registry. I rebased onto current main and synchronized this PR at 215 registered / 221 total tools. I also corrected the new public argument to manifest_output_path, added structured DSN/SES/bridge capability facts to the export result, and added a Freerouting v2.3.0-owned DSN corpus fixture alongside the real KiCad board fixture. The focused Specctra suite passes (10 passed, 1 live Java test intentionally ignored), and all six documentation/count contract tests pass. Fresh CI is now authoritative.

neusse:
Post-guidance-stack refresh: merged current `upstream/main` (through #364) into this branch without changing the DSN export design. The complete local gate passes: formatting, Clippy with warnings denied, workspace lib/test targets, and workspace doc tests. Updated head: `b7d8ebff6a7fdcb4813f9cf95373df526944439d`.

PR #343, opened 2026-08-26
fix(cli): require explicit init for guidance

### LINKED ISSUE #242: [BUG] MCP startup silently reinstalls guidance after uninstall
## Summary

Starting Konnect as an MCP server silently installs bundled client guidance whenever that client's marker is absent. This makes `konnect uninstall --client codex` non-durable: the next ordinary MCP restart recreates all six skills.

This is separate from #238. v0.6.1 correctly fixes help/argument dispatch, but the first-launch server path still calls `run_install_silent(client)` before MCP initialization.

## Reproduction (Konnect v0.6.1)

1. `konnect uninstall --client codex`
2. Verify `~/.agents/skills/{konnect,kicad-*}` are absent and `~/.konnect/.installed-codex` is absent.
3. Start the normal server command:
   `konnect --client codex --config <config>`
4. Send a valid MCP `initialize` request.

Observed:

```text
Before start: 0 native Codex skills; marker absent
MCP initialize: success, server version 0.6.1
After start: 6 native Codex skills; marker present
stderr: [konnect] Silent Codex install complete: 6 skills, 0 agents
```

The same defaulting rule means a piped server launch without `--client` selects Claude and can silently write Claude skills, agents, hooks, and settings.

## Cause

The server path in `crates/konnect/src/main.rs` unconditionally treats a missing marker as permission to install:

```rust
if install::needs_install(client) {
    let _ = install::run_install_silent(client);
}
```

The marker records that guidance was installed, not that the user declined it. Uninstall necessarily removes the marker, so restart reverses the user's explicit uninstall.

## Impact

- Explicit uninstall does not persist across MCP restarts.
- A server process mutates client configuration as a startup side effect.
- Versioned companion/plugin integrations cannot remain the sole guidance provider.
- Launching without an explicit client retains the risk of unexpected Claude filesystem changes.

## Recommended direction

Make MCP server startup non-mutating. Install guidance only through an explicit `konnect init --client ...` command.

If backward-compatible first-launch installation must remain, add an explicit server option such as `--no-install-guidance` and persist a separate declined/suppressed state so uninstall is not interpreted as consent to reinstall.

## Acceptance criteria

- `konnect uninstall --client codex` remains uninstalled after subsequent MCP starts.
- Starting the MCP server does not create or modify skills, agents, hooks, or client settings unless installation was explicitly requested.
- The same guarantee applies to Claude and Codex.
- A regression test covers uninstall → server start → guidance remains absent.
- Existing explicit `konnect init --client ...` behavior remains available.

neusse:
## Summary

- remove the silent guidance installer from normal piped MCP startup
- keep `--client` accepted and validated for command compatibility
- document that guidance installation requires an explicit `konnect init`
- regression-test clean-home and install/uninstall/start sequences for both Claude and Codex with the real CLI process

Closes #242.

## Root cause and design

Server startup treated an absent `.installed-{client}` marker as a first launch and called the silent installer. `uninstall` deliberately removes that same marker, so the next MCP launch could not distinguish an explicit uninstall from a new installation and recreated the guidance.

This selects the simpler policy approved in #242: server startup performs no guidance installation at all. Installation remains available only through the explicit `konnect init [--client ...]` command. No tombstone or new persistent state is needed.

- `--client claude` and `--client codex` remain valid server arguments.
- Existing explicit `init`, `status`, and `uninstall` behavior is unchanged.
- TTY/double-click setup behavior is unchanged.
- A fresh MCP registration no longer installs guidance implicitly; users who want bundled guidance run `konnect init` once.

## Tests

- `python -m unittest discover -s plugin/tests -v` (KiCad 10 bundled Python)

The real-process home-directory regression runs on Linux and macOS. It is intentionally excluded on Windows because `dirs::home_dir()` uses `SHGetKnownFolderPath` there and ignores `HOME`/`USERPROFILE`; attempting to redirect it would risk the runner's real client configuration. The normal Windows unit/integration suite remains green. Live-KiCad tests retain their repository-defined `ignored` status.

## Risk and rollback

Risk is limited to removing an implicit side effect. Rollback is the single commit, but would restore the destructive post-uninstall behavior. No file format, IPC protocol, packaging, schema, or configuration-key changes are involved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,8 +152,10 @@ konnect status --client codex
 konnect uninstall --client codex
 ```
 
-Keep `--client codex` in the MCP server command so first-launch setup also stays
-inside Codex's directories. For example, register a standalone binary with the
+MCP server startup never installs or restores guidance. Run `konnect init`
+explicitly when you want those files installed; after `konnect uninstall`,
+starting the server leaves them removed. `--client` remains accepted in server
+commands for compatibility. For example, register a standalone binary with the
 Codex CLI using:
 
 ```bash
--- a/crates/konnect/src/install.rs
+++ b/crates/konnect/src/install.rs
@@ -1,7 +1,7 @@
 //! Client-aware installer for Konnect's bundled guidance.
 //!
-//! Handles client-scoped install, uninstall, status, first-launch setup, and
-//! Claude hook integration without writing into another client's directories.
+//! Handles explicit client-scoped install, uninstall, status, and Claude hook
+//! integration without writing into another client's directories.
 
 use crate::manifest::{AGENTS, HOOK_SKILLS, SKILLS};
 use anyhow::{bail, Context, Result};
@@ -95,10 +95,6 @@ pub fn run_install(client: InstallClient) -> Result<()> {
     run_install_at(client, &InstallPaths::for_current_user()?, true)
 }
 
-pub fn run_install_silent(client: InstallClient) -> Result<()> {
-    run_install_at(client, &InstallPaths::for_current_user()?, false)
-}
-
 pub fn run_uninstall(client: InstallClient) -> Result<()> {
     run_uninstall_at(client, &InstallPaths::for_current_user()?, true)
 }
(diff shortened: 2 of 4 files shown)


neusse:
Release-integration note: Konnect v0.10.0 landed after this branchs earlier green CI, so the old run could not prove compatibility with the released base. I rebased onto current main and re-ran the installer/argument tests: 10 focused unit tests plus the advertised-help transaction test pass. The production behavior remains platform-independent; the real clean-home process test is still Unix-only because Windows home discovery cannot be safely redirected by that fixture. Fresh CI is now authoritative.

neusse:
Release 10 blind-side revalidation note: v0.10.0 changed the surrounding CLI, installer tests, and catalogue documentation after this PR was opened. This branch was rebased onto current upstream and the explicit-init/non-mutating-startup behavior was rerun against the v0.10 command surface. CI is green on Linux, Windows, and macOS.

PR #344, opened 2026-08-26
fix(windows): discover per-user KiCad installations

### LINKED ISSUE #254: [BUG] Per-user Windows KiCad installations are not discovered
Reported by @neusse in discussion #165 ("Per-user Windows KiCad libraries are not discovered") and by @woozer in #152 ("Konnect did not automatically find this per-user KiCad installation's symbol, footprint, and 3D-model directories"). Confirmed — two independent reporters, two separate code paths.

## Symptom

With KiCad installed per-user on Windows (under `%LOCALAPPDATA%\Programs\`, which is what KiCad's installer does when the user declines the machine-wide install), the standard KICAD10 symbol and footprint roots resolve to nothing. Both reporters had to set the KiCad 10 environment variables or add project-local aliases for libraries that were installed normally.

## Root cause

**Libraries** — `kicad_share_roots()`, `crates/konnect-core/src/tools/mod.rs:1119-1139`, probes only:

```
C:\KiCad\10.0\share\kicad
C:\Program Files\KiCad\10.0\share\kicad
C:\KiCad\9.0\share\kicad
C:\Program Files\KiCad\9.0\share\kicad
C:\KiCad\8.0\share\kicad
C:\Program Files\KiCad\8.0\share\kicad
```

`find_kicad_library_dirs` (`mod.rs:1196-1220`) checks the `KICAD{10,9,8}_{SYMBOL,FOOTPRINT,3DMODEL}_DIR` environment variables first, then these roots. No `%LOCALAPPDATA%\Programs\KiCad\...` anywhere.

**Binary** — `find_kicad_binary` (`verification.rs:428-465`) has the same gap with a different list: `C:\Program Files\KiCad`, `C:\KiCad`, `D:\KiCad`, `D:\Program Files\KiCad`. Also no AppData path.

`kicad_user_path_vars()` (`library.rs:1183-1200`) reads KiCad's own `kicad_common.json`, but that supplies user-defined path *variables*, not the install root, so it does not rescue this.

## Direction

- Probe `%LOCALAPPDATA%\Programs\KiCad\<ver>\` in both `kicad_share_roots()` and `find_kicad_binary`, and unify the two hardcoded lists — they have already drifted apart, which is how this ended up half-fixed in neither place.
- Prefer reading the install location from KiCad's own configuration or the Windows registry uninstall key over guessing paths, so a non-default install directory works too.
- The 3D-model root has the same problem and matters more than it looks: #239 reported the USB-C connector body missing from both the 3D render and the STEP export while the footprint preflight said a model was present. That may be this issue rather than a separate one — worth checking together.
- Whatever the resolution, an unresolved library root should be reported, not silently treated as "no libraries here".

neusse:
## Problem and scope

Per-user Windows KiCad installations under `%LOCALAPPDATA%\Programs\KiCad` were invisible to Konnect's executable and bundled-library discovery. The installer, core library resolver, UI launcher, and Python settings dialog had separate candidate lists that had already drifted.

This also blocks reliable PCB/Freerouting work: the native Specctra bridge and manufacturing flow need to locate `kicad-cli`, the KiCad UI, stock symbols/footprints, and 3D models on machines where KiCad was installed for the current user.

## Design

- Add one read-only `konnect-core::kicad_install` module for executable and shared-library discovery.
- Read both HKCU and HKLM uninstall records, including 32-bit and 64-bit registry views.
- Retain older KiCad vendor-key compatibility and existing platform paths.
- Add `%LOCALAPPDATA%`, `%ProgramFiles%`, and `%ProgramFiles(x86)%` fallbacks.
- Cache resolved installation roots for the process lifetime so registry enumeration occurs once.
- Preserve explicit `kicad_cli` and `kicad_binary` configuration precedence; derive the GUI beside an explicitly selected CLI before falling back to another install.
- Update the KiCad settings dialog so the path it persists can come from a current-user install.

No discovered path is written automatically. Windows/KiCad remain the persisted source of automatic discovery, and an explicit Konnect configuration remains the durable override.

No MCP tools, schemas, CLI flags, environment variables, config keys, or file formats change. Existing explicit paths and system-wide installs retain precedence and behavior. The only new runtime dependency is the Windows-only MIT-licensed `winreg` crate.

## Verification

- `cargo build --release -p konnect`
- `python -m unittest discover -s plugin/tests -v`
- Live Windows registry-isolated check with `LOCALAPPDATA` pointed at a nonexistent directory: found the HKCU KiCad 10.0.5 installation and verified `symbols`, `footprints`, and `3dmodels`.
- `konnect status --client codex`: found `C:\Users\georg\AppData\Local\Programs\KiCad\10.0\bin\kicad-cli.exe`.

Not run locally: `cargo deny check licenses` because `cargo-deny` is not installed. CI's license job remains the authoritative gate.

## Risk and rollback

Risk is limited to installation selection order. Explicit absolute configuration wins, then the configured CLI's sibling UI, followed by registry and well-known locations, newest supported KiCad first. Reverting this commit restores the previous independent path lists; no user data or configuration migration is involved.

Closes #254

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2381,6 +2381,7 @@ dependencies = [
  "tracing",
  "usvg",
  "uuid",
+ "winreg",
  "zip",
 ]
 
@@ -4595,6 +4596,16 @@ name = "winnow"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23b97319f7b8343df12cc98938e5c3eb436064524c8d2b4e30a1d3a36eecdf81"
+
+[[package]]
+name = "winreg"
+version = "0.56.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d6f32a0ff4a9f6f01231eb2059cc85479330739333e0e58cadf03b6af2cca10"
+dependencies = [
+ "cfg-if",
+ "windows-sys 0.61.2",
+]
 
 [[package]]
 name = "wit-bindgen"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,7 @@ base64 = "0.23"
 
 # Platform directories (home dir, config dir)
 dirs = "6"
+winreg = "0.56"
 
 # SVG parsing (import_svg_logo)
 usvg = "0.48"
--- a/crates/konnect-core/Cargo.toml
+++ b/crates/konnect-core/Cargo.toml
@@ -30,6 +30,9 @@ usvg.workspace = true
 zip.workspace = true
 tempfile = "3"
 
+[target.'cfg(windows)'.dependencies]
+winreg.workspace = true
+
 [dev-dependencies]
 # The pcb_components fallback tests spawn a mock KiCAD NNG endpoint to prove
 # a reachable-but-rejecting KiCAD never triggers the file-editing fallback.
--- a/crates/konnect-core/src/kicad_install.rs
+++ b/crates/konnect-core/src/kicad_install.rs
@@ -1,0 +1,395 @@
+//! KiCad installation discovery.
+//!
+//! Keep executable and bundled-library discovery behind this module so every
+//! caller sees the same installations in the same order.  Discovery is
+//! intentionally read-only: Windows and KiCad already persist install state,
+//! while an explicit Konnect configuration remains the durable override.
+
+use std::path::{Path, PathBuf};
+
+#[cfg(target_os = "windows")]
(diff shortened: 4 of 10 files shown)


neusse:
Release-integration note: Konnect v0.10.0 landed after this branchs earlier green CI, creating a blind-side compatibility risk despite a clean GitHub merge state. I rebased onto current main, retained unified registry/per-user discovery, and added a one-time explicit warning when no bundled KiCad library root resolves so missing stock symbols, footprints, and 3D models are not silent. The focused Rust discovery tests pass (3 passed, 1 live-install test intentionally ignored), and the two Python settings-dialog discovery tests pass. Fresh CI is now authoritative.

Issue #325, opened 2026-08-25
Issue: VS Code Copilot only exposes a subset of Konnect tools as callable


simachines:
## Summary

When using the Konnect MCP server with **VS Code Copilot** as the MCP client, the
client discovers all 203 Konnect tools (with `eager_toolsets = true`), but the
Copilot agent runtime only injects a **subset** into the model's callable tool
list. Calling any tool outside that subset returns:

```
Tool mcp_konnect_<name> is currently disabled by the user, and cannot be called.
```

This error is **not** produced by the Konnect server (its error taxonomy is
`toolset_not_loaded` / `unknown_tool`). It originates from the VS Code Copilot
client layer, which appears to cap the number of MCP tools it exposes to the
model to protect context budget.

## Environment

- Konnect MCP server (stdio transport)
- `konnect.toml`: `eager_toolsets = true`
- Client: VS Code Copilot agent (MCP client)
- All 19 toolsets / 203 tools confirmed loaded server-side via `list_toolboxes`
  and `get_active_toolsets`
- KiCad 10.0 running, IPC responsive (`check_kicad_ui` → `ipc_responsive: true`)

## Observed behavior

| Tool | Result |
|------|--------|
| `get_board_extents` | ✅ works |
| `set_active_layer` | ✅ works |
| `check_kicad_ui` | ✅ works |
| `get_board_info` | ❌ "disabled by the user" |
| `get_design_rules` | ❌ "disabled by the user" |
| `get_nets_list` | ❌ "disabled by the user" |
| `get_schematic_layout` | ❌ "disabled by the user" |
| `get_layer_list` | ❌ "disabled by the user" |
| `get_active_toolsets` | ❌ "disabled by the user" (intermittent) |

The set of callable tools is **unstable** — it changes between calls within the
same session. Tools that worked earlier (e.g. `get_design_rules`,
`get_active_toolsets`) later return "disabled by the user."

## Key observations

1. The server is healthy: all 203 tools are registered and loaded
   (`eager_toolsets = true` → `router.load_all()`).
2. The server's own call log (`get_recent_calls`) shows **no trace** of the
   blocked tools — they never reach the server. The block happens client-side.
3. The `activate_*` meta-tools (e.g. `activate_component_and_schematic_management`)
   print the full tool manifest as **text**, but the listed tools are **not
   callable** — text manifests don't make tools callable.
4. Reloading the VS Code window does **not** fix it.

## Hypothesis

[...]

PR #347, opened 2026-08-27
docs(troubleshooting): explain Copilot tool limits

neusse:
## Problem and scope

Issue #325 documents a VS Code Copilot failure mode where Konnect tools are discovered but only an arbitrary subset is callable. The existing troubleshooting advice recommended `eager_toolsets = true` for clients that cache the initial tool list, but that does not solve a client-wide callable-tool ceiling.

This documentation-only change:

- distinguishes initial-list caching from Copilot's total tool budget;
- explains why the `currently disabled by the user` error never reaches Konnect;
- records the reporter's measured 128-tool ceiling without treating it as a Konnect limit;
- points readers to the community two-tool relay demonstrated in #325;
- makes clear that the relay is not shipped or reviewed by Konnect; and
- points to the planned native compact tool surface in the roadmap.

## Root cause and design

The evidence in #325 shows two client-side behaviors: Copilot does not refresh the tool list after `notifications/tools/list_changed` when lazy loading is used, while eager loading exceeds Copilot's aggregate callable-tool budget. The troubleshooting guidance now steers users according to those distinct cases and preserves #325 for the native server-side compact-surface work.

## Compatibility

Documentation only. No runtime behavior, API, configuration default, or generated asset changes.

## Verification

- `cargo test -p konnect --locked --test asset_references --test doc_tool_counts`
- `git diff --check`

## Risk and rollback

The primary risk is future Copilot behavior changing. The text attributes the measured ceiling to the #325 report and presents the relay as a current community workaround. Rollback is a single documentation commit.

Refs #325

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -185,7 +185,8 @@ call it. That is the symptom in
 [#169](https://github.com/mixelpixx/Konnect/issues/169) — reported against
 Claude Desktop.
 
-The fix for those clients is to make the *first* listing complete:
+For clients that cache the initial list but do not also cap the number of
+callable tools, the fix is to make the *first* listing complete:
 
 ```json
 { "eager_toolsets": true }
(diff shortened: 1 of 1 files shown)


neusse:
Release-integration note: Konnect v0.10.0 landed after this documentation branchs earlier green CI. I rebased it onto current main and rechecked the troubleshooting text against the released 214-domain-tool/20-toolset server. The guidance remains correct because it describes the Copilot client-side aggregate tool ceiling and stale initial-list behavior rather than a fixed Konnect tool count. No code behavior changed; fresh CI is now authoritative.

neusse:
Release 10 blind-side revalidation note: v0.10.0 changed the documented catalogue to 20 toolsets / 214 registered / 220 total tools after this PR was opened. The troubleshooting text was rebased and rechecked against the released v0.10 client-routing behavior; it remains a documentation workaround for client tool caps, not a claim that Konnect itself is missing tools. CI is green.

PR #348, opened 2026-08-27
fix(pcb): preserve library user text during refresh

### LINKED ISSUE #331: update_footprints_from_library refuses most official library footprints: fp_text user is unsupported
## What happens

`update_footprints_from_library` (#232) refuses any library footprint carrying `fp_text user` — and that is most of them: **393 of 400 randomly sampled official KiCad-10 library footprints** carry `(fp_text user "${REFERENCE}")` on `F.Fab`. On a board placed from official libraries, an unfiltered run reports `status: conflict` with `unsupported_library_footprint` diagnostics for nearly every candidate.

Verified live during the #232 merge gate: the stock ecc83 demo project conflicts on all 15 footprints until the fab text is removed from the library entry, after which plan → apply → noop convergence works end to end.

## Why it refuses (and why that's currently right)

`validate_supported_children` (`crates/konnect-core/src/tools/pcb_footprint_update.rs`) only accepts `fp_text` kinds `reference`/`value`. The refusal is the honest choice today, because applying an update while *dropping* the fab text would silently delete library artwork from the board — the #244 shape of failure.

## The path to fixing it

`konnect_ipc::IpcGraphicDefinition::Text` already exists, maps to a footprint text child, and is live-proven (it's how the #237 fixtures place `Dwgs.User` text). The work is:

1. Teach `parse_library_footprint` the standard fab-text shape — `(fp_text user "…" (at …) (layer …) (uuid …) (effects (font (size N N) (thickness T))))` — mapping to `IpcGraphicDefinition::Text`. Refuse anything beyond it (justify, hide, italic, non-square size) exactly as the validator refuses today.
2. Verify the round-trip against KiCad's own serialization: `Text` currently carries no stroke thickness, and library files universally write `(thickness 0.15)` for 1 mm text. If the rebuilt text re-saves with a different thickness, the tool's own dry-run-after-apply stops converging to `noop` — that convergence property is the built-in detector for any lossy mapping, so the acceptance test is: apply against an official-library footprint, KiCad save, dry run again, assert `noop`.
3. The live gate from #232's merge comment is reusable as-is; just skip the fab-text-stripping step.

Until then the tool is fully usable on libraries without fab text (e.g. Konnect-authored ones), and its refusal names the exact child.

neusse:
Closes #331.

## Problem

`update_footprints_from_library` rejected the `fp_text user` blocks emitted by ordinary KiCad 10 footprint libraries. A sample of the installed official footprint corpus found user text in 15,238 of 15,447 files, which made the safety validator refuse most real library refreshes.

The parser already extracted those text blocks, but the typed IPC representation did not retain their explicit font stroke width. Simply allowing them would therefore make refresh lossy.

## Change

- accept the conservative, losslessly representable `fp_text user` shape produced by KiCad
- preserve explicit font stroke width through parsing, mirroring, and IPC construction
- continue to fail closed for unsupported modifiers such as justification, italic text, unlocked placement, non-square fonts, or missing thickness
- add a KiCad-serialized fixture plus positive and negative regression coverage
- extend the ignored live-KiCad test to verify explicit thickness, save the board, and prove a subsequent dry run converges to `noop`

No MCP tool schema or public command surface changes.

## Verification

- `git diff --check`
- live KiCad 10 PCB Editor: `footprint_library_update_apply_then_dry_run_is_noop` passed against a disposable board through the API socket, including save and post-save no-op verification
- installed official KiCad corpus scan: 14,921 of 15,447 footprints fit the deliberately narrow supported shape; uncommon variants remain structured conflicts instead of being silently altered

### DIFF
--- a/crates/konnect-core/src/tools/pcb_components.rs
+++ b/crates/konnect-core/src/tools/pcb_components.rs
@@ -380,6 +380,14 @@ fn text_size(node: &konnect_sexp::SexpNode) -> f64 {
         .unwrap_or(1.0)
 }
 
+/// Explicit font stroke width, falling back to KiCad's 15%-of-size default.
+fn text_stroke_width(node: &konnect_sexp::SexpNode) -> f64 {
+    node.find("effects")
+        .and_then(|effects| effects.find("font"))
+        .and_then(|font| font.find_f64("thickness"))
+        .unwrap_or_else(|| text_size(node) * 0.15)
+}
+
 /// Text position and angle from `(at x y [rot])`.
 fn text_at(node: &konnect_sexp::SexpNode, kind: &str) -> anyhow::Result<((f64, f64), f64)> {
     let at = node
@@ -522,6 +530,7 @@ pub(crate) fn extract_graphic_definitions(
             rotation,
             layer: graphic_layer(text, "fp_text")?,
             size: text_size(text),
+            stroke_width: text_stroke_width(text),
         });
     }
     for property in footprint.find_all("property") {
@@ -546,6 +555,7 @@ pub(crate) fn extract_graphic_definitions(
             rotation,
             layer,
             size: text_size(property),
+            stroke_width: text_stroke_width(property),
         });
     }
     Ok(graphics)
--- a/crates/konnect-core/src/tools/pcb_components.rs
+++ b/crates/konnect-core/src/tools/pcb_components.rs
@@ -530,7 +530,7 @@ pub(crate) fn extract_graphic_definitions(
             rotation,
             layer: graphic_layer(text, "fp_text")?,
             size: text_size(text),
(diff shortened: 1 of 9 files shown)


neusse:
Release-integration note: Konnect v0.10.0 landed after this branchs earlier green CI and changed the base underneath it. That created a blind-side compatibility risk even though GitHub still reported this PR mergeable. I rebased the branch onto current main and re-reviewed the public types against the current naming rules. The refresh text field is now stroke_width_mm, and the fail-closed parser rejects repeated uuid/tstamp identifiers or both forms together instead of accepting input the typed rebuild cannot preserve. Targeted konnect-core footprint-refresh tests (24) and the full konnect-ipc suite pass. Fresh CI is now authoritative; the pre-v0.10 run is not.

PR #349, opened 2026-08-27
fix(schematic): resolve analysis pins by placed unit

neusse:
## Summary

Multi-unit symbols are now resolved through the unit actually placed on the sheet. This prevents analysis and connectivity repair from reporting or snapping to phantom pins, while allowing component queries to see nets on later units.

The PR also completes the requested `sch_batch` audit with handler-level tests against the real KiCad ECC83 fixture: bulk move, shared-field edit, and both reference-delete tools operate on all three units without changing a neighboring component or the embedded library definition.

Part of #182.

## Approach

The affected analysis and export handlers now reuse the existing shared `placed_pins_by_reference` / `placed_pins` path, which selects common pins plus the placed unit before applying that unit's transform. This removes the duplicated `extract_lib_pins` logic that superimposed every library unit on every placement.

`export_netlist_summary` still emits one entry per placed symbol instance and preserves unresolved instances with an empty pin list. `fix_connectivity` changes only its candidate pin endpoints; labels and wire endpoints are unchanged.

`design_review.rs` is intentionally excluded because PR #336 owns that file. Its author has been asked to make the same unit-aware correction and add ECC83 coverage.

## Compatibility and safety

No tool names, schemas, response fields, configuration, IPC, or file formats change. Existing single-unit behavior is preserved; multi-unit coordinates, connectivity, and per-placement pin counts are corrected.

The only mutating production path touched is `fix_connectivity`; its existing dry-run behavior, conditional atomic write, and stale-file protection are unchanged. The narrower target set prevents a wire from being moved to an unplaced unit's phantom pin. Rollback is one commit.

## Validation

- [x] `cargo test -p konnect-core --locked multi_unit_ -- --nocapture` (34 passed)
- [x] Real KiCad ECC83 schematic fixture exercised through registered MCP tool handlers
- [ ] Live KiCad GUI/e2e checks were not run; these S-expression paths do not require a running editor, and the relevant environment-dependent tests remain ignored as designed

## Review checklist

- [x] No new public names or renames.
- [x] File mutations retain the existing atomic-write and stale-source protections.
- [x] No IPC mutation changes.
- [x] No tools were added or removed, so tool counts and directory documentation are unchanged.

### DIFF
--- a/crates/konnect-core/src/tools/sch_analysis.rs
+++ b/crates/konnect-core/src/tools/sch_analysis.rs
@@ -6,7 +6,9 @@
 use crate::mcp::protocol::CallToolResult;
 use crate::tool;
 use crate::tools::sch_connectivity::{net_graph_for, pt_key, ConnectivityIndex};
-use crate::tools::{get_path, opt_f64, require_f64, require_str, ToolContext, ToolDef};
+use crate::tools::{
+    get_path, opt_f64, placed_pins_by_reference, require_f64, require_str, ToolContext, ToolDef,
+};
 use konnect_schematic_editor as cse;
 use konnect_sexp::{
     geometry::{point_on_segment, points_coincident},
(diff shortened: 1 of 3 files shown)


neusse:
Release-integration note: Konnect v0.10.0 landed after this branchs earlier green CI, so I rebased the multi-unit analysis patch onto current main and re-ran the multi-unit suite. All 34 focused tests pass across analysis, export, sch_batch, and existing component mutations. This PR intentionally remains Part of #182: design_review still has unit-blind walks owned by the remaining work, so the issue must not close on this merge. Fresh CI is now authoritative.

neusse:
Release 10 blind-side revalidation note: v0.10.0 changed the surrounding schematic guidance and repository test surface after this PR was opened. This draft was rebased onto current upstream, the real ECC83 multi-unit slice was rerun, and all platform CI is green. Scope remains deliberately partial: design_review in #336 must become unit-aware before #182 closes.

Issue #357, opened 2026-08-28
[BUG] Claude agents and skills can claim evidence they never collect


neusse:
## Summary

The Claude-facing package in Konnect v0.10.0 installs six skills, nine reference files, and two agents, but the pieces are not connected into one evidence-driven workflow:

- all nine reference files are installed with zero incoming links from a parent skill or agent;
- neither Claude agent preloads the skills that define its workflow;
- both agents request or report checks that their prescribed toolsets cannot run;
- duplicated inline guidance has drifted from the hidden references;
- manufacturing guidance describes checks the released validator does not implement;
- several electrical/fabrication tables are unsourced, internally contradictory, or unsafe as universal defaults.

The result is guidance that can look complete while Claude either never sees the detailed material or fills a success field for evidence it was never instructed to collect.

## Pinned release evidence

Audited release: [v0.10.0](https://github.com/mixelpixx/Konnect/releases/tag/v0.10.0), tag commit `866933e8aca1c115479963463cc7e34370b5822b`.

Package inventory from the released manifest:

- 6 skills
- 9 reference files
- 2 Claude agents
- 1 Claude hook (tracked separately in #358)

Related direction:

- #295 contains the benchmark-driven Claude workflow proposal.
- [`ROADMAP.md` section 5](https://github.com/mixelpixx/Konnect/blob/866933e8aca1c115479963463cc7e34370b5822b/ROADMAP.md#L90-L98) explicitly invites skill-layer PRs for the evidence hierarchy, one live-board owner, Freerouting-first workflow, and physical pin maps.
- #356 tracks the separate electrically unsafe symbol-ID and pin-map corrections.

## Finding 1: all nine installed references are unreachable through progressive disclosure

The installer correctly copies each reference under its parent skill. However, a repository-wide incoming-link check found **zero mentions** of every reference filename outside the file itself:

- `kicad-schematic/references/common-lib-ids.md`
- `kicad-schematic/references/wiring-patterns.md`
- `kicad-pcb/references/design-rules.md`
- `kicad-pcb/references/layer-reference.md`
- `kicad-pcb/references/trace-width-table.md`
- `kicad-review/references/design-checklist.md`
- `kicad-review/references/error-taxonomy.md`
- `kicad-manufacture/references/gerber-layers.md`
- `kicad-manufacture/references/jlcpcb-rules.md`

[...]

PR #363, opened 2026-08-29
fix(guidance): make evidence workflows executable

neusse:
## Problem and scope

The bundled schematic builder and design reviewer report ERC/DRC and readiness fields that their setup does not load or execute. The schematic workflow also lacks direct short detection, rendered layout acceptance, and a fail-closed result when evidence is unavailable.

This is PR B for #357. It is independent in Git history but should merge after #362, which supplies the agent skill preloads and router wiring.

## Design

- Load and execute direct ERC, DRC, short, connection, save, and render tools before reporting their results.
- Make every reported validation field use PASS/FAIL/BLOCKED/N/A rather than an unchecked success box.
- Replace universal decoupling/ESD/bulk-capacitance rules with requirements- and datasheet-dependent review criteria.
- Establish the evidence hierarchy: requirements and datasheets, direct KiCad evidence, direct Konnect evidence, aggregate reports, then heuristics.
- Treat orphan and single-pin checks as candidates that require corroboration.
- Make missing, failed, structurally impossible, or contradictory required evidence produce `INCOMPLETE`.
- Require functional grouping, label-inclusive overlap inspection, and page-boundary acceptance from an inspected schematic render.
- Add mechanical guards that bind agent output claims to loaded/executed evidence and keep agent/skill completion boundaries aligned.

## Compatibility

Bundled Claude guidance and its asset tests change. MCP tools, schemas, response shapes, file formats, installer paths, and Codex behavior are unchanged.

## Validation

The red baseline identified every missing toolset/call and every absent completion marker before the prose changed. The following now pass:

- `cargo test -p konnect --test asset_references`

Real-KiCad GUI tests remain intentionally ignored outside their dedicated workflow.

## Risk and rollback

The main risk is additional tool/context usage during full builds and reviews; those calls are the evidence the existing reports already claim. The change is guidance-only and can be reverted without migration.

Part of #357. Depends on #362 for the connected-agent package.

### DIFF
--- a/crates/konnect/assets/agents/kicad-design-review-agent.md
+++ b/crates/konnect/assets/agents/kicad-design-review-agent.md
@@ -2,6 +2,10 @@
 name: kicad-design-review-agent
 description: "Performs a thorough hardware design review of a KiCAD project. Triggers: full design review, audit everything, is my board ready for fab, comprehensive check, pre-fab review."
 model: sonnet
+skills:
+  - konnect
+  - kicad-review
+  - kicad-manufacture
 tools:
   - mcp__konnect__*
 maxTurns: 25
--- a/crates/konnect/assets/agents/kicad-design-review-agent.md
+++ b/crates/konnect/assets/agents/kicad-design-review-agent.md
@@ -9,7 +9,7 @@ maxTurns: 25
 
 ## System Prompt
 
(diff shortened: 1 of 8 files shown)


neusse:
Integration check update: simulating the intended guidance PR order found an overlap with #362 in the review skill and asset-reference tests. I merged the #362 topic branch into this branch and resolved the overlap by preserving both contracts: reachable skill/agent/reference wiring and executable evidence boundaries. The combined branch passes cargo fmt --all -- --check, workspace Clippy with warnings denied, all workspace lib/test targets, and workspace doc tests. CI is rerunning on the new head.

PR #369, opened 2026-08-29
fix(project): report actual generator versions

### LINKED ISSUE #368: get_project_info reports project filename as kicad_version
## Summary

`get_project_info` currently populates `kicad_version` from `meta.filename` in the `.kicad_pro` file. That value is the project filename, not a KiCad version.

Observed with a generated test project:

```json
{"kicad_version":"widget.kicad_pro"}
```

## Expected behavior

Report version information from fields that actually represent versions, without claiming a single version when the schematic and PCB disagree.

Proposed response fields:

- `project_file_version` from `.kicad_pro` `meta.version`
- `schematic_generator_version` from the sibling `.kicad_sch`
- `pcb_generator_version` from the sibling `.kicad_pcb`
- compatibility field `kicad_version`, set only when both sibling versions agree or only one sibling version is available; otherwise `null`

## Cause

The current handler reads:

```rust
pro.get("meta").and_then(|m| m.get("filename"))
```

The sibling KiCad S-expression files contain the actual `generator_version` values.

## Verification

The proposed fix reports the generated project's `10.0` schematic and PCB generator versions, preserves the project-file format version separately, and includes a regression test for mismatched sibling versions.

dubesinhower:
## Summary

Correct `get_project_info` so `kicad_version` no longer contains the project filename. Add explicit project-file, schematic-generator, and PCB-generator version fields.

Closes #368

## Approach

The handler previously read `.kicad_pro` `meta.filename` into `kicad_version`. It now:

- reads `meta.version` as `project_file_version`
- parses `generator_version` from sibling `.kicad_sch` and `.kicad_pcb` files
- keeps `kicad_version` as a compatibility field when both sibling versions agree or only one is available
- returns `null` for `kicad_version` when both sibling versions exist but disagree

A regression test covers both matching and mismatched sibling versions.

## Compatibility and safety

This is an additive response-schema change plus a correction to the existing `kicad_version` value. Consumers that treated the previous filename value as a version were already receiving invalid data. No KiCad files or IPC state are mutated; sibling metadata reads are best-effort, and missing or unparseable siblings produce `null` version fields.

## Validation

- [x] `cargo test -p konnect-core get_project_info_ --locked`
- [x] `cargo test -p konnect-core --locked --lib --tests`
- [ ] `cargo test --workspace --locked --lib --tests` — ran; all relevant tests passed, but `every_installed_demo_board_scans_losslessly` fails on a locally installed KiCad 10 RoyalBlue demo containing two unparsed zones. The identical failure was reproduced on an untouched checkout of upstream `main`.
- [x] Rebuilt MCP server called against an existing KiCad 10 project; returned matching `10.0` sibling generator versions and the separate project-file format version.

## Review checklist

- [x] New names follow existing project/version terminology; no public field is removed.
- [x] New behavior and the disagreement path have regression coverage.
- [x] File mutation checklist is not applicable; this tool is read-only.
- [x] IPC mutation checklist is not applicable; this tool performs no IPC mutation.


neusse (commented):
Reviewed current head `49b7d0e56aac8b97d3ad066e9b67d55997b11b06` against current `main` after #372 merged.

The implementation looks correct and focused: it stops reporting the project filename as `kicad_version`, separates the project-file format version from the sibling schematic/PCB generator versions, preserves a compatibility value when the siblings agree or only one exists, and returns `null` when they disagree. The read-only behavior and regression coverage match #368. Because this adds public response fields and corrects the meaning of an existing field, it belongs in minor-release notes.

This is the first PR in the lowest-rework queue:

`#369 -> #374 -> #376 -> #379 -> #381 -> corrected #388 -> #389 -> #392 -> #393 -> #394 -> #390`

#391 remains draft and comes later. We will handle the independent #382 IPC work after #390 so it absorbs the combined IPC rebase rather than forcing this longer train to move again.

Please update #369 onto current `main` now and rerun the complete, unfiltered contributor gate. The previous validation notes could legitimately exclude the old #372 zone failure; that exception is gone now. Once the updated head is green, this should be the next merge candidate. Please do not update every later branch yet: updating only the next PR after its predecessor merges avoids repeated rebases.

PR #372, opened 2026-08-29
fix(sexp): preserve arc-only zone outlines

### LINKED ISSUE #370: KiCad 10 arc-only zone outlines are skipped by board scanner
## Summary

`konnect_sexp::board::zones` skips valid KiCad 10 zones whose authored polygon path consists entirely of `(arc (start ...) (mid ...) (end ...))` entries.

This is observable in KiCad 10's public `royalblue54L_feather` demo: two copper zones are counted as unreadable even though pcbnew authored them and their net, layer, and outline geometry are complete.

## Reproduction

On untouched `mixelpixx/Konnect@428c1bd` with the KiCad 10 demo corpus installed:

```text
cargo test -p konnect-sexp --test board_fixtures every_installed_demo_board_scans_losslessly -- --nocapture
```

Observed:

```text
royalblue54L_feather/RoyalBlue54L-Feather.kicad_pcb: skipped 0 segments / 0 vias / 2 zones from a pcbnew-authored board
```

The two rejected paths contain eight and seven ordered `arc` elements respectively and no `xy` elements.

## Root cause

`zones` currently collects only direct `(xy ...)` children from `(polygon (pts ...))` and then requires at least three points. Complete arc elements are silently ignored, so an arc-only outline appears empty.

## Expected behavior

- Preserve the authored order and all start/mid/end coordinates of `xy` and `arc` outline elements.
- Retain each zone's resolved net and layer set.
- Keep point-only zone behavior compatible.
- Continue counting incomplete, non-finite, or unknown outline elements as unreadable instead of fabricating geometry.

## Proposed scope

Add an ordered zone-outline element representation to `konnect-sexp`, keep the existing point-only view unchanged for straight polygons, add a reduced fixture containing only the two public-demo zone forms, and rerun the installed-demo conformance test plus the full contributor gates.

dubesinhower:
Closes #370

## Problem and scope

KiCad's installed `royalblue54L_feather` demo contains two valid copper zones whose authored polygon paths consist entirely of ordered `arc` elements. The point-only `konnect_sexp::board::zones` scanner sees no `xy` vertices in those paths, so the installed-demo conformance test counted both zones as unreadable.

This PR changes only read-only board geometry scanning and its tests.

## API-compatible design

- Preserve the existing public `ZoneOutline { net, layers, points }` struct and `zones()` point-only behavior unchanged for downstream Rust struct literals and exhaustive patterns.
- Add the separate `LosslessZoneOutline`, `ZoneOutlineElement`, and `lossless_zone_outlines()` API for exact ordered `xy` and three-point `arc` geometry.
- Reject an entire lossless zone when an outline contains an unknown element, an incomplete or duplicate arc tag, non-finite coordinates, or trailing coordinate data. No partial geometry is returned as lossless.
- Keep `Scan::skipped` as the fail-closed signal for malformed matching nodes.

## KiCad-authored evidence

- Replace the inline hand-reduced wrapper with a 4 KB board serialized by KiCad 10.0.5.
- Record the installed demo path, generator/file versions, source SHA-256, and both source zone UUIDs in `zone_outline_elements.README.md`.
- Preserve every ordered arc and all start/mid/end coordinates from both source zones.
- Add a KiCad-API-authored mixed `xy`/`arc` outline and assert both its complete ordered geometry and the legacy `zones()` point projection.
- Cover incomplete arcs, duplicate required tags, non-finite coordinates, trailing arc/point coordinate data, and unknown outline element kinds.

The full 3.6 MB public demo is not copied into the repository; the provenance-backed reduced fixture retains only the load-bearing geometry.

## Compatibility

There is no MCP, JSON, file-format, or write-path change. Existing Rust consumers of `ZoneOutline` and `zones()` remain source-compatible and behavior-compatible. The lossless scanner and types are new opt-in public API.

## Verification

- `cargo test -p konnect-sexp --locked --lib --tests`
- `cargo test -p konnect-sexp --locked --doc`
- `cargo clippy -p konnect-sexp --locked --all-targets -- -D warnings`
- Installed KiCad 10 corpus: 19 boards, 765 parsed zones, zero skipped zones

All pass locally. Existing tests that require dedicated live-GUI or E2E workflows retain their existing ignored annotations.

## Risk and rollback

Risk is limited to the new read-only lossless scanner and the installed-demo conformance oracle switching to it. Unknown or ambiguous geometry still refuses rather than disappearing silently. Reverting this PR restores the previous scanner behavior without touching project files or mutation paths.


neusse (changes requested):
The core defect and parser direction are valid. I reproduced the failure against the installed KiCad 10 corpus: current `main` parses 763 zones and skips these 2; this branch parses all 765 with zero skipped. The exact coordinates in the test are genuine demo-board golden data, not production hard-coding.

I am requesting changes for the compatibility and evidence gaps documented in #370:

1. `ZoneOutline::elements` is not source-compatible for external Rust consumers constructing or exhaustively matching the existing public struct. Please preserve the existing API through a separate lossless type/scanner, or coordinate and document an intentional incompatible pre-1.0 release and migration. The PR description must not call the current field addition merely additive.
2. Replace the hand-reduced inline wrapper with a KiCad-saved fixture carrying durable source/version/hash/zone provenance, per the repository fixture rule.
3. Assert every ordered arc and all start/mid/end coordinates, not only the first and last elements, and add a KiCad-authored mixed `xy`/`arc` case with its compatibility projection.
4. Extend the malformed-shape evidence so duplicate/incomplete/non-finite coordinates and unknown outline elements cannot be silently accepted or discarded under a lossless result.

The full reproduction and refined acceptance criteria are in https://github.com/mixelpixx/Konnect/issues/370#issuecomment-5466210716.

Please keep working in this PR; a replacement is not needed. Once these points are addressed, the demonstrated functional fix should be straightforward to re-review.

dubesinhower:
Addressed the requested compatibility and evidence changes in `9aedb2d`:

1. Restored `ZoneOutline` and `zones()` to their original source-compatible public shape and behavior; exact curved geometry now uses the separate opt-in `LosslessZoneOutline` / `lossless_zone_outlines()` API.
2. Replaced the inline wrapper with a 4 KB KiCad 10.0.5-saved board plus provenance README recording the installed source path, generator/file versions, SHA-256, and both source zone UUIDs.
3. Asserted every ordered arc and every start/mid/end coordinate for both source zones, and added a KiCad-API-authored mixed `xy`/`arc` path with its complete lossless result and legacy point projection.
4. Added fail-closed coverage for incomplete arcs, duplicate required tags, non-finite values, trailing point/arc coordinate data, and unknown outline elements.

The installed KiCad corpus now reports 765 parsed zones and zero skipped zones. All ten CI checks are green across Linux, macOS, Windows, Nix, clippy, formatting, packaging, plugin, licensing, and the schematic viewer.

Ready for re-review.

neusse (approved):
Re-reviewed exact head 9aedb2dc3cab3557207a1d13c56d1ec51d8fc8f1. The requested compatibility and evidence changes are complete: the existing ZoneOutline/zones API remains intact, the lossless geometry has a separate opt-in representation, the fixture is KiCad-authored with provenance, the full ordered arc and mixed-path geometry is asserted, malformed forms fail closed, and the installed-demo result is 765 parsed zones with zero skipped. All ten required checks are green. Approved.

PR #374, opened 2026-08-30
fix(pcb): preserve footprint properties during refresh

### LINKED ISSUE #373: update_footprints_from_library refuses generated and custom properties
## What happens

`update_footprints_from_library` refuses KiCad 10 library footprints that carry generated or custom top-level properties. A stock example is `Capacitor_SMD:C_0603_1608Metric`, whose official library footprint contains:

```scheme
(property "KiLib_Generator" "SMD_2terminal_chip_molded" ...)
```

On untouched `main` at `428c1bd`, a focused parser probe fails before a refresh plan can be built:

```text
property 'KiLib_Generator' is not supported losslessly by typed library refresh
```

The public tool therefore reports `status: conflict` with an `unsupported_library_footprint` diagnostic for an otherwise supported official footprint.

## Root cause

`validate_supported_children` accepts a top-level `property` node, then whitelists only `Reference`, `Value`, `Datasheet`, and `Description`. `parse_library_footprint` consequently never carries other properties into the typed refresh representation.

This does not need a lossy exception. KiCad's IPC `Footprint.items` collection can carry `kiapi.board.types.Field` children, and Konnect already handles `Field` items when transforming footprint children.

## Proposed design

1. Parse losslessly representable non-mandatory library properties into typed `Field` children, retaining name, value, position, layer, visibility, and text attributes. Library child identifiers are definition-local, so KiCad should issue identifiers for the refreshed board children.
2. During rebuild, preserve instance-only custom fields and refresh same-name library-owned fields deterministically; reject ambiguous duplicate names.
3. Keep the four mandatory properties on their existing first-class fields.
4. Validate every property clause. If any property cannot be represented by the IPC model, name that property and clause in a conflict and refuse before `BeginCommit` rather than silently dropping it.
5. Cover `KiLib_Generator`, an ordinary custom property, preservation of instance-only fields, front/back transforms, apply-to-noop convergence, and pre-commit refusal with KiCad-authored fixtures.

## Compatibility

This is a behavior fix with no tool-schema or response-schema change. Previously refused, representable footprints become refreshable; genuinely unrepresentable properties continue to fail closed with a more specific diagnostic.

Implementation: #374

dubesinhower:
## Summary

- parse losslessly representable non-mandatory footprint properties as typed `Field` children
- preserve instance-only fields while deterministically refreshing same-name library fields
- transform property positions, angles, layers, visibility, and text attributes with front/back footprint placement
- validate mandatory and custom properties through one lossless clause parser while keeping `Reference`, `Value`, `Datasheet`, and `Description` in their existing first-class fields
- reject duplicate property names and repeated scalar clauses (`hide`, `knockout`, font `bold`, `italic`, and `line_spacing`) before a commit begins

Closes #373

## Problem and approach

KiCad 10 footprints can carry non-mandatory properties such as `KiLib_Generator`. The previous refresh parser refused those properties outright. The initial PR added typed-field support, but mandatory properties bypassed that parser and several repeated scalar clauses could still be accepted by last-value-wins overwriting.

This change now walks every property once, enforces name uniqueness across mandatory and custom properties, and validates every authored clause through the same parser. Mandatory values continue to use the current dedicated IPC fields; custom values are emitted as typed `Field` children. Unsupported or ambiguous input fails closed before `BeginCommit`.

The included KiCad-authored fixtures cover generator metadata, an ordinary custom property, visibility and text presentation, all four mandatory property names, and the repeated scalar cases raised in review. One fixture is the stock KiCad 10.0.5 `Capacitor_SMD:C_0603_1608Metric` output and pins its `KiLib_Generator` field through the production parser.

## Compatibility, risk, and rollback

- No MCP tool or response schema changes.
- Losslessly representable stock footprints retain the same refresh behavior, including instance-only fields and existing placement/net preservation.
- The intentional behavior change is stricter refusal of duplicate property names, unsupported mandatory-property clauses, and repeated scalar clauses that were previously ignored or overwritten.
- Refusals occur during planning, before a write transaction begins.
- Rollback is a straight revert of the three commits in this PR; no persisted data migration is involved.

## Verification

- `cargo test -p konnect-core pcb_footprint_update::tests -- --nocapture` (31 passed)
- `cargo test --workspace --locked --lib --tests` (passed, including `every_installed_demo_board_scans_losslessly`)
- `cargo test -p konnect --test live_kicad_tools footprint_library_update_apply_then_dry_run_is_noop -- --ignored --exact --nocapture --test-threads=1` (1 passed)

The ignored live-GUI regression ran against KiCad 10.0.5 on Windows using an isolated KiCad profile and a disposable board. It exercised dry-run readiness, live IPC apply and readback, save, a post-save dry-run no-op, and DRC confirmation that `lib_footprint_mismatch` was absent. A separate Konnect readback after the run confirmed the requested board was still open and the board-info source was `ipc`.

## Guard evidence

- Temporarily bypassing mandatory-property parsing and global property-name uniqueness made the regression fail with all eight accepted cases: unsupported clauses and duplicate names for `Reference`, `Value`, `Datasheet`, and `Description`.
- Temporarily bypassing the repeated-scalar guards made the regression fail for all five clauses: `hide`, `knockout`, `bold`, `italic`, and `line_spacing`.
- Restoring the guards returns the focused suite to 31/31 passing.

### DIFF
--- a/crates/konnect-core/src/tools/pcb_components.rs
+++ b/crates/konnect-core/src/tools/pcb_components.rs
@@ -410,6 +410,19 @@ pub(crate) fn extract_field_placement(source: &str) -> konnect_ipc::IpcFieldPlac
 pub(crate) fn extract_graphic_definitions(
     source: &str,
 ) -> anyhow::Result<Vec<konnect_ipc::IpcGraphicDefinition>> {
+    extract_graphic_definitions_with_properties(source, true)
+}
+
+pub(crate) fn extract_graphic_definitions_without_properties(
+    source: &str,
+) -> anyhow::Result<Vec<konnect_ipc::IpcGraphicDefinition>> {
+    extract_graphic_definitions_with_properties(source, false)
+}
+
+fn extract_graphic_definitions_with_properties(
+    source: &str,
+    include_properties: bool,
+) -> anyhow::Result<Vec<konnect_ipc::IpcGraphicDefinition>> {
     use konnect_ipc::IpcGraphicDefinition as Graphic;
     let footprint = konnect_sexp::parse_sexp(source)?;
     let mut graphics = Vec::new();
@@ -493,7 +506,12 @@ pub(crate) fn extract_graphic_definitions(
             stroke_width_mm: text_stroke_width(text),
         });
     }
-    for property in footprint.find_all("property") {
+    let properties = if include_properties {
+        footprint.find_all("property")
+    } else {
+        Vec::new()
+    };
+    for property in properties {
         let name = property.get(1).and_then(konnect_sexp::SexpNode::as_str);
         // Reference and Value travel as first-class fields; hidden built-ins
         // (Footprint, Datasheet, …) are not drawn.
--- a/crates/konnect-core/src/tools/pcb_footprint_update.rs
+++ b/crates/konnect-core/src/tools/pcb_footprint_update.rs
@@ -20,6 +20,7 @@ struct LibraryFootprint {
     attributes: kiapi::board::types::FootprintAttributes,
     datasheet: Option<String>,
     description_field: Option<String>,
+    properties: Vec<kiapi::board::types::Field>,
     pads: Vec<konnect_ipc::IpcPadDefinition>,
     graphics: Vec<konnect_ipc::IpcGraphicDefinition>,
     models: Vec<kiapi::board::types::Footprint3DModel>,
@@ -733,11 +734,14 @@ fn parse_library_footprint(library_id: &str, source: &str) -> Result<LibraryFoot
 
     validate_supported_children(&root)?;
     let pads = super::pcb_components::extract_pad_definitions(source)?;
-    let graphics = super::pcb_components::extract_graphic_definitions(source)?;
+    // Custom properties travel as typed Field items below. Treating visible
+    // properties as generic graphics as well would duplicate their text.
+    let graphics = super::pcb_components::extract_graphic_definitions_without_properties(source)?;
     let models = parse_models(&root)?;
     let attributes = parse_attributes(&root)?;
     let datasheet = property_value(&root, "Datasheet");
     let description_field = property_value(&root, "Description");
+    let properties = parse_custom_properties(&root)?;
     let definition = kiapi::board::types::Footprint {
         id: Some(kiapi::common::types::LibraryIdentifier {
             library_nickname: library_nickname.to_string(),
@@ -757,6 +761,7 @@ fn parse_library_footprint(library_id: &str, source: &str) -> Result<LibraryFoot
         attributes,
         datasheet,
         description_field,
+        properties,
         pads,
         graphics,
         models,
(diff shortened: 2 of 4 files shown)


neusse (commented):
Reviewed current head `6652ec6de1e87c04a7159542e6f8570d2eae0ea7` and the #373 acceptance criteria.

The unique implementation looks sound: generated/custom footprint properties are carried through typed `Field` items, mandatory fields remain first-class, instance-only properties are preserved, same-name library fields refresh deterministically, ambiguous duplicates and unrepresentable clauses refuse before commit, transforms are covered, and apply-to-noop convergence is tested with a KiCad-authored fixture. This is a behavior fix with no tool-schema change.

Queue position: **after #369**. Please do not rebase this branch yet. Once #369 is merged, update #374 directly onto the then-current `main` and rerun the full unfiltered gate. Waiting means one update rather than rebasing it once now and again after #369.

The deterministic planner/mock-IPC evidence is strong. A live KiCad refresh of a stock footprint carrying `KiLib_Generator` plus a custom property, followed by save and readback, would be valuable proportional evidence. I do not consider that missing from #373 itself, but it should be included in the real-KiCad pre-release validation for the release that ships this change.

neusse:
This is now the active PR in the dependency queue after #369. The current head (`6652ec6de1e87c04a7159542e6f8570d2eae0ea7`) is mergeable, but it is 11 commits behind current `main`, and its green CI results were produced against that older base. Please update this branch onto current `main` and rerun the complete, unfiltered contributor gate. Once the updated exact head is green, we will review it for merge approval. We will hold #376 and the later stacked PRs until #374 is complete and merged so each successor only needs to be restacked once and we do not create unnecessary rebase work.

neusse (changes requested):
Reviewing the current rebased head `edc6f453e2675c1986af49bf9f3153664853bd08`. The branch is clean, all ten hosted checks pass, and the complete local fmt/clippy/workspace-test/doc-test gate passes. I nevertheless found two reproducible gaps against #373 that need correction before merge.

## Spec

1. **Mandatory properties bypass the new lossless validation.** `parse_library_footprint` reads `Datasheet` and `Description` through `property_value`, while `parse_custom_properties` excludes `Reference`, `Value`, `Datasheet`, and `Description`. `property_value` takes the first matching value and ignores the remainder of the property node. A temporary regression probe using the KiCad-authored fixture added `(unlocked yes)` to `Datasheet`; parsing incorrectly returned `Ok`. Duplicate mandatory names would likewise select the first value. This conflicts with #373's requirement to validate every property clause and refuse with the property and clause named before `BeginCommit`. Please validate all four mandatory property nodes losslessly while continuing to store them in their existing first-class fields, and reject duplicate mandatory names.

2. **Repeated scalar clauses are silently overwritten.** `hide` and `knockout` assign on every occurrence; `bold`, `italic`, and `line_spacing` do the same inside `effects/font`. A second temporary probe added conflicting `(hide yes)` and `(hide no)` clauses to `AssemblyVendor`; parsing again returned `Ok`, with the last value winning. Please track whether each scalar clause has already appeared and reject repeats with the property and clause named.

## Standards/evidence

- `GOVERNANCE.md` requires neutering every new guard and confirming its regression test fails. Please record that mutation evidence for the added rejection paths.
- The PR description still reports the pre-rebase workspace-test failure and skipped-test workaround even though the current unfiltered gate is green. Please update it for this head and include the compatibility and risk/rollback notes requested by `CONTRIBUTING.md`.

The typed-field design, instance-only preservation, transforms, convergence coverage, and focused scope otherwise look sound.

dubesinhower:
@neusse The requested changes from your review of edc6f45 are addressed on current head 9d6e8bc: mandatory properties now use the lossless parser with duplicate-name refusal; repeated scalar clauses refuse; KiCad-authored regression coverage and guard-neutering evidence are recorded; and the PR description now carries the current compatibility, rollback, full local-gate, and live KiCad apply/save/readback evidence. All ten hosted checks are green on this exact head. Could you please re-review #374 when convenient?

neusse (approved):
Re-reviewed exact head `9d6e8bce1fde7803073859a598d00f1550a9d32a` after the requested changes. The mandatory-property and repeated-scalar gaps are fixed, mutation evidence and compatibility/risk notes are recorded, the stock KiCad 10 fixture matches the installed library source, all ten hosted checks pass, and the focused plus full local gates pass. No remaining standards or #373 specification findings.

PR #376, opened 2026-08-30
fix(schematic): respect KiCad editor locks

### LINKED ISSUE #375: Block schematic writes while KiCad owns a sibling lock
## Problem

Konnect's schematic mutation tools use revision-checked atomic replacement, but their shared write paths do not inspect KiCad's sibling `~<schematic>.lck` file. A mutation can therefore succeed while Eeschema owns the document; the next GUI save can overwrite Konnect's change from newer unsaved editor state.

## Reproduction

Reproduced on untouched `mixelpixx/Konnect:main` at `428c1bd`:

1. Copy an upstream-owned KiCad schematic fixture to a temporary `repro.kicad_sch`.
2. Add sibling `~repro.kicad_sch.lck` using KiCad's JSON shape: `{"username":"konnect-repro","hostname":"konnect-repro-host"}`.
3. Load `sch_wiring` and call `add_wire` for the temporary schematic.

Observed: `add_wire` reports success and the schematic SHA-256 changes while the lock remains present.

Expected: a typed `conflict` is returned before any target, scratch file, or transaction journal is changed.

## Proposed safety policy

KiCad 10's `LOCKFILE` writes only `username` and `hostname`; it records no PID, start time, or document-instance token. Konnect therefore cannot reliably prove that a same-host or remote lock is stale.

- Gate every `.kicad_sch` mutation at the shared writer boundary.
- Treat any present sibling lock as a conflict, including well-formed, stale-looking, foreign-host, empty, and malformed files. Uncertainty must not mutate.
- Preserve unlocked closed-file writes and the existing exact-revision checks.
- For multi-file transactions, check every target after acquiring Konnect's cooperative locks and before persisting the write-ahead journal, then check again at the final target-write boundary.
- Keep the response additive by reusing the existing public `conflict` kind and reporting the locked schematic path.

## Acceptance tests

- no KiCad lock: the existing atomic write succeeds;
- well-formed live-owner lock: typed conflict, no target change;
- stale-looking/foreign-owner lock: typed conflict, no target change;
- malformed lock: typed conflict, no target change;
- transaction lock conflict: no target and no journal change;
- file changed after planning: existing revision conflict remains intact.

dubesinhower:
## Summary

- block every shared `.kicad_sch` write when KiCad's sibling `~<name>.kicad_sch.lck` exists
- fail closed for valid, foreign-host, empty, malformed, or otherwise unreadable lock entries without modifying the document, lock, scratch files, or a new transaction journal
- surface the refusal through the existing structured `conflict` result while leaving read-only tools available
- recheck at the final atomic-write boundary and protect transaction recovery as well as new commits

## Root cause and design

On untouched `main`, the public `add_wire` tool succeeded and changed a schematic even while its KiCad editor lock existed. KiCad's lock format contains only `username` and `hostname`; it has no PID, process-start time, or document token that can prove liveness or staleness. Konnect therefore must not infer that a stale-looking lock is safe.

The guard lives in the shared atomic writer, checks before scratch-file creation and immediately before replacement, and is also called before transaction journal creation/recovery work. Lock contents are intentionally not parsed and locks are never removed automatically. See KiCad's [LOCKFILE source](https://docs.kicad.org/doxygen/lockfile_8h_source.html).

## Compatibility

This is additive. It changes no tool names, parameters, schemas, or success payloads. Locked schematic writes use the existing `conflict` error kind and include the blocked path.

## Verification

Verified on exact rebased head `96130b7967d1ec90cedd50346a355c3be7769183`:

- `cargo test --locked -p konnect-sexp kicad_schematic_lock -- --nocapture` — pass (3 tests)
- `cargo test --locked -p konnect-sexp editor_lock -- --nocapture` — pass (2 tests)
- `cargo test --locked -p konnect-core a_kicad_schematic_lock_is_a_typed_conflict --lib -- --nocapture` — pass (1 test)
- `cargo test --workspace --locked --lib --tests` — pass; environment-dependent GUI/Java/KiCad tests remain explicitly ignored
- hosted CI — all 10 required checks pass on this head across Ubuntu, Windows, macOS, clippy, formatting, packaging, licenses, Nix, Python, and the schematic viewer

Real KiCad 10 lifecycle probe on a disposable schematic copy:

- Eeschema created the expected sibling `~LockProbe.kicad_sch.lck` (43 bytes).
- The exact PR binary's public `add_wire` call returned `isError: true` with structured error kind `conflict`.
- SHA-256 of the schematic was identical before and after the refusal, the KiCad lock remained untouched, and no scratch or transaction-journal sibling appeared.
- After the hidden editor process was confirmed stopped, its intentionally surviving stale lock was removed explicitly; the identical `add_wire` call then succeeded, changed the disposable copy, and left no scratch or journal sibling.

## Risk and rollback

The policy intentionally favors preserving unsaved editor state over automatic stale-lock recovery. A confirmed stale lock requires explicit operator action after checking that no editor owns the schematic. Reverting this commit restores the prior behavior without a migration.

Closes #375

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -192,6 +192,11 @@ Konnect/
 - Cooperative lock files live under `KONNECT_STATE_DIR/locks` when that
   absolute override is set, otherwise under the platform local-data directory
   (`konnect/locks`). Reads never create files in the KiCad project.
+- Schematic writes also refuse while KiCad's sibling `~<name>.kicad_sch.lck`
+  exists. KiCad records only a username and hostname, so Konnect cannot prove
+  that a same-host or remote lock is stale; valid, foreign, empty, and malformed
+  locks all fail closed. The check runs before a transaction journal is created
+  and again at the final target-write boundary.
 - Multi-file schematic changes use project-local
   `.konnect-transaction-*.json` write-ahead journals. These journals contain
   complete before/after images and must be treated as sensitive project data.
--- a/crates/konnect-core/src/mcp/handler.rs
+++ b/crates/konnect-core/src/mcp/handler.rs
@@ -307,6 +307,24 @@ impl McpHandler {
                         Some("invalid_argument".to_string()),
                     )
                 }
+                Err(e) if kicad_editor_locked_path(&e).is_some() => {
+                    let path = kicad_editor_locked_path(&e)
+                        .expect("guard matched")
+                        .display()
+                        .to_string();
+                    (
+                        CallToolResult::error_kind(
+                            ToolErrorKind::Conflict {
+                                paths: vec![path.clone()],
+                            },
+                            format!(
+                                "Schematic '{path}' has a KiCad editor lock. Close Eeschema, or resolve a stale lock only after confirming no editor owns the file, then retry."
+                            ),
+                        ),
+                        CallStatus::Error,
+                        Some("conflict".to_string()),
+                    )
+                }
                 Err(e) => {
                     warn!(tool = %name, error = %e, "tool handler returned anyhow::Error");
                     let kind = ToolErrorKind::HandlerError {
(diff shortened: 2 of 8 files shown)


neusse (commented):
Reviewed current head `c66920145f6763d0329179bb2be6b16c9b95798f` against #375.

The implementation matches the requested safety policy: the guard is centralized at the shared schematic write boundary, any present or unreadable KiCad sibling lock fails closed, lock contents are not guessed or removed, transactions check before journal creation and again at the final write boundary, existing revision checks remain intact, and the MCP response reuses the structured `conflict` kind. The positive and no-write cases are covered. I found no separate standards or specification defect in the unique change.

Queue position: **after #374**. Please wait for #374 to merge, then update #376 once onto current `main` and run the complete unfiltered gate. This branch also overlaps intervening documentation/writer history from its older base; the merge simulation is clean, but updating before final review is still required so CI proves the actual merge-time tree.

neusse (approved):
Final review of exact head 96130b7967d1ec90cedd50346a355c3be7769183 against #375:\n\n- Specification review: no missing, partial, incorrect, or out-of-scope behavior.\n- Standards review: no hard findings; only a non-blocking crate-private naming judgement.\n- The rebased patch is unchanged from the previously reviewed commit.\n- Guard-neutralization evidence: replacing the lock guard with a no-op made both transaction lock regressions fail and made the public MCP structured-conflict regression fail.\n- All ten required hosted checks pass on this head.\n\nNo blocking findings.

Issue #18, opened 2026-07-10
"KiCAD IPC socket path not configured"


BR-Tavares:
"KiCAD IPC socket path not configured"
Quick diagnostic & setup guide — Konnect (KiCad ↔ MCP)
When this error appears
Any Konnect tool that needs to talk to KiCad live (save_project, live PCB editing, check_kicad_ui, etc.) returns:

KiCAD IPC socket path not configured. Either launch this plugin from KiCAD
(sets KICAD_API_SOCKET), or set the ipc_socket_path in settings.
— even with KiCad open and the API enabled.

Correct path, step by step
1 → Open KiCad normally.

2 → Go to Edit → Preferences → Plugins (IPC API section) and check "Enable KiCad API". Confirm a line like this appears:

Listening on ipc://C:\Users\<your_user>\AppData\Local\Temp\kicad\api.sock
👉 Copy that whole address (starting with ipc://) — it is unique to this machine/user.

3 → In KiCad, open Tools → External Plugins (or the plugin toolbar icon) and click the Konnect icon. This opens the plugin's settings dialog.

⚠️ Known UI bug: this window may open smaller than intended, cutting the Save/OK button off-screen (a Windows DPI/scaling issue). If you can't see the button:

Try resizing the window by dragging its bottom edge, or
Use Tab to cycle through fields until the button is focused (even if not visible), then press Enter or Space to confirm.
4 → Paste the address from step 2 into the IPC socket path field. Save.

5 → Confirm your AI client (Claude Code, Claude Desktop, etc.) has the konnect MCP server registered in its own MCP config (project .mcp.json, or claude_desktop_config.json), pointing to the konnect.exe binary and the plugin's settings.json. This registration is separate and does not happen automatically just from configuring the plugin inside KiCad.

6 → Restart the AI client session (close and reopen Claude Code / Claude Desktop) so it spawns a fresh Konnect process that reads the newly saved ipc_socket_path.

7 → Verify: have the AI call open_project (or equivalent). Expected result:

{ "kicad_ui_running": true, "message": "KiCAD is running and IPC is available." }
One-sentence summary
There are two separate configurations that both need to be correct at the same time: (a) the ipc_socket_path saved in the plugin's settings.json inside KiCad, and (b) the konnect server registration in the AI client's own MCP config. Neither happens automatically on its own — and KiCad's settings dialog has a visual bug that can hide the save button.

[...]

PR #382, opened 2026-08-30
fix(ipc): auto-detect KiCad's IPC socket when the env var is unset

### LINKED ISSUE #424: feat(ipc): discover a live same-user KiCad socket at startup
## Problem

A standalone Konnect server does not inherit `KICAD_API_SOCKET` from KiCad. On Unix, a supported KiCad endpoint may exist at a predictable local socket, but Konnect currently requires manual configuration.

PR #382 implements bounded startup-time discovery. This issue defines the deliberately narrow contract that PR may close; dynamic discovery after startup is separate.

## Accepted scope

At server startup:

- preserve precedence: explicit configuration, then a nonblank `KICAD_API_SOCKET`, then a detected platform candidate, otherwise unresolved;
- on Linux/macOS, adopt a candidate only when it is the expected local-user socket and a bounded connection proves that something is listening;
- reject stale paths, non-sockets, foreign-owned sockets, and probes that cannot complete within the bound;
- leave Windows behavior unchanged until the NNG named-pipe mapping can be probed and tested;
- report/log the actual selected source without exposing credentials;
- apply the same resolved configuration semantics to the executable and KiCad FFI entrypoints.

## Explicit limitation

Discovery in this issue runs once at startup. Documentation must state that KiCad must already be running with its API enabled before Konnect starts, or the Konnect client/server must be restarted. It must not promise configuration-free operation for the usual client-first/KiCad-second order.

Dynamic re-probing is tracked separately and is not silently implied here.

## Evidence responsibility

Cross-platform compatibility is a project responsibility, not a requirement that one contributor personally own every supported environment.

- The contributor supplies deterministic tests, required hosted CI, and real-environment evidence from an affected platform they can reasonably access.
- Maintainers, original reporters, and community testers supply additional operating-system and KiCad-version observations.
- A missing secondary-platform observation is recorded as validation debt. It does not block a focused, reversible PR when required CI is green, the behavior is deterministically covered, and the missing observation is not needed to establish a material safety property.
- Stale paths, foreign ownership, precedence, and bounded probing are better established by deterministic tests and negative controls than by repeating artificial manual setups on every OS.

## Acceptance

- Deterministic tests cover precedence, blank env values, a live same-user socket, stale path, non-socket, foreign ownership, and a bounded unresponsive/saturated probe.
- The ownership and timeout guards are each neutralized once and their tests fail.
- Production diagnostics accurately describe their coverage; either all direct production IPC entrypoints use the shared diagnostic seam or the documentation narrows the claim.
- FFI diagnostics are emitted through an initialized logging path or are not claimed.
- One real Linux **or** macOS run records the KiCad version, absence of explicit `ipc_address` and `KICAD_API_SOCKET`, selected endpoint and source, and a successful IPC response.
- Hosted Linux, macOS, and Windows CI passes. Windows tests prove this PR does not invent Unix-style discovery or otherwise change Windows behavior.
- The other Unix platform is a visible post-merge compatibility-validation item for maintainers/reporters/community testers; its absence alone does not block this PR.
- The PR is reconstructed on current main, passes all required checks, and documents the startup-order limitation prominently.

PR #382 may use `Closes #424` when the deterministic acceptance items, required CI, and one real Unix observation are satisfied. Any remaining secondary-platform validation must stay visible in the completion comment and release checklist rather than being represented as completed.

pauliuszaleckas:
`KICAD_API_SOCKET` is exported only to plugins KiCad launches itself. A standalone
server started by an MCP client therefore sees an empty address, and every IPC call
falls back to file editing with nothing said about why — on Linux the socket sits at
the entirely predictable `/tmp/kicad/api.sock` and had to be pasted into `settings.json`
by hand.

This resolves the address as **config > `KICAD_API_SOCKET` > the platform default**
(`<temp dir>/kicad/api.sock`, plus `/tmp/kicad/api.sock` on macOS where the temp dir
resolves under `/var/folders`), and reports which of the three answered.

Not filed as an issue; #18 and #39 are the closed issues this friction produced, both
resolved at the time by documentation and by making the env var apply on every load
path.

### A candidate counts only when something is listening

KiCad does not unlink `api.sock` on exit, so a socket from a session two days ago sits
exactly where a live one would. One `connect()` separates them, with two guards:

- **Bounded.** `connect()` on an `AF_UNIX` stream blocks while the listener's backlog
  is full — it does not fail fast the way TCP does — and this runs from
  `Config::load_resolved` *before* tracing is initialized. Unbounded, a wedged KiCad
  would hang the server before one line reached stderr, and the client would just see a
  process that never answers `initialize`. Capped at 250 ms.
- **Owner-checked.** A detected socket is adopted as the board endpoint unread, so only
  one owned by the user running Konnect qualifies. The shared `/tmp` candidate sits in a
  world-writable directory, and it is macOS-only for the same reason — on Linux the temp
  dir already *is* `/tmp` unless `TMPDIR` says otherwise, so the fallback bought no
  coverage.

### Windows gains no auto-detection, deliberately

NNG's `ipc://` is a named pipe: there is nothing at the path to connect to, so
`is_listening` answers `false` and detection finds nothing. Windows keeps exactly the
setup it had before — config, or `KICAD_API_SOCKET`, or nothing.

Taking the default on trust there was the alternative, and it is worse: it makes
`IpcAddressSource::Unresolved` unreachable and retires the two messages a Windows user
needs most — the "socket path not configured" error carrying the settings-dialog steps,
and the startup warning listing the candidates — replacing them with a dial failure
against an address nobody chose. Probing the pipe properly (`CreateFileW` against the
name NNG derives) is the real answer and is left for a change that can be tested on
Windows.

### Diagnostics

Startup logs the resolved address and its source, and warns with the candidate list when
there is none. An IPC failure that never reached KiCad now warns with the address it
tried; KiCad-*rejected* calls stay silent, since they fail closed and report themselves.

Wiring that wanted one seam per failure shape. `with_ipc_classified` was already one;
`pcb_export` still had a private `with_ipc` — the last of four per-toolset copies — and
it becomes `tools::with_ipc` beside the classified gate, so the next toolset gets the
warning for free.

Also fixed on the way: `ffi.rs` resolved nothing at all, so a cdylib loaded by KiCad with
`--config` never picked up `KICAD_API_SOCKET` — the gap #39 closed for `main.rs` only.
Both entry points now load through one `Config::load_resolved`.

### New dependencies

`socket2` and `libc`, both `[target.'cfg(unix)'.dependencies]` on `konnect-ipc` — for
the bounded connect and for `geteuid` respectively. Both were already in the build graph
transitively.

### Known limitation

Detection runs once, at startup. The usual MCP launch order is client → server → *then*
the user opens KiCad, and in that order the address stays unresolved for the session.
`launch_kicad_ui(wait_ready)` snapshots the empty address before spawning KiCad, so it
can only time out. Covering it means re-probing at the read site while the address is
empty — a `ToolContext` accessor rather than a plain field — which touches ~8 production
call sites and needs a detection opt-out threaded through the test contexts (50 of them
construct `ipc_address: String::new()` precisely to mean "no live KiCad"). That is worth
its own PR rather than this one.

### Verification

Verified end to end: with no config and no env var, `check_kicad_ui` returns
`ipc_responsive: true` against a running KiCad.

Local gate: `--lib --tests` 1385 passed, `--doc` 8 passed, `clippy -D warnings` clean,
`fmt --check` clean. One unrelated failure —
`konnect-sexp`'s `every_installed_demo_board_scans_losslessly`, which skips 2 zones on
the `royalblue54L_feather` demo board — reproduces identically on `main` and is
untouched by this change. The schematic-viewer suite could not run locally
(`libsoup-3.0` / `webkit2gtk-4.1` absent); it depends only on
`konnect-schematic-editor`, which this does not touch, and CI's viewer job covers it.

---

Closes #424 under its risk-proportionate evidence contract. The reported real-Unix end-to-end run, deterministic safety tests, and required hosted CI own the merge gate. The second Unix platform remains visible post-merge compatibility validation for maintainers, reporters, and community testers. Dynamic client-first/KiCad-second discovery is #425 and is not attempted here.


neusse:
I completed a two-axis review of head 7eba8ff445dd933f25c54d316ac435bd1080cd71 against current upstream/main 2b49179470e4c22abff6917f0ad44dda9492c5db.

The direction is useful, but this head is not ready to merge.

### Required changes

1. Link this PR to #424 and use Closes #424 only when that issue's full acceptance checklist is satisfied. Dynamic client-first/KiCad-second discovery is now separated into #425.
2. Reconstruct or rebase on current main and rerun all ten checks. The existing checks are from August 30 against a base 41 commits behind current main.
3. Correct TROUBLESHOOTING.md. The statement that no configuration is needed with KiCad open must say KiCad must already be open when Konnect starts; this implementation probes only once.
4. Add decisive tests for the two safety promises:
   - reject a foreign-owned Unix socket;
   - prove a saturated/unresponsive listener returns within the bound.
   Neutralize each guard and record the failing regression.
5. Narrow or complete the diagnostics claim. project.rs, verification.rs, and some routing paths construct KiCadIpcClient directly rather than using the shared with_ipc seam; check_kicad_ui converts ping failure to false without the promised warning. FFI calls ipc_source.log() without establishing that a tracing subscriber exists.
6. Keep the verified behavior already present: config > nonblank environment > detected candidate, same-user Unix selection, stale-socket rejection, derived source reporting, and deliberately unchanged Windows behavior.

### Review result

Standards: four findings—stale base/checks, missing issue accounting, missing ownership-guard evidence, and unproved current-main compatibility.

Spec: four findings—the documentation overpromises beyond startup-only behavior, diagnostics are incomplete, timeout/foreign-owner evidence is missing, and the PR currently mixes work without an accepted issue boundary.

No material Fowler code-smell problem was found; the shared IPC helper and isolated socket module are reasonable. Once the focused contract above is satisfied, this can be re-reviewed as the foundation before #421 is adapted.

pauliuszaleckas:
Rebuilt on current main (`d73adfa`) and pushed. Taking the required changes in order, including the one I have *not* satisfied.

**1. Issue linkage.** The PR body now says *Tracked by #424*, deliberately not `Closes` — see point 5 for what is still missing from that checklist. #425 (dynamic re-probing) is not attempted here.

**2. Rebase.** Done; the checks now run against a head built on current main.

**3. TROUBLESHOOTING.md corrected.** The overpromise is gone. It now leads with the fact that resolution happens **at startup, once**, and spells out why that is easy to get wrong: an MCP client normally launches the Konnect server *before* the user opens KiCad, and such a server stays unresolved for its whole life no matter what is opened later. Start KiCad first, restart the server once KiCad is up, or set `ipc_address`, which never depends on ordering. Windows is called out separately as detecting nothing either way.

**4. Decisive tests for both safety promises.**

*Foreign ownership* — two tests, neither needing a second account:
- `/etc/passwd` is a genuinely foreign-owned path on every Linux and macOS, so `is_owned_by_us` is asserted against real foreign ownership (skipped when running as root, where nothing is foreign).
- The gate is also exercised over a socket that **is** live: `is_listening` now takes the ownership rule as a parameter, so a listening endpoint whose owner check is made to fail is proved not to be adopted. That is the shared-`/tmp` squatter, without needing a squatter.

*The bound* — this one changed shape once I measured it, and the first version of the test was not decisive. Facts, on Linux:
- a **blocking** `connect()` to an `AF_UNIX` socket whose `listen(1)` backlog nothing drains never returns (measured — the experiment had to be killed);
- `connect_timeout` connects non-blocking, and the kernel answers a full backlog immediately with `POLLHUP` — ~16 µs.

So the guard that prevents a startup hang is *using a bounded, non-blocking connect at all*, not the specific 250 ms. The test saturates a `listen(1)` queue and runs the probe **on a worker thread**, failing if no answer arrives within a flat 2 s — a literal, not `PROBE_TIMEOUT * n`, which would move with the value under test. Neutralizing the guard therefore fails the test in 2.00 s instead of hanging it. The `PROBE_TIMEOUT` doc comment now records the measured behaviour rather than asserting it.

*Negative controls*, each guard neutralized once:

| Neutralized | Result |
|---|---|
| ownership gate removed | `a_live_socket_that_fails_the_ownership_check_is_not_detected` fails |
| `connect_timeout` → blocking `connect` | `a_saturated_listener_gives_up_within_the_bound` fails in 2.00 s |
| existence counts as listening | 2 tests fail (stale socket + saturated listener) |
| shared `/tmp` candidate on every platform | `the_shared_tmp_fallback_is_macos_only` fails |

(The first attempt at the timeout control — raising `PROBE_TIMEOUT` to 5 s — passed, which is what exposed that the old assertion never reached the timeout path. Recording it because it is the reason the test was rewritten.)

**5. Diagnostics claim — narrowed, and two real gaps fixed.**

- **FFI**: you were right that `ipc_source.log()` had no subscriber. `main.rs` installs one before reporting; `ffi.rs` installed none, so the report — including the warning naming every candidate probed — went nowhere. It now installs one with `try_init`, which leaves a subscriber the host already installed alone.
- **`check_kicad_ui`**: it converts a ping failure to `false`, and `ping()`'s existing warning did not name the endpoint. It does now — that is the one IPC failure that never reaches a caller as an error, so nothing else recorded which address went unheard.
- **The claim itself is narrowed** rather than overstated. The guide now says the warning covers the shared IPC helper — every tool that falls back to a file — and explicitly says it is *not* a global "IPC failed" log: the health tools and `get_project_info` dial KiCad directly and report the outcome in their own response (`ipc_responsive`, `connected`) rather than falling back to anything, so those fields are what to read. A KiCad that answered and *refused* is not warned about anywhere; that is a tool error and says so. (The remaining direct construction in `pcb_routing.rs` is inside an `#[ignore]`d live test, not a production path.)

**What I have not done, and cannot from here.** #424 asks for real Linux and macOS evidence recording the selected endpoint and source, and Windows tests proving no behavioural regression. This machine has no display and no Xvfb, so I cannot bring up a KiCad GUI to capture a live resolution, and I have no macOS or Windows host. Everything above is unit-level and reasoned from KiCad's own defaults. If that evidence is required before merge, tell me which route you prefer — I can capture it on a machine with a display, or add a GUI KiCad session to `e2e-kicad.yml` so it is captured in CI on every run.

**6. Verified behaviour preserved:** precedence (config > non-blank env > detected candidate), same-user Unix selection, stale-socket rejection, derived source reporting, and Windows unchanged — all still covered by their tests.

Gate on the rebased head: workspace tests, doctests, `clippy -D warnings`, `fmt --check`, plus `crates/schematic-viewer` — all clean.

neusse:
Maintainer direction on the remaining evidence: please use the real-machine capture route you offered. Do not expand this focused socket-discovery PR into a new multi-platform GUI-CI framework.

Please provide:

1. Linux: start KiCad with IPC enabled, start Konnect without `ipc_address` or `KICAD_API_SOCKET`, and record the selected endpoint plus the reported source.
2. macOS: repeat the same check, including which platform candidate was selected.
3. Windows: record that this PR leaves discovery behavior unchanged and does not fabricate a Unix-style endpoint.

For each platform, include the KiCad version, Konnect command/config conditions, the observed diagnostic or `get_installation_info` evidence, and whether a stale/non-listening candidate was rejected. If a platform genuinely cannot be exercised, mark that item `BLOCKED` rather than substituting reasoning for observation.

Please collect the evidence before refreshing the branch again. Recent merges have moved `main`; once the evidence is attached and the acceptance checklist is complete, rebuild once on then-current `main`, change the issue linkage to `Closes #424`, and rerun all ten checks. That avoids spending another CI cycle before the only remaining substantive gate is resolved.

#421 remains intentionally held until this PR settles.

neusse:
Correction to my previous maintainer direction: requiring one contributor to produce live Linux, macOS, and Windows captures was too broad. Most contributors reasonably have one development environment; cross-platform validation belongs to the project as a whole.

The acceptance contract in #424 is now risk-proportionate:

- deterministic tests and negative controls own stale sockets, ownership, precedence, and bounded-connect safety;
- required hosted CI owns Linux/macOS/Windows regression coverage;
- one real Linux **or** macOS discovery run is sufficient for this focused PR;
- the other Unix platform becomes explicit post-merge compatibility validation for maintainers, reporters, or community testers;
- Windows needs CI proof that behavior remains unchanged, not a manual Unix-discovery demonstration.

The PR body already reports a successful end-to-end run with no config and no environment variable where `check_kicad_ui` returned `ipc_responsive: true` against running KiCad. I accept that as the required real-Unix observation. If the exact KiCad version and selected-source log are still readily available, adding them would improve the record, but recreating a discarded session is not a merge blocker.

Please now do only the work that actually belongs to the contributor: rebuild once on current `main`, change the linkage to `Closes #424`, preserve the focused behavior and tests, and run all ten required checks on the final head. Do not add GUI CI or seek access to operating systems you do not have.

After merge, maintainers will ask the original Linux reporter from #39 to confirm the manual socket configuration is no longer needed and will recruit a macOS user for the remaining compatibility observation. Any result becomes a bug report or follow-up fix instead of holding this contribution indefinitely.

neusse:
I handled the remaining maintainer-owned mechanics: the PR now closes #424 under the revised evidence contract, and GitHub has updated the branch with current `main`. Fresh required checks should start on the new head.

No further contributor action is requested unless CI exposes a real regression. Once the checks finish, maintainers will review the exact new head and move it forward.

neusse:
Maintainer follow-up commit `cb4cad9` resolves the four deterministic review gaps without asking the contributor for another environment:

- startup diagnostics now reuse the existing endpoint redaction and have a direct regression test proving credentials, query values, and fragments do not reach the log message;
- whitespace-only `KICAD_API_SOCKET` values now fall through to discovery;
- Unix tests now prove an owned regular file is rejected independently of the ownership guard;
- Windows CI now exercises the production non-Unix probe instead of a fake closure.

Focused Windows tests pass locally. I also neutralized the whitespace, redaction, and Windows no-probe guards individually and confirmed each regression test fails before restoring the implementation. Linux owns the Unix-only regular-file test in hosted CI.

The `cdbc5f4` merge-from-main commit was generated by my maintainer branch refresh, not by the contributor. It adds no unique diff and I am treating that mechanical history noise as non-blocking rather than force-rewriting the contributor's branch.

Fresh required checks are now the remaining gate on exact head `cb4cad9`.

neusse:
Final-review follow-up is pushed at `37b9a25` and includes current `main`.

The remaining credential-exposure path is now closed across all diagnostics introduced or widened by this PR:

- IPC debug destination logging uses the shared redactor;
- dial errors expose only the redacted endpoint;
- ping warnings redact the endpoint;
- fallback warnings redact the structured address and log only the safe typed root cause;
- endpoint redaction now lives in `konnect-ipc`, so both the client and higher layers use one implementation.

A direct dial-failure regression test uses an endpoint containing user/password, query token, and fragment. It proves none reaches the returned diagnostic. Neutralizing the dial redaction made that test fail with the full secret-bearing endpoint, then it passed after restoration.

Focused tests, formatting, and Clippy pass locally. Fresh required CI on exact head `37b9a25` is now the remaining gate. No additional contributor action is requested unless CI finds a regression.

neusse:
Final maintainer review complete on exact head `37b9a2546043a074584557338f75b960b83ff23f`.

Acceptance accounting for #424:

- precedence is explicit config, then nonblank `KICAD_API_SOCKET`, then a detected candidate, otherwise unresolved;
- Unix discovery requires a live same-user socket and deterministically rejects stale paths, owned non-sockets, foreign ownership, and bounded/unresponsive probes;
- startup, dial-error, ping, and fallback diagnostics redact credentials, query values, and fragments;
- executable and FFI startup paths share resolved configuration semantics and FFI initializes logging;
- Windows production tests prove no Unix-style endpoint is fabricated;
- documentation prominently states discovery happens once at startup and does not cover client-first/KiCad-second ordering;
- the reported real-Unix run satisfies the merge observation; the second Unix platform remains visible compatibility-validation debt under the newly merged governance rule;
- `Closes #424` resolves correctly;
- the branch includes current `main` and all ten required checks passed on this exact head.

Standards review: no remaining blocking finding. The maintainer-generated merge commits are mechanical branch-refresh history with no unique product diff and are accepted without rewriting the contributor's branch.

Marked `status:ready-to-merge`. Merge remains subject to the local explicit approval gate for this exact head.

PR #381, opened 2026-08-30
feat(runtime): report installation provenance

dubesinhower:
Closes #380.

## Root cause

Konnect's package semver was the only build identity. Release builds did not embed the exact checkout, client `status` inspected static guidance rather than the serving process, and the runtime had no process-scoped provenance collector or conservative on-disk-versus-running comparison.

## Implementation

- add an always-visible, zero-argument `get_installation_info` MCP tool;
- report the serving version, embedded commit and source when available, build profile/target, and current executable path;
- classify KiCad PCM only from the exact public plugin identifier and executable runtime type, otherwise return `unknown`;
- probe the executable at the serving path and the resolved `kicad-cli` with read-only, kill-on-drop, five-second-bounded `--version` calls;
- report `newer_than_running` only when both values are comparable stable versions;
- return the resolved IPC endpoint with credentials, query data, and fragment data redacted;
- return Windows, macOS, Linux, and PCM-specific restart guidance;
- embed Git HEAD for source builds without invoking Git, and embed `${{ github.sha }}` in release artifacts;
- update public docs and generated tool counts (221 registered + 7 meta = 228 total; baseline 21).

The diagnostic does not mutate configuration or installation state, scan processes, dump environment data, or expose subprocess error text.

## Dependency and branch shape

- based directly on `main` after merged #379 (`e7aeaa6a1b353fa758c5246b58a18899275d13ef`);
- exactly two unique commits: the provenance feature and its stdio portability follow-up;
- the previous history-only upstream merge commit has been removed;
- `git range-diff` reports both reconstructed commits as patch-identical to the reviewed commits from the previous head.

## Regression coverage

- five unit tests cover exact PCM-manifest classification, endpoint redaction, conservative version comparison, embedded commit shape, and platform guidance;
- a real stdio MCP test covers always-visible registration, serving-process output, on-disk probing, newer comparison, and endpoint redaction;
- the endpoint-redaction guard was deliberately neutralized during development and the stdio regression failed on its synthetic query/fragment values, then passed after restoration;
- documentation registry/count contracts and asset references pass;
- the final release binary reports `konnect 0.11.0` and contains exact branch head `a2c9e20bc643479f7aaf8ecd066eb0529e9711de`.

## Verification on the final head

- `cargo xtask fix-doc-counts` — pass; documentation already current, no rewrite required
- `cargo fmt --all -- --check` — pass
- `cargo clippy --workspace --locked --all-targets -- -D warnings` — pass
[...]

Issue #385, opened 2026-08-30
Use KiCad board locks as cold-start stale-file evidence


dubesinhower:
## Context

#361 closed #240 with process-lifetime board memory and explicitly documented its cold-start limitation: after Konnect restarts, an unreachable transport cannot tell whether KiCad previously crashed with unsaved board state. The #240 design discussion identified KiCad's sibling lock as possible cold-start evidence but deferred it to a separate follow-up.

## Proposed conservative follow-up

Before any unreachable-IPC direct-file fallback, inspect the exact requested board's `~<filename>.lck` sibling:

- absent lock: retain the current closed-board fallback;
- present or unreadable lock: structured `stale_target` refusal and no file write;
- lock for another board: do not taint the requested board.

The revision-aware writer should check the board lock again immediately before its atomic rename, using the same typed lock-path derivation already used for schematics. This cannot make the external lock protocol atomic, but it narrows the last-check/rename interval.

## Trade-off requiring maintainer confirmation

KiCad 10 lock contents do not prove process ownership or freshness. A lock left by an unusual clean exit will therefore produce a conservative false-positive refusal. The proposal deliberately prefers an explicit recovery step over a potentially stale successful file write; it does not add a force flag or automatically remove locks.

## Acceptance

- Present, malformed/stale-looking, or unreadable exact-board locks refuse and leave bytes unchanged.
- An absent lock retains current offline behavior.
- A different board's lock has no effect.
- Both IPC fallback gates and the final revision-aware writer check the exact board.
- Structured errors identify the target and explain that the saved state may be stale.

A draft implementation and deterministic tests are ready, but this issue should confirm that the conservative false-positive policy is acceptable before merge.

neusse:
Maintainer decision: **accepted, with the fail-closed contract below.** This is enough direction to implement #385; the remaining work is engineering and evidence, not another policy decision.

A KiCad lock for the **exact requested board** is sufficient evidence to refuse an automatic direct-file mutation. That is the right tradeoff: a false-positive refusal is recoverable by the user, while a false-negative write can compete with or overwrite unsaved editor state.

Required behavior:

- Derive only the requested board's sibling lock (`~<board filename>.lck`) from the exact canonical board path. A lock for another board has no effect.
- If that exact lock is present, malformed, or cannot be inspected reliably, refuse the write without changing any board bytes.
[...]

Issue #383, opened 2026-08-30
Hierarchical placement records only the first reused child instance


dubesinhower:
## Problem

When a child schematic is reused in more than one hierarchical instance, component and power-symbol placement currently carries only the first saved instance path into the new symbol. The file can therefore contain a symbol whose project/path identity is incomplete even though the root hierarchy proves that the child has multiple instances.

That becomes observable later as missing or inconsistent references in hierarchy-aware reads, netlist identity, and multi-unit mutations. A successful placement response also cannot prove which hierarchy instances were actually committed.

## Proposed behavior

- Resolve every instance path for the exact child document before mutation.
- Write every affected project/path entry for single placement, batch placement, and power-symbol placement.
- Refuse missing, foreign, duplicate, or obsolete pre-existing instance state as structured `stale_target` rather than repairing or guessing.
- Reload the committed child and derive the response's document, UUID, reference, coordinates, rotation, library identity, and hierarchy paths from the observed symbol.
- Leave the wrong document byte-identical on any targeting failure.

## Compatibility and safety

Existing placement inputs and response fields remain. Successful responses gain observed hierarchy evidence; inconsistent hierarchy state that could previously be written through now refuses. The existing hierarchy-link operation remains the repair path because it updates the parent link and child paths together.

## Acceptance

- A uniquely nested child receives its exact instance path.
- A reused child receives every instance path, with no first-entry shortcut.
- Single, batch, and power-symbol placement share the same rule.
- Stale or ambiguous instance state produces a structured no-write refusal.
- Positive results are derived from a committed-file reload.
- Tests cover wrong-document isolation, stale state, ambiguity, and real multi-unit/hierarchy fixtures.

neusse:
Completed by PR #389 (merge commit `dbd2948890eb6781d1df3a8343b2a813b33bbb10`).

Acceptance evidence:
- Unique and reused child placement records the complete structurally observed hierarchy path set.
- Single-component, batch, and power-symbol placement use the same preflight and committed-file readback contract.
- Missing, foreign, duplicate, obsolete, or malformed project/path/reference/unit metadata returns structured `stale_target` before mutation.
- Wrong-document and ambiguous-ownership paths preserve the target and related root files byte-for-byte.
- Positive results are derived from the reloaded committed symbol.
- Tests use the KiCad-authored complex-hierarchy fixture, including reused sheets and multi-unit symbols.

[...]

PR #379, opened 2026-08-30
feat(review): add hierarchy scope to standalone audits

dubesinhower:
## Summary

Closes #378.

Add an explicit, additive `schematic_scope` contract to `audit_decoupling`, `audit_connections`, `audit_power_rails`, and `check_bom_health`.

- `file` remains the default and preserves existing one-file counts.
- `hierarchy` audits every reachable KiCad sheet instance.
- Both modes report scope, coverage, status, and diagnostics.
- Per-sheet hierarchy results and findings identify their source file and `sheet_instance_path`.
- Missing or cyclic child references produce `partial` or `failed`, never clean-looking success.

## Root cause and design

The standalone handlers read the supplied schematic directly. Hierarchy traversal existed only inside `run_design_review`, so standalone calls had no way to request hierarchy coverage or state that they had stopped at the root file.

This change refactors that traversal into one shared inspection path. Reused child files are audited once per sheet placement in hierarchy mode. `run_design_review` continues deduplicating file-level audit work, while its symbol and net coverage is now instance-aware.

## Compatibility and risk

The new input is optional and defaults to `file`; existing callers retain their original counts. Response changes are additive, and the existing decoupling `scope: "schematic_connectivity"` field is preserved. Because the public response shape grows, this belongs in minor-release planning.

Hierarchy mode can perform and return more work in proportion to the number of sheet instances. Findings from a reused child file intentionally appear once per placement and carry the instance path that distinguishes them. The change is read-only and does not modify KiCad files.

## Regression coverage

Focused tests cover:

- schema/default behavior for all four standalone audits;
- hierarchy coverage across a reused child file;
- one result per sheet instance with file and instance counts;
- unchanged one-file default behavior;
- structured rejection of invalid scope;
- incomplete status and diagnostics for missing and cyclic children;
- shared, instance-aware `run_design_review` coverage;
- shipped skill and tool-directory references.

The main hierarchy guard was neutralized by forcing a hierarchy request through file scope; `hierarchy_scope_counts_reused_child_instances_for_every_audit` failed as expected, then passed after restoration.

## Verification

Verified after contracting the PR to one feature commit directly on current `main`. The locally tested commit and published head `f9a8f3f068f3c57b888da5e1df55501765b89a82` have identical trees and stable patch IDs.

- `cargo fmt --all -- --check` — pass
- `cargo clippy --workspace --locked --all-targets -- -D warnings` — pass
- `cargo test --workspace --locked --lib --tests` — pass, including the full installed KiCad demo corpus
[...]

PR #388, opened 2026-08-30
fix(project): resolve schematic ownership structurally

### LINKED ISSUE #189: project_root_for's ancestor walk is still unbounded when the file's directory has no table and no project
Follow-up to the fix in `ec705c3` (from @Pytonballoon810's #188), which stops the search when the file's own directory carries a `sym-lib-table` or `fp-lib-table`. That covers the reproducible failure and the hermetic-fixture case. It does not bound the walk in general.

## Residual

`project_root_for` (`crates/konnect-core/src/tools/library.rs`) still climbs ancestors without limit when the schematic's directory carries **neither** a library table **nor** a `.kicad_pro`:

```
~/Documents/Other.kicad_pro      <- unrelated project
~/Documents/work/foo.kicad_sch   <- no table, no project beside it
```

The walk leaves `work/`, finds `Other.kicad_pro`, and every library lookup for `foo.kicad_sch` then resolves `${KIPRJMOD}` against `~/Documents`. Silently — the caller gets the wrong library, not an error.

Narrower than the original bug, since a real project normally has its `.kicad_pro` beside or above its schematics, but the failure mode is the same and it is invisible when it happens.

## Why it wasn't fixed in the hotfix

Bounding the walk is a design decision rather than a one-liner, and it landed as a patch on a release cut hours earlier. Options, none obviously right:

- Stop at a VCS root (`.git`) — pragmatic, but couples library resolution to source control.
- Depth limit — arbitrary, and wrong for a deeply nested sheet hierarchy.
- Never cross above the user's home directory — helps the common case, does nothing for a shared parent inside it.
- Require an explicit `project` argument when resolution is ambiguous, and report the ambiguity — the most honest, the most disruptive.

## Acceptance

- [ ] A schematic with no table and no project beside it, nested under an unrelated `.kicad_pro`, does not silently resolve against that project.
- [ ] Whatever bound is chosen is stated in the doc comment, with the reasoning.
- [ ] A deep sheet hierarchy (`<proj>/sheets/sub/x.kicad_sch`) still resolves at the project root — that is `a_sheet_in_a_subdirectory_resolves_against_the_project_table`.
- [ ] If ambiguity is reported rather than guessed, it is a structured error naming both candidate roots.

dubesinhower:
## Summary

Resolve saved schematic ownership from project sheet-tree membership before
symbol loading or ERC root detection. A schematic beneath an unrelated project,
an unreadable candidate hierarchy, or multiple proven owners now returns the
existing structured `conflict` instead of guessing or silently becoming
projectless. The error identifies the schematic directory and all candidate
root schematics.

Closes #189.

## Behavior and scope

- Preserve library-table authority beside the schematic and allow genuinely
  projectless files when no ancestor candidate exists.
- Resolve relative paths against the working directory before searching
  ancestors; a sibling `.kicad_pro` does not bypass membership verification.
- Retain candidate roots independently of successful reads. Incomplete sheet
  traversal cannot establish uniqueness; an unreadable discovery directory
  returns an explicit `file_not_found` refusal.
- Share structural traversal between placement, library lookup, and ERC.
  Resolve the batch library context once before processing any components.
- Rebuild only the ownership change on current `main` after #381, preserving
  the accepted #372/#374 prerequisite implementations and fixtures.

No tool, argument, new public error kind, or later placement-instance feature is
introduced. The unused `ambiguous_target` addition is excluded. The behavior
change and recovery guidance are documented for the next minor release in
`docs/PROJECT_OWNERSHIP.md`, `docs/API_MIGRATIONS.md`, and `DEV.md`.

## Validation

The maintainer-reported defects were reproduced by three failing regressions
before correction. The relative-path defect was also reproduced in an isolated
subprocess. The ownership suite covers native repeated/deep hierarchy paths,
all candidate diagnostics, local table authority, missing and unreadable roots,
partial traversal, cycles, depth limits, and relative paths.

The six-handler preflight regression exercises placement, batch placement,
symbol refresh, replacement, power-symbol insertion, and ERC with unchanged
target bytes. Five deliberate guard regressions were detected, including a
batch-preflight bypass that otherwise writes the file. Every temporary mutation
was restored before the final checks.

The complete contributor gate passed on commit `7e2aa54`:

- `cargo fmt --all -- --check` — exit 0.
- `cargo clippy --workspace --locked --all-targets -- -D warnings` — exit 0.
- `cargo test --workspace --locked --lib --tests` — exit 0; 1,412 passed,
  19 intentionally ignored by the standard gate.
- `cargo test --workspace --locked --doc` — exit 0; 8 passed, 2 ignored.

The focused ownership suite passed all 17 tests. Ignored checks require live
KiCad/IPC, Java, deliberate long-timeout execution, or are documentation examples.

The relative-path subprocess regression compares canonical file and directory
paths, so macOS's `/var` and `/private/var` aliases do not cause a false failure.
The initial hosted run exposed this assertion issue; the final commit includes
the portable comparison without changing ownership behavior.

[Hosted CI](https://github.com/mixelpixx/Konnect/actions/runs/33694745973) passed
all 10 checks on `7e2aa54`, including Linux, macOS, and Windows check/test/doc
jobs, Nix packaging, the schematic viewer, Python plugin, Clippy, formatting,
dependency licenses, and PCM packaging validation.

## Limits and risk

The fixture files are byte-preserved copies of KiCad's complex-hierarchy demo;
tests alter only temporary copies. Ownership is based on saved files read during
preflight, not live editor state or an atomic snapshot of the whole project.
Existing target-file revision checks remain in force. The local Windows gate
and hosted Linux, macOS, and Windows gates passed. Live KiCad IPC and release
benchmarks were not exercised; no editor session is required by this
closed-file change.

The intentional compatibility risk is that previously silent, incorrect
resolution becomes a refusal. Recovery is to restore the saved hierarchy or
separate an independent document from the unrelated project, rather than retry
against an arbitrarily selected candidate.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -258,8 +258,8 @@ Tool-call failures are typed via the `ToolErrorKind` enum in `crates/konnect-cor
 | `toolset_not_loaded` | Tool exists but its toolset isn't loaded yet |
 | `unknown_tool` | Tool name doesn't exist in any toolset |
 | `invalid_argument` | Required argument missing/malformed |
-| `file_not_found` | Referenced file doesn't exist |
-| `conflict` | The file changed after it was read, or a write would replace paths that already exist — carries them |
+| `file_not_found` | Referenced file or project-discovery directory does not exist or cannot be read |
+| `conflict` | The file changed, a write would replace existing paths, or schematic project ownership cannot be proven uniquely — carries the affected paths; ownership conflicts include the schematic directory and all candidate roots |
 | `handler_error` | Catch-all for unmigrated `anyhow::Error` returns |
 
 ### Producing structured errors in a handler
--- a/crates/konnect-core/src/mcp/error.rs
+++ b/crates/konnect-core/src/mcp/error.rs
@@ -47,9 +47,10 @@ pub enum ToolErrorKind {
     UnknownTool { tool: String },
     /// A required argument is missing or malformed.
     InvalidArgument { field: String, reason: String },
-    /// A referenced file doesn't exist or can't be read.
+    /// A referenced file or discovery directory doesn't exist or can't be read.
     FileNotFound { path: String },
-    /// A mutation would replace one or more existing filesystem targets.
+    /// A mutation conflicts with filesystem state, or schematic ownership
+    /// cannot be proven uniquely. Paths identify the conflicting evidence.
     Conflict { paths: Vec<String> },
     /// A board was live earlier in this server process, but IPC is now gone;
     /// its saved file may be stale relative to lost editor state.
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -1157,46 +1157,34 @@ fn global_sym_lib_table() -> PathBuf {
     super::kicad_config_dir().join("sym-lib-table")
 }
 
-/// Directory of the nearest ancestor of `file` that holds a `.kicad_pro`,
-/// falling back to the file's own directory when it belongs to no project — a
-/// loose schematic keeps resolving against the tables beside it.
+/// Directory of the structurally proven project that owns `file`, falling back
+/// to the file's own directory when it belongs to no project — a loose
+/// schematic keeps resolving against the tables beside it.
 ///
 /// The project file is found by scanning for the extension rather than by name:
 /// a sheet's filename says nothing about what the project is called.
 ///
-/// A library table sitting beside `file` ends the search before it starts. The
-/// ancestor walk is unbounded, so without that it can leave the file's own
-/// directory and latch onto an unrelated `.kicad_pro` further up — a project
-/// nested inside another project's folder, or a stray file in a shared parent —
-/// and then resolve every library against the wrong `KIPRJMOD`. A directory
-/// carrying its own `sym-lib-table` or `fp-lib-table` is stating where its
-/// libraries come from, and that is the more specific answer.
-pub(crate) fn project_root_for(file: &Path) -> Option<PathBuf> {
-    let start = file.parent()?;
+/// A library table sitting beside `file` is the most specific answer. An
+/// ancestor project is accepted only when its parsed root schematic reaches
+/// this file. Unproven or multiple owners produce the existing structured
+/// conflict, naming the schematic directory and all candidate roots (#189).
+pub(crate) fn project_root_for(
+    file: &Path,
+) -> Result<Option<PathBuf>, crate::tools::SchematicTargetError> {
+    let Some(start) = file.parent() else {
+        return Ok(None);
+    };
     if holds_lib_table(start) {
-        return Some(start.to_path_buf());
+        return Ok(Some(start.to_path_buf()));
     }
-    start
-        .ancestors()
-        .find(|dir| holds_kicad_pro(dir))
-        .map(Path::to_path_buf)
-        .or_else(|| Some(start.to_path_buf()))
+    Ok(crate::tools::resolve_schematic_ownership(file)?
+        .and_then(|ownership| ownership.project_file.parent().map(Path::to_path_buf))
+        .or_else(|| Some(start.to_path_buf())))
 }
 
 /// Whether `dir` carries a library table of its own.
 fn holds_lib_table(dir: &Path) -> bool {
     dir.join("sym-lib-table").is_file() || dir.join("fp-lib-table").is_file()
-}
-
-/// Whether `dir` contains a `.kicad_pro`. An unreadable directory holds none.
-fn holds_kicad_pro(dir: &Path) -> bool {
-    std::fs::read_dir(dir).is_ok_and(|entries| {
-        entries.flatten().any(|e| {
-            e.path()
-                .extension()
-                .is_some_and(|ext| ext.eq_ignore_ascii_case("kicad_pro"))
-        })
-    })
 }
 
 /// Symbol libraries resolved as KiCad does: project `sym-lib-table` (shadowing
@@ -1228,8 +1216,8 @@ impl KiCadSymbolSource {
     /// hierarchical sheet under `<proj>/sheets/` therefore still resolves
     /// against `<proj>/sym-lib-table`: KiCad anchors `KIPRJMOD` at the project,
     /// not at the sheet.
-    pub(crate) fn for_file(file: &Path) -> Self {
-        Self::new(project_root_for(file))
+    pub(crate) fn for_file(file: &Path) -> Result<Self, crate::tools::SchematicTargetError> {
+        Ok(Self::new(project_root_for(file)?))
     }
 
     /// Project entries first so they shadow same-nickname global ones.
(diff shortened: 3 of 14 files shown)


neusse (changes requested):
Reviewed current head `09d38d9f3f24a7facffbbe7434d59f87170f8119` against the final maintainer policy in #189 and current `main` after #372. Changes are required for both behavior and branch ancestry.

## 1. The confirmed #189 policy is not implemented yet

The final issue decision says that when candidate ancestor projects exist but ownership cannot be proven, Konnect must return the existing structured `conflict`, naming the schematic directory and every candidate project root. That includes:

- a candidate `.kicad_pro` with no matching root schematic;
- an unreadable root schematic;
- a parsed root that does not reach the target;
- more than one candidate/proven owner.

A genuinely loose schematic with **no ancestor project candidate at all** remains allowed.

The current resolver instead skips missing/unreadable roots, treats an unreachable target as projectless, and maps multiple proven owners to `ambiguous_target`. The tests `loose_sheet_under_an_unrelated_project_stays_loose` and `missing_or_unreadable_candidate_root_is_not_ownership_evidence` preserve the opposite behavior. Please retain candidate roots separately from proven owners, return `conflict` whenever candidates exist but unique membership cannot be established, include the target directory and all candidate roots in the error, and rewrite those tests. If `AmbiguousTarget` is no longer used by this PR after that correction, defer it to the later PR that actually needs it rather than adding an unrelated public error kind here.

The shared structural hierarchy traversal, deep-sheet fixture correction, centralized resolver direction, and batch preflight are otherwise the right design.

## 2. Rebuild this stack only after the foundation queue lands

Every #388-#394 branch still contains copied prerequisite commits, including the superseded pre-review #372 implementation (`1f4de18`/`a4a9b0b`). Since corrected #372 is now on `main`, merge simulation produces real conflicts in `crates/konnect-sexp/src/board.rs`. Resolving by choosing this branch's file would risk restoring the rejected source-breaking `ZoneOutline` representation and dropping the KiCad-authored evidence fixture.

Please do **not** rebase #388 repeatedly while the foundation PRs are moving. We will merge:

`#369 -> #374 -> #376 -> #379 -> #381`

After #381 lands, rebuild #388 from that `main`, retaining and correcting only the unique ownership commit (`09d38d9f` today). An `--onto` rebase or a fresh branch plus cherry-pick and `--force-with-lease` are both reasonable; the important result is that already-merged prerequisite commits and the obsolete #372 copy disappear from this PR. Then rerun all checks on the contracted diff.

The #189 discussion also contains an unresolved contributor-coordination direction involving @anyn99 and @neusse. We will settle that separately; please avoid spending time on the restack until the foundation queue and that coordination are resolved.

Once corrected and restacked, #388 becomes the base for #389 and the two later lanes. Its changed refusal behavior must be called out in release notes.

neusse:
## Stack coordination

This PR is now the key that unlocks the remaining schematic work. The prerequisite foundation through #381 has merged, but this branch still carries eight prerequisite commits that are already represented on current `main`, followed by the unique ownership-resolution commit.

Before rewriting the branch, ownership of #189 still needs to be settled as requested by Chris: @anyn99 volunteered first, while @dubesinhower opened this implementation. Please resolve that coordination on #189 and assign the issue to the agreed implementer.

**The most efficient approach is:**

1. Rebuild #388 from current `main` instead of conventionally rebasing all nine commits.
2. Retain only the unique ownership-resolution work and correct it to match the confirmed #189 policy:
   - a table beside the schematic remains authoritative;
   - accept an ancestor project only when its root schematic proves sheet-tree membership;
   - return a structured `conflict` naming the schematic directory and every candidate root when an ancestor exists but ownership cannot be proven;
   - continue allowing a loose schematic when no candidate ancestor exists;
   - preflight batch ownership once before mutation;
   - correct the deep-sheet fixture to model a real root sheet referencing the child.
3. Run the focused ownership, hierarchy, library-resolution, and batch atomicity tests, then the complete required gate.
4. Merge #388 before reconstructing the dependent file-only chain: #389 → #392 → #393 → #394.
5. Rebuild #390 separately after that chain because its requested-board IPC behavior is an independent lane.
6. Keep draft #391 on hold until #385 receives the maintainer policy decision and supporting live evidence.

This avoids preserving obsolete prerequisite copies, makes this PR reviewable as one focused change, and prevents every later PR from repeating the same conflict work.

dubesinhower:
The review findings still apply to head `09d38d9f`. I confirmed that #381 is merged and current `main` is `5829af60`; this PR can be reconstructed from that foundation once the #189 contributor handoff is agreed.

I responded to the ownership-coordination request on #189: https://github.com/mixelpixx/Konnect/issues/189#issuecomment-5513312829. I am ready to finish the correction if @anyn99 agrees, and will preserve the existing branch until that decision and assignment are recorded.

The correction will retain candidate roots independently of proven owners, use the existing structured `conflict` for unproven/ambiguous ownership with all candidate roots and the schematic directory, preserve adjacent library-table authority and genuinely projectless files, and keep ownership preflight before batch mutation. The contradictory projectless/ambiguity tests need rewriting. The deep-sheet library test already contains a root-to-child Sheetfile link; that coverage needs retaining and checking against native hierarchy fixtures.

I also compared the local descendants with current `main`: they still carry the superseded #372 zone representation and lack the accepted #374 mandatory-property validation and native fixture. Rebuilding only the unique ownership work avoids restoring those obsolete prerequisite copies. The unrelated `AmbiguousTarget` public kind will be deferred to the later change that needs it.

The sequence remains #388 → #389 → #392 → #393 → #394, then #390. No code correction, branch rewrite, or fresh merge-ready gate is claimed yet; the coordination prerequisite is still open.

neusse:
here is @anyn99's respose.  looks like we can continue.

anyn99
 left a comment 
[(mixelpixx/Konnect#189)](https://github.com/mixelpixx/Konnect/issues/189#issuecomment-5515082274)
No objections here, please go ahead [@dubesinhower](https://github.com/dubesinhower) !
For the future: I am not here for the creds, just for making this software better! I will always find new ways to improve the code, so no need to wait out of courtesy 😉

neusse (approved):
Reviewed current head `7e2aa5447d51a97cd73c3fa0262e40c574f53a7d` against the confirmed #189 policy and current `main`.

The prior requested changes are resolved: the branch is contracted to one unique commit, preserves adjacent library-table authority, accepts only structurally proven ownership, reports the existing structured `conflict` with the schematic directory and every candidate root when ownership cannot be established uniquely, allows a genuinely loose schematic with no ancestor candidate, uses shared cycle/depth-bounded traversal, and preflights symbol-loading/placement paths before writes. The corrected fixtures are captured from KiCad's complex-hierarchy demo, and the compatibility/refusal behavior is documented for the next minor release.

Focused local verification passed: all 17 ownership tests, the ambiguous-placement byte-preservation regression, and the six-handler symbol-loading/ERC no-write preflight regression. All ten hosted checks are green.

The existing incomplete batch/power-symbol instance metadata is intentionally tracked by #383 and its dependent PR #389; it is not a regression or omitted acceptance item in #189.

PR #390, opened 2026-08-30
fix(ipc): bind operations to the requested board

dubesinhower:
## Summary

Bind generic KiCad IPC reads and writes to the exact board document selected for the path-bearing MCP request. With multiple boards open, a lower helper can no longer rediscover the first board and substitute it for the validated target.

Closes #384. This is complementary to the test-only direct-caller coverage in #241 and does not claim or replace that assigned work.

Stacked on #389. The only new commit after that PR is `fix(ipc): bind operations to the requested board`.

## Approach

- Carry the exact typed `DocumentSpecifier` returned by `find_open_board` with the requested path.
- Re-observe that binding before generic read/write commands.
- Distinguish no document, wrong document, duplicate matches, and a disappeared bound document in `BoardTargetError`.
- Preserve those classifications through `IpcFailure` and map them to structured `wrong_document`, `ambiguous_target`, or `stale_target` MCP errors.
- Exercise typed request headers through the existing document-answering mock.

## Compatibility and safety

The normal single-open-document behavior and explicit low-level `*_in(DocumentSpecifier, ...)` APIs remain. Wrong, duplicate, or disappeared document identities now refuse instead of reaching a generic command or a different board. No input schema or existing response field changes.

## Validation

- [x] `cargo fmt --all -- --check`
- [x] `cargo test --workspace --locked --lib --tests` — including 927 passed / 9 ignored in `konnect-core`
- [x] `cargo test --workspace --locked --doc` — 8 passed / 2 ignored
- [x] `cargo clippy --workspace --locked --all-targets -- -D warnings`
- [x] Requested-board-listed-second read/write headers, zero/wrong/duplicate tables, disappeared binding, structured MCP mapping, wrong-document no-write gate, IPC, conformance, and integration tests

The remaining close-after-final-recheck race fails closed because the typed command still carries the previously bound document. Ignored checks require a live KiCad/Java or another external runtime; the document protocol behavior is covered by deterministic mock IPC.

PR #389, opened 2026-08-30
fix(schematic): preserve all hierarchical placement instances

dubesinhower:
## Summary

Preserve every hierarchical instance path when placing a component or power
symbol into a reused child schematic, and derive placement results from the
committed child document.

Closes #383.

This is the next focused file-only PR after merged #388. The branch is rebuilt
from current `upstream/main` (`806277e`) and contains one unique commit,
`fix(schematic): bind placement to all sheet instances` (`7cb00de`). The agreed
sequence after this PR is #392 -> #393 -> #394; #390 remains a separate IPC lane.

## Behavior

- Resolve every hierarchy instance for the exact target before mutation.
- Use the same hierarchy identity for single component, batch component, and
  power-symbol placement.
- Refuse missing, foreign, duplicate, malformed, or obsolete existing instance
  metadata as structured `stale_target` before writing.
- Write every observed project/path entry for a unique or reused child.
- Reload the committed child and derive document, UUID, library identity,
  reference, value, coordinates, rotation, unit, project, and instance paths
  from the observed symbol.
- Return `stale_target` when post-write readback cannot prove the intended
  document or symbol. This may follow a committed write, so the error and
  migration notes tell callers to inspect/reload before retrying.

## Compatibility and release impact

Existing tools, inputs, and response fields remain. Successful placement
responses gain saved hierarchy and readback evidence. Inconsistent saved state
that could previously be written through now refuses. `DEV.md` documents the
new `stale_target` kind, and the tool directory, ownership guide, and API
migration notes record the additive response shape, repair path, limitations,
and next-minor-release obligation.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --locked --all-targets -- -D warnings`
- `cargo test --workspace --locked --lib --tests`
- `cargo test --workspace --locked --doc`
- KiCad-authored complex-hierarchy fixture coverage for unique and reused
  children, a real multi-unit symbol, all three placement paths, stale and
  ambiguous no-write behavior, wrong-document isolation, and committed-file
  readback
- Every new preflight, all-instance write, and readback guard was individually
  disabled and its regression test failed; sources were restored afterward

The complete gate passed on `7cb00de`, based on exact current main `806277e`.
Ignored tests require live KiCad, Java, another operating system, or another
external runtime; this file-only behavior does not depend on unsaved editor
state. Hosted checks must pass on this rewritten head before merge.

## Risk and rollback

The change affects saved schematic instance metadata and placement response
[...]

PR #391, opened 2026-08-30
fix(board): refuse stale locked file fallbacks

dubesinhower:
## Summary

Use the exact requested board's KiCad sibling lock as conservative cold-start evidence before an unreachable-IPC direct-file fallback. A present or unreadable lock now produces a structured no-write refusal; a different board's lock does not taint the request.

Draft for the policy decision in #385. This is the separately deferred cold-start follow-up to #361/#240, not a correction to their process-lifetime memory.

Stacked on #390. The only new commit after that PR is `fix(board): refuse stale locked file fallbacks`.

## Approach

- Derive exact schematic and board sibling lock paths in `konnect-sexp`.
- Check the requested board lock in both unreachable-IPC fallback gates.
- Treat presence or inspection failure as `stale_target`; retain offline file mode when absent.
- Recheck both design-document lock types in the revision-aware writer immediately before atomic rename.
- Keep exact-board isolation and byte-identical refusal tests at the tool boundary.

## Compatibility and safety

This intentionally narrows behavior: a cold-start board carrying a sibling lock can no longer enter file mode automatically. No force flag is added. A stale lock left after an unusual clean exit can therefore cause a conservative false positive; the issue asks maintainers to confirm that policy before this draft becomes ready.

KiCad's external lock protocol cannot make the last metadata-check/rename interval atomic. The double check narrows it to the smallest available window but does not claim to eliminate it.

## Validation

- [x] `cargo test --workspace --locked --lib --tests` — including 930 passed / 9 ignored in `konnect-core`
- [x] `cargo test --workspace --locked --doc` — 8 passed / 2 ignored
- [x] Exact lock derivation, present/absent/malformed/unreadable/wrong-board lock, conditional writer, unchanged-byte refusal, board-session, conformance, and integration tests

Not run: cross-platform live KiCad lock-lifecycle measurement. That evidence is the explicit design-review boundary for this draft; deterministic tests cover Konnect's behavior once the filesystem evidence exists.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,9 @@ permissions: {}
 env:
   CARGO_TERM_COLOR: always
   BINARY_NAME: konnect
+  # Embedded into get_installation_info so release artifacts identify the
+  # exact source commit even when built outside a Git working tree.
+  KONNECT_BUILD_COMMIT: ${{ github.sha }}
 
 jobs:
   build:
--- a/DEV.md
+++ b/DEV.md
@@ -192,6 +192,11 @@ Konnect/
 - Cooperative lock files live under `KONNECT_STATE_DIR/locks` when that
   absolute override is set, otherwise under the platform local-data directory
   (`konnect/locks`). Reads never create files in the KiCad project.
+- Schematic writes also refuse while KiCad's sibling `~<name>.kicad_sch.lck`
+  exists. KiCad records only a username and hostname, so Konnect cannot prove
+  that a same-host or remote lock is stale; valid, foreign, empty, and malformed
+  locks all fail closed. The check runs before a transaction journal is created
+  and again at the final target-write boundary.
 - Multi-file schematic changes use project-local
   `.konnect-transaction-*.json` write-ahead journals. These journals contain
   complete before/after images and must be treated as sensitive project data.
--- a/DEV.md
+++ b/DEV.md
@@ -77,7 +77,7 @@ Konnect/
 │   │       ├── router/
 │   │       │   ├── mod.rs           # ToolRouter: load/unload toolsets
 │   │       │   ├── registry.rs      # Static toolset metadata + tools_for() dispatcher
-│   │       │   └── meta_tools.rs    # 6 always-visible meta-tools
+│   │       │   └── meta_tools.rs    # 7 always-visible meta-tools
 │   │       └── tools/
 │   │           ├── mod.rs            # ToolDef, ToolContext, tool! macro, helpers, kicad_config_dir()
 │   │           ├── cli.rs            # kicad-cli v10 subprocess wrapper (verified against actual binary)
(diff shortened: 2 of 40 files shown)


neusse (commented):
Reviewed current draft head `1474147d3d201d0c4c4f0cc185904ae34ab303ab` against #385.

Keeping this as a draft is correct. The deterministic implementation appears internally consistent—exact board lock derivation, present/unreadable lock refusal, wrong-board isolation, fallback-gate checks, final writer recheck, and byte-identical no-write evidence—but the issue still requires a maintainer policy decision accepting conservative false positives from stale KiCad locks. The PR also correctly acknowledges that cross-platform live KiCad lock-lifecycle behavior has not been measured and that the external lock protocol cannot make the final metadata-check/rename interval atomic.

Please do not rebase this draft while the queue is moving. It comes only after #390. Once #390 is merged and the #385 policy is explicitly decided, rebuild #391 on current `main` with only `1474147d` (or its rebased equivalent), add the agreed recovery guidance, collect the live cross-platform lifecycle evidence, and then decide whether it should become ready for review.

neusse:
## Stack coordination

This draft currently inherits the entire old #388 → #389 → #390 history and adds one unique stale-locked-file fallback commit. It is also still policy-blocked: #385 has the `needs:maintainer-decision` label and no maintainer decision has been recorded. Rebasing now would spend effort on behavior that may still change.

**The most efficient approach is:**

1. Leave this PR in draft and do not rebase it yet.
2. Land the clean schematic sequence first: #388 → #389 → #392 → #393 → #394.
3. Rebuild and settle the independent #390 IPC change, including its overlap with #382.
4. Obtain the explicit #385 policy decision defining when a KiCad lock is trustworthy stale-file evidence, plus the required live cross-platform evidence.
5. Rebuild #391 once from then-current `main`, carrying only its unique stale-lock behavior (`1474147d`) and adapting its tests to the approved policy.
6. Add the correct issue-accounting relationship: this PR currently has no closing issue reference. Use `Closes #385` only if the final focused PR satisfies every current acceptance criterion; otherwise use `Part of #385` and document the remaining work.
7. Run the focused lock/fallback tests and the complete required gate before requesting review.

This avoids multiple speculative rebases and ensures the final review is about one agreed behavior rather than twelve inherited commits plus an unresolved policy choice.

neusse:
The policy decision in #385 is now settled and needs:maintainer-decision has been removed. This PR still cannot enter the merge queue in its current cumulative/conflicting form.

**Required path into the merge queue:**

1. Wait for the corrected, focused #390 to merge.
2. Reconstruct #391 once from then-current upstream/main, carrying only the unique exact-board lock behavior represented by 1474147d; do not replay the inherited stack.
3. Implement the accepted #385 contract:
   - only the exact requested board's canonical sibling ~<board filename>.lck may veto that board's fallback;
   - present, malformed, or unreadable exact lock refuses with byte-identical output;
   - another board's lock has no effect;
   - absence of a lock is not proof of safety and cannot bypass IPC, session-history, or document-identity gates;
   - ambiguous/unresolved open-document identities fail closed;
   - recheck immediately before atomic replacement;
   - no PID/age inference, automatic removal, retry loop, or force/ignore option.
4. Reuse structured unsafe_file_fallback, with result-derived evidence/message distinguishing lock present from lock unreadable. Do not use stale_target for this condition.
5. Keep board fallback scoped to the board lock. A schematic lock must not independently veto an unrelated board write merely because the shared writer can edit several document types.
6. Add Closes #385 once the focused PR owns the complete acceptance checklist.
7. Provide deterministic present/absent/unreadable/malformed/other-board and late-appearing-lock tests, byte-identical refusal assertions, and real-KiCad lock lifecycle observations on supported operating systems. Record unavailable evidence as BLOCKED.
8. Run the complete required gate and request review only after GitHub reports the exact reconstructed head mergeable with all required checks green.

At that point #391 can follow #390 in the merge queue. Its old August 30 green checks are historical only.

dubesinhower:
Preparatory Windows evidence for the accepted #385 contract, collected while validating #407. No branch changes or restacking were performed; #391 remains held until #390 lands.

With a disposable saved board open in KiCad PCB Editor 10.0.5 on Windows x86_64:

- The exact sibling `~PR407 Evidence.kicad_pcb.lck` was present (43 bytes).
- After a normal editor close, that exact lock was absent. Window enumeration and Konnect's bounded UI/IPC check confirmed the PCB Editor had exited.
- The board's SHA-256 remained `EF1AE5C9BFF611056CA81ABD25C1D9773F866266659D7F5AC41D518814FC1FAD` throughout the read-only test and clean close.
- The board filename and parent path contain spaces. No lock contents were interpreted and no lock was manually removed.

This is evidence of the Windows open/clean-close lifecycle only. Crash/forced-exit behavior, Linux/macOS lifecycle behavior, and the final reconstructed #391 implementation remain unverified here. Those items must not be marked passing from this observation.

neusse:
Reconstructed on current `main` as #486 after #480 landed. The replacement contains only the unique #385 work, preserves @dubesinhower's commit authorship, and adjusts the implementation to the accepted maintainer decision: `unsafe_file_fallback` with distinct lock evidence rather than the old draft's `stale_target` category. Closing this cumulative draft so review and CI continue on the focused replacement.

Issue #395, opened 2026-08-30
Design Priority 1 semantic editor navigation MVP


dubesinhower:
## Problem

Konnect can inspect and mutate KiCad design data, but a client cannot reliably
establish and change the live editor context it is discussing with a user. It
cannot enumerate addressable schematic/PCB documents, observe the exact
selection, activate a hierarchical sheet instance, select and reveal an exact
KIID, or cross-probe a symbol to its footprint with result-derived proof.

The unsafe alternatives are guessing from open-document order, inferring live
state from saved files, or treating mouse/keyboard/window automation as the
primary interface.

This issue proposes one focused Priority 1 semantic Navigation MVP. It does not
cover Priority 2 batch inspection, arbitrary viewport control, notifications,
or structural design edits.

## Foundations and related work

- #388 and #389 establish structural project and hierarchical-instance
  identity.
- #390 binds generic IPC operations to the exact requested document.
- Draft #391 is an independent offline board-lock policy decision and is not a
  prerequisite for read-only navigation.
- #256 covers launching a specific PCB editor and selecting the resulting IPC
  endpoint. This design may consume that capability later, but does not
  duplicate its process-launch/session-selection work.

## Proposed semantic surface

Names here are provisional and require maintainer approval. The important
contract is the semantic split and evidence model.

1. `get_editor_state`
   - report running/addressable schematic and PCB frames;
   - report open typed documents;
   - report the active editor, document, and schematic sheet instance only
     when KiCad exposes them;
   - report the running KiCad version and a capability matrix for observation,
     selection mutation, activation, reveal/fit/center, and cross-probe.
2. `get_editor_selection`
   - query one explicit project/document/editor/sheet context;
   - return stable object KIIDs/UUIDs and object types with live-IPC evidence.
3. `resolve_navigation_target`
   - resolve one explicit project/document/editor/sheet plus KIID/UUID;
   - optionally accept a human reference such as `C10` only when saved-design
     structure produces exactly one live-addressable target;
   - return structured candidates on ambiguity rather than choosing one.
4. `set_editor_selection`
   - semantic `clear`, `add`, and `remove` modes;
   - exact document/sheet/object targeting;
   - read selection back and derive success from the observed set.
5. `activate_editor_target`
   - activate an exact already-addressable schematic, PCB, or hierarchical
     sheet instance when the running KiCad API supports it;
   - select and reveal the exact object when supported;
   - report activation/reveal/center/fit independently so unavailable behavior
[...]

PR #396, opened 2026-08-30
feat(navigation): observe editor capabilities

dubesinhower:
## Summary

- add a typed KiCad IPC observation model for the running version, addressable schematic/PCB editors, open-document identities, and navigation capability availability
- add provisional `get_editor_state` MCP observation behind the `editor_navigation` toolset
- preserve the distinction between live IPC evidence and structural file evidence; never infer active context from open-document order
- return structured `editor_unavailable`, `unsupported_capability`, and `stale_target` refusals

## Navigation MVP rationale

This is N1, the observation foundation for Priority 1 navigation. Later selection, target resolution, mutation, activation, and cross-probe slices need a typed capability/document snapshot so they can fail closed instead of silently retargeting.

Public naming remains provisional under #395, so this PR is intentionally a draft.

## Dependency

- depends on #390 (and therefore #389 → #388 → #381)
- does not depend on draft #391; read-only navigation does not require its stale cold-start board-lock policy

## Intentionally unsupported

- enumerating every independent KiCad endpoint/frame from one configured endpoint
- foreground editor, active document, and active hierarchical-sheet observation where KiCad 10 has no stable typed query
- activation, reveal/center/fit, selection mutation, and cross-probe (later slices)

Unsupported capabilities are reported explicitly. Active fields remain `null`; open-document order is not presented as active-state evidence.

## Verification

- `cargo test -p konnect-ipc --test mock_server_test` — 38 passed, 1 ignored
- `cargo test -p konnect-core editor_navigation --lib` — 3 passed
- `cargo test --workspace --all-targets` — pass (live/installed-tool tests remain ignored by their existing contracts)
- `cargo clippy --workspace --all-targets -- -D warnings` — pass
- `xtask fix-doc-counts --check` — 21 toolsets, 222 registered tools, 229 total; current
- `git diff --check` — pass
- `cargo test -p konnect-ipc --test live_kicad_test editor_state_observation_reports_real_version_and_honest_capabilities -- --ignored --exact --nocapture --test-threads=1` — 1 passed against the live KiCad 10.0.5 Windows named pipe while a disposable KiCad-authored project was loaded in the project manager; this proves version/capability observation only, not live-document selection or activation

## Risks

- KiCad 10 schematic `DocumentSpecifier` exposes a hierarchical sheet path rather than a schematic filename; this PR keeps `document_path` null rather than deriving it from disk
- capability availability is version- and runtime-probe-derived; later slices must still verify each operation by result readback
- public MCP names and request compatibility are awaiting maintainer direction in #395

## Checklist

- [x] New tests added for affected behavior
- [x] Documentation/tool counts updated
- [x] Public behavior outside the new toolset preserved

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,9 @@ permissions: {}
 env:
   CARGO_TERM_COLOR: always
   BINARY_NAME: konnect
+  # Embedded into get_installation_info so release artifacts identify the
+  # exact source commit even when built outside a Git working tree.
+  KONNECT_BUILD_COMMIT: ${{ github.sha }}
 
 jobs:
   build:
--- a/DEV.md
+++ b/DEV.md
@@ -192,6 +192,11 @@ Konnect/
 - Cooperative lock files live under `KONNECT_STATE_DIR/locks` when that
   absolute override is set, otherwise under the platform local-data directory
   (`konnect/locks`). Reads never create files in the KiCad project.
+- Schematic writes also refuse while KiCad's sibling `~<name>.kicad_sch.lck`
+  exists. KiCad records only a username and hostname, so Konnect cannot prove
+  that a same-host or remote lock is stale; valid, foreign, empty, and malformed
+  locks all fail closed. The check runs before a transaction journal is created
+  and again at the final target-write boundary.
 - Multi-file schematic changes use project-local
   `.konnect-transaction-*.json` write-ahead journals. These journals contain
   complete before/after images and must be treated as sensitive project data.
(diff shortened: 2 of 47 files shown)


neusse:
Thanks for keeping this as a draft and documenting the dependency chain. Please pause additional navigation slices for now and keep the immediate focus on #369. It is now green and mergeable, so we can review its current head and begin landing this stack from the bottom. After each prerequisite lands, we can ask for only its immediate successor to be restacked once. Rebasing #396 now, or adding N3-N6 above it, would multiply the conflict-resolution work because this branch still carries the older stack through #390. Please preserve the unique #396 work so it can be rebuilt on the final IPC base after #390 and after the public API decisions in #395 are settled.

neusse:
Reconstructed as focused PR #488 directly on current main. The replacement preserves @dubesinhower authorship on the unique N1 commit, drops the cumulative prerequisite history, reconciles the current board-safety contracts, and passes the complete local gate. Closing this superseded draft so the merge queue has one authoritative N1 PR.

PR #397, opened 2026-08-30
feat(navigation): observe exact editor selection

dubesinhower:
## Summary

- add exact project/document/sheet-bound `GetSelection` observation with a before/after open-document freshness check
- decode selected board objects and the stable schematic line/label protobufs into KIID/UUID results using exact full type URLs
- add provisional `get_editor_selection` to the `editor_navigation` toolset
- add structured wrong-project, wrong-document, wrong-sheet-instance, ambiguity, stale-state, unsupported-type, and malformed-item refusals

## Navigation MVP rationale

This is N2, the readback foundation for verified navigation. Selection mutation and cross-probe cannot claim success unless the client can observe the exact selected KIIIDs in the intended project, document, editor, and hierarchy instance.

Public naming and compatibility remain provisional under #395, so this PR is intentionally a draft.

## Dependency

- depends on draft #396 (N1), which depends on #390
- does not depend on draft #391

## Behavior and refusals

The request must echo one exact identity returned by `get_editor_state`. `GetOpenDocuments` is matched before `GetSelection` and read again afterward because KiCad's `SelectionResponse` carries no response header. No first-document fallback exists.

The full observation fails closed for:

- wrong project, PCB document, or schematic instance path
- duplicate exact document identities
- document disappearance or retargeting during readback
- malformed selected protobufs or empty KIIIDs
- selected object types not represented by the bundled stable KiCad protobuf model
- unsupported commands and unreachable/closed endpoints

## Intentionally unsupported

- heuristic protobuf wire decoding for unmodelled schematic symbols or other unknown objects
- treating references such as `C10` as identity (N3)
- selection mutation (N4), activation/reveal (N5), and cross-probe (N6)

The vendored KiCad 10 schematic schema models lines and labels but not schematic symbols. Such selections return `unsupported_capability`; they are not guessed from bytes or saved files.

## Verification

- `cargo test -p konnect-ipc --test mock_server_test` — 44 passed, 1 ignored
- `cargo test -p konnect-core editor_navigation --lib` — 6 passed
- `cargo test --workspace --all-targets` — pass (933 core tests; all non-environmental targets passed)
- `cargo clippy --workspace --all-targets -- -D warnings` — pass
- `xtask fix-doc-counts --check` — 21 toolsets, 223 registered tools, 230 total; current
- `git diff --check` — pass

## Remaining risks

- KiCad 10's schematic protobuf surface cannot serialize every selectable schematic object; callers receive typed unsupported results until the stable protocol expands
- the response lacks a document header; the before/after open-document readback detects context churn but cannot make the two transport calls atomic
- public request names and compatibility are awaiting maintainer direction in #395

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,9 @@ permissions: {}
 env:
   CARGO_TERM_COLOR: always
   BINARY_NAME: konnect
+  # Embedded into get_installation_info so release artifacts identify the
+  # exact source commit even when built outside a Git working tree.
+  KONNECT_BUILD_COMMIT: ${{ github.sha }}
 
 jobs:
   build:
--- a/DEV.md
+++ b/DEV.md
@@ -192,6 +192,11 @@ Konnect/
 - Cooperative lock files live under `KONNECT_STATE_DIR/locks` when that
   absolute override is set, otherwise under the platform local-data directory
   (`konnect/locks`). Reads never create files in the KiCad project.
+- Schematic writes also refuse while KiCad's sibling `~<name>.kicad_sch.lck`
+  exists. KiCad records only a username and hostname, so Konnect cannot prove
+  that a same-host or remote lock is stale; valid, foreign, empty, and malformed
+  locks all fail closed. The check runs before a transaction journal is created
+  and again at the final target-write boundary.
 - Multi-file schematic changes use project-local
   `.konnect-transaction-*.json` write-ahead journals. These journals contain
   complete before/after images and must be treated as sensitive project data.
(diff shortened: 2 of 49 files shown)


neusse:
Same coordination request here: please pause further work on this draft while we start with #369 and land the existing stack from the bottom up. #397 is stacked on #396 and all of its earlier prerequisites, so every change below it currently ripples through the whole branch. Once #396 has been rebuilt on the final accepted IPC base and the #395 public API decisions are settled, #397 can be restacked once and reviewed against stable prerequisites. Holding here will avoid repeated rebases and keep the eventual review and test evidence meaningful.

neusse:
The unique N2 portion of this cumulative draft has been reconstructed directly on the merged N1 foundation as #491.

#491 preserves the original authorship, retains the already-merged safety work and tests, and passes the complete local gate. This old stacked draft is now superseded and is being closed so there is only one active implementation path.

PR #398, opened 2026-08-30
feat(navigation): resolve exact editor targets

dubesinhower:
## Navigation MVP slice N3

Adds deterministic semantic resolution for one object in an explicitly named KiCad project, open document, editor, and schematic hierarchy instance.

### Why this belongs in the MVP

Selection mutation, activation, and cross-probing must not guess which duplicate reference, document, or hierarchy instance a caller meant. This slice establishes the exact target contract before any state-changing navigation work.

### Dependency

- Depends on #397 (N2 selection observation), which depends on #396 and the non-draft Priority 0 IPC foundation through #390.
- Tracks the provisional public semantic surface in #395.
- Does not depend on draft #391; read-only navigation does not use its stale board-lock policy.

### Behavior

- Requires an explicit project, document, editor, and schematic sheet-instance path where applicable.
- Resolves stable KIID/UUID identifiers first.
- Accepts a human reference such as `C10` only when it resolves to exactly one object; duplicate matches return structured candidates.
- Proves the exact live context with KiCad `GetOpenDocuments` and keeps that IPC evidence separate from saved-structure evidence.
- Refuses missing/stale objects, wrong projects/documents/sheets, duplicate KIIIDs, ambiguous references, closed editors, and unsupported live document identity without retargeting.

### Intentionally unsupported here

- No selection mutation, activation/reveal, cross-probing, or raw `RunAction` surface.
- The public MCP name and request schema remain provisional pending #395, so this PR is draft.

### Verification

- `cargo test -p konnect-core navigation_target --lib` — 4 passed
- `cargo test -p konnect-core editor_navigation --lib` — 8 passed
- `cargo test -p konnect-ipc --test mock_server_test` — 44 passed, 1 ignored
- `cargo test --workspace --all-targets` — passed (live-GUI-only tests remained ignored)
- `cargo clippy --workspace --all-targets -- -D warnings`
- `xtask fix-doc-counts --check` — 21 toolsets, 224 registered, 231 total
- `git diff --check`

### Remaining risks

- KiCad 10's stable schematic IPC schema does not serialize every selected object type, so later mutations must refuse unobservable readback rather than infer success.
- Live integration could not be exercised without a running KiCad IPC endpoint; coverage is hermetic and transport-faithful.
- Public compatibility details remain subject to the maintainer decision in #395.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,9 @@ permissions: {}
 env:
   CARGO_TERM_COLOR: always
   BINARY_NAME: konnect
+  # Embedded into get_installation_info so release artifacts identify the
+  # exact source commit even when built outside a Git working tree.
+  KONNECT_BUILD_COMMIT: ${{ github.sha }}
 
 jobs:
   build:
--- a/DEV.md
+++ b/DEV.md
@@ -192,6 +192,11 @@ Konnect/
 - Cooperative lock files live under `KONNECT_STATE_DIR/locks` when that
   absolute override is set, otherwise under the platform local-data directory
   (`konnect/locks`). Reads never create files in the KiCad project.
+- Schematic writes also refuse while KiCad's sibling `~<name>.kicad_sch.lck`
+  exists. KiCad records only a username and hostname, so Konnect cannot prove
+  that a same-host or remote lock is stale; valid, foreign, empty, and malformed
+  locks all fail closed. The check runs before a transaction journal is created
+  and again at the final target-write boundary.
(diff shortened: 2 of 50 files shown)


neusse:
The unique N3 portion of this cumulative draft has been reconstructed directly on current `main` as #492, after N2 landed in #491.

#492 preserves the original authorship, adapts the implementation to #189's current ownership-error contract, adds regression coverage for that adaptation, and passes the complete local gate. This old stacked draft is now superseded and is being closed so there is only one active implementation path.

PR #399, opened 2026-08-30
feat(navigation): verify editor selection mutations

dubesinhower:
## Navigation MVP slice N4

Adds semantic clear/add/remove selection operations for an exact KiCad editor context, with stable-KIID target validation and result-derived readback.

### Why this belongs in the MVP

Navigation needs a safe way to express selection intent without exposing raw actions or treating a successful transport round trip as proof that KiCad changed state.

### Dependency

- Depends on #398 (N3 exact target resolution), which depends on #397 and #396.
- Uses the typed Priority 0 IPC foundation through #390 and does not depend on draft #391.
- Tracks provisional public naming and compatibility in #395; this PR remains draft.

### Behavior

- Uses typed `ClearSelection`, `AddToSelection`, and `RemoveFromSelection` commands only.
- Requires an exact project/document/editor/sheet context; schematic instances are addressed by canonical KIID path.
- Resolves every add/remove KIID in the explicit saved structure before sending a mutation.
- Reads the exact selection before and after the command and requires the complete observed set to equal the requested transition.
- Returns before/after stable object identities and explicit live readback evidence.
- Refuses duplicate/empty IDs, stale/missing targets, wrong projects/documents/sheets, closed editors, unsupported commands/object decoding, and post-operation mismatch.

### Intentionally unsupported here

- Selection does not imply viewport reveal, centering, fitting, document activation, or cross-probing.
- No raw `RunAction` API is exposed.

### Verification

- `cargo test -p konnect-ipc --test mock_server_test` — 47 passed, 1 ignored
- `cargo test -p konnect-core editor_navigation --lib` — 12 passed
- `cargo test -p konnect-core --lib` — 943 passed, 9 ignored
- `cargo test --workspace --all-targets` — passed (live/environment-only tests remained ignored)
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo clippy -p konnect-ipc --tests -- -D warnings` after the final exact-list comparison tightening
- `xtask fix-doc-counts --check` — 21 toolsets, 225 registered, 232 total
- `git diff --check`

Hermetic transport tests exercise clear, add, and remove success; exact document headers; duplicate validation before transport; wrong/stale inherited context refusals; typed unsupported mapping; and a successful mutation response followed by unchanged selection, which returns `readback_mismatch`.

### Remaining risks

- KiCad 10's stable schematic protobuf cannot decode selected symbols, so symbol selection remains explicitly unsupported instead of unverifiable.
- KiCad selection replies have no document header; bounded before/after `GetOpenDocuments` checks are the available freshness boundary and are not atomic.
- No live endpoint was available, so real-GUI success is not claimed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,9 @@ permissions: {}
 env:
   CARGO_TERM_COLOR: always
   BINARY_NAME: konnect
+  # Embedded into get_installation_info so release artifacts identify the
+  # exact source commit even when built outside a Git working tree.
+  KONNECT_BUILD_COMMIT: ${{ github.sha }}
 
 jobs:
   build:
--- a/DEV.md
+++ b/DEV.md
@@ -192,6 +192,11 @@ Konnect/
 - Cooperative lock files live under `KONNECT_STATE_DIR/locks` when that
   absolute override is set, otherwise under the platform local-data directory
   (`konnect/locks`). Reads never create files in the KiCad project.
+- Schematic writes also refuse while KiCad's sibling `~<name>.kicad_sch.lck`
+  exists. KiCad records only a username and hostname, so Konnect cannot prove
+  that a same-host or remote lock is stale; valid, foreign, empty, and malformed
+  locks all fail closed. The check runs before a transaction journal is created
+  and again at the final target-write boundary.
 - Multi-file schematic changes use project-local
   `.konnect-transaction-*.json` write-ahead journals. These journals contain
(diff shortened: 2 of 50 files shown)


neusse:
Superseded by #493, which reconstructs this N4 change directly on the now-merged N3 baseline (#492 / current `main`). The replacement preserves the original implementation and @dubesinhower authorship, resolves the one current-main structured-error conflict without dropping either behavior, and has passed the full local gate. Closing this cumulative branch so review and CI follow the minimal sequential PR.

PR #402, opened 2026-08-30
fix(schematic): count pins in find_single_pin_nets, not labels

### LINKED ISSUE #430: fix(schematic): report nets that reach at most one pin, not one label
## Problem

`find_single_pin_nets` counts **label instances per net name**, not what the net
reaches. That is wrong in both directions, and the two errors hide each other:

- a net drawn the ordinary way — one label on a wire reaching two or more pins —
  has a label count of 1 and is reported as a single-pin net. On a real sheet
  this was 11 of 11 nets, every one of them correctly wired;
- a genuine single-pin net disappears the moment it carries a second label,
  because the count becomes 2.

The tool's own description promises "nets with only one label/connection", and
the review skill sends it looking for orphan labels and stubs left by deleted
components. It finds neither reliably.

## Accepted semantics

Report a net that reaches **at most one pin**, counted on the shared net graph:

- a net reaching zero pins is reported, with `pin_count` distinguishing it from
  one reaching a single pin. A label whose net reaches nothing is the orphan
  label and the deleted-component stub the tool is asked for, and
  `find_orphan_items` names the dangling wire end without ever naming the net;
- component pins and hierarchical sheet pins count as connections;
- a symbol that names a net rather than consuming one does not: a power symbol
  (`#PWR…`) and `PWR_FLAG` (`#FLG…`). Counting the symbol that named a rail
  would hide every rail reaching exactly one component pin, which is the defect
  the tool exists to find;
- labels of one name on wire segments that never meet are one net to KiCad, so
  the graph roots a name sits on pool before pins are counted;
- the report is ordered deterministically.

## Cross-sheet scope

The answer is per sheet. A net named by a global label, a hierarchical label, or
a power symbol can continue on another sheet, and the response must say so
rather than leave the caller to infer it from a label kind. Reporting such a net
is still correct — a rail reaching one pin here is worth showing — but it is a
lead, not a finding.

## Response compatibility

`single_pin_net_count`, and per net `net`, `x`, `y`, and `type`, keep their
current names and meanings. `type` stays the *first* label's kind, which is
precisely why it is not sufficient evidence: local labels are extracted first,
so a net carrying both a local and a global label reports `NetLabel` and says
nothing about the global one.

Additive fields: `pin_count`, `label_count`, `pins`, `label_types` (every
distinct kind, sorted), `cross_sheet_unverified`.

Which nets appear does change, because that is the defect being fixed. No
consumer can keep the old set — it was wrong in both directions — and a consumer
reading `single_pin_net_count` as a defect count gets a smaller, truer number
with no migration.

## Acceptance

- Deterministic tests cover: one label and two pins; two labels and one pin; two
  disconnected roots of one name pooling, both with one pin between them and
  with a pin on each; a power-symbol rail reaching one pin; a `PWR_FLAG` net; a
  power rail reaching many pins; a hierarchical sheet pin counting as a pin; a
  net reaching no pin; a net carrying a local *and* a global label; a locally
  named net not flagged cross-sheet; a child sheet's hierarchical net; report
  ordering.
- The fixtures are real KiCad output, force-resaved by a supported KiCad and
  reload-checked, with a parent/child pair for the hierarchy boundary. Expected
  pin counts are KiCad's netlist, not this crate's parser, and provenance is
  retained beside the fixture.
- Each guard — pin counting, power-symbol exclusion, root pooling, sheet-pin
  counting, the cross-sheet flag, and the distinct-kind set — is neutralized
  once and its tests fail.
- The response compatibility contract above is documented at the handler.
- The PR is rebuilt on current main and passes the complete gate.

pauliuszaleckas:
`find_single_pin_nets` counted *label instances* per net name and reported every net
carrying exactly one. Pins, wires and junctions were never consulted, so a net drawn the
ordinary way — one label on a wire that reaches two or more pins — came back as a
single-pin net. Redrawing a real board's schematic through the tools, all 11 nets it
reported were false positives: `VBUS` (five pins), `ACT`, `ACT_A`, `CC1`, `CC2`,
`PWR_LED_A`, and the five `*_INT` nets between U1 and its series resistors.

It missed the case it exists for just as readily — a net that really does reach one pin
is invisible as soon as it carries a second label.

Closes #430, which captures the semantics and the acceptance checklist. #249 was the
same pin blindness one tool over (`find_orphan_items`) and closed without reaching this
one.

### The count comes from the shared net graph

Connection points are grouped by their union-find root — the `ConnectivityIndex` and
`net_graph_for` machinery #323 gave the other connectivity tools — and a net is reported
when it reaches at most one. Each entry carries `pin_count`, `label_count` and the
`pins` that were counted, so the label count is still there to read and is no longer the
answer.

Two labels of one name on wire segments that never touch are still one net to KiCad, so
their roots pool before the count is taken. `SPLIT` in the fixture is the case that makes
pooling load-bearing: a pin under each of two disconnected roots, which KiCad resolves to
one 2-pin net and this tool must therefore not report.

### Two connection points are treated unalike

- **A hierarchical sheet pin counts.** A net leaving the sheet through one reaches
  whatever is on the other side; not counting it would report every such net, which is
  the same false-positive shape being fixed here.
- **A power symbol's own pin does not.** It names the rail rather than consuming it, and
  counting it would hide the rail that reaches exactly one component pin — the defect
  worth reporting. The rule is `is_power_symbol_reference`, KiCad's `#` prefix, now with
  one home for a heuristic `design_review` carried inline too. Deliberately *not*
  `LabelKind::PowerSymbol`: that is derived from the `(power)` marker plus a `power_in`
  pin, so `PWR_FLAG`'s `power_out` pin passes it — `#FLG…` is what actually excludes it.

So `find_orphan_items` and this tool disagree about a power symbol's pin on purpose — an
unwired `#PWR01` is an orphan there and not a consumer here. The handler doc says so,
since the whole point of `sch_connectivity` is that these tools cannot drift silently.

### Zero counts as well as one

A label whose net reaches nothing is the orphan label and the deleted-component stub the
review skill sends this tool looking for, and the label count did report both.
`find_orphan_items` names the dangling wire end but never the net, so reporting only `1`
would have quietly retired a documented behaviour. `pin_count` carries the real number.

### Shared rather than local

The label-to-root walk is now `sch_connectivity::label_roots`. `find_shorted_nets` reads
that same relation from the other end — root-first, since more than one name on a root
is a short — and the two drifting apart is exactly what that module exists to prevent.

The report is also sorted by net name. It came out of a `HashMap` in whatever order it
held them, so two runs of one binary over one sheet could disagree.

### Cross-sheet scope is stated, not inferred

The answer is per sheet, so a net named by a global or hierarchical label may well
continue on another one. `type` alone cannot say so: local labels are extracted first, so
a net carrying a global label too reports `NetLabel` and hides the one fact a caller
needs. Two additive fields carry the evidence instead:

- **`label_types`** — every distinct kind that named the net, sorted. `MIXED` reports
  `["GlobalLabel", "NetLabel"]` where `type` alone says `NetLabel`.
- **`cross_sheet_unverified`** — true when any label naming the net can carry it off this
  sheet. Power symbols included, since KiCad makes a global label out of one: `VCC` is
  flagged even though the project-wide netlist happens to agree it reaches one pin,
  because a single sheet cannot know that.

A flagged net is still reported — a rail reaching one pin here is worth showing — but the
flag marks it a lead rather than a finding. Ranking a `NetLabel` net above a hierarchical
one is the follow-up.

### Response compatibility

Every existing field keeps its name and meaning: `single_pin_net_count`, and per net
`net`, `x`, `y`, and `type`. The additive fields are `pin_count`, `label_count`, `pins`,
`label_types`, and `cross_sheet_unverified`.

The one real change is which nets appear, because that was the defect: membership went
from "exactly one label instance names it" to "the net reaches at most one pin", and
`single_pin_net_count` counts the latter. A consumer reading it as a defect count gets a
smaller, truer number and needs no migration; one that had learned to ignore this tool's
noise can stop. No consumer can keep the old set, since it was wrong in both directions.
No tool, argument, or existing field was renamed or removed.

Recorded in `docs/API_MIGRATIONS.md` and at the handler.

### Fixtures

A KiCad 10.0.5 parent/child pair, built through Konnect against the stock `Device:R` and
`power:` libraries, then parsed and force-resaved by eeschema — the same recipe, and the
same reason, as `structural_scans_kicad10`:

```
kicad-cli sch upgrade --force single_pin_nets.kicad_sch
kicad-cli sch upgrade --force single_pin_nets_child.kicad_sch
```

What is committed is that Eeschema serialization: `(generator "eeschema")`,
`(version 20260306)`, KiCad's own library records and field positions. Provenance and the
whole case table are in `single_pin_nets.README.md`.

**Every expected pin count is KiCad's, not this crate's** — taken from the netlist
exporter, which is also the oracle the tests assert against:

| Net | KiCad resolves | Shape |
|---|---|---|
| `/TWO_PIN` | 2 — R1.2, R2.1 | the false positive |
| `/LONE` | 1 — R3.2 | two labels, two disconnected roots, one pin |
| `/SPLIT` | 2 — R4.2, R5.1 | the same shape with a pin on each root |
| `VCC` | 1 — R6.2 | `#PWR001` is not a pin |
| `/FLAGGED` | 1 — R7.2 | `#FLG01` is not a pin either |
| `/SHEET_NET` | 2 — R8.2, R10.2 | across the sheet boundary |
| `MIXED` | 1 — R9.2 | local *and* global label on one net |
| `GND` | 9 | the rail that is not a defect |
| `/Child/CHILD_LOCAL` | 2 | a local net the child answers in full |
| `STUB` | absent | reaches no pin, so there is no net to export |

`kicad-cli sch erc` independently agrees: `isolated_pin_label` on both `LONE` labels and
both `MIXED` labels, `label_dangling` on `STUB`, and no violation for `SPLIT`, `TWO_PIN`,
`SHEET_NET` or `GND`.

The dispatch boilerplate in this file is down one copy: `power_symbol_net_tests`' helper
is hoisted to a module both test modules use.

### Negative controls

Six guards neutralized one at a time against the 15 `single_pin_net_tests`, on this exact
head:

| Neutralized | Tests failing |
|---|---|
| count label instances again | 8 |
| power-symbol / `#FLG` exclusion | 4 |
| only the lowest root counts | 2 |
| cross-sheet flag always false | 3 |
| sheet pins do not count | 2 |
| `label_types` keeps only the first kind | 1 |

Restored tree: 15/15 pass.

### Validation

Run on this exact head, Fedora 44, rustc/cargo 1.96.0 (the pinned toolchain), KiCad
10.0.5:

| Command | Result |
|---|---|
| `cargo test --workspace --locked --lib --tests` | 1687 passed, 0 failed, 24 ignored |
| `cargo test --workspace --locked --doc` | 8 passed, 0 failed, 2 ignored |
| `cargo clippy --workspace --locked -- -D warnings` | clean |
| `cd crates/schematic-viewer && cargo test` | 20 passed, 0 failed |
| `kicad-cli sch export netlist --format kicadsexpr` on the fixture | table above |
| `kicad-cli sch erc --format json --severity-all` on the fixture | as described above |

**Not run:** the 24 `#[ignore]`d tests, which need a live KiCad IPC session,
Freerouting, or a discovered `kicad-cli` — none exercise `find_single_pin_nets`. The
`e2e-kicad` workflow is weekly/dispatch and not a per-PR gate.

### Risk and rollback

Read-only: `find_single_pin_nets` opens a `.kicad_sch`, parses it, and returns JSON. No
write path, no IPC, no `.kicad_sch` is modified, so a wrong answer cannot corrupt a
project. The blast radius is one tool's report and the `design_review` audit that reads
it — `is_power_symbol_reference` replaces an inline `#` check there with the identical
rule.

Rollback is `git revert` of the single commit. It restores the previous membership rule
and drops the five additive fields; nothing persisted needs undoing, and no caller state
migrates in either direction.

### DIFF
--- a/crates/konnect-core/src/tools/design_review.rs
+++ b/crates/konnect-core/src/tools/design_review.rs
@@ -1338,8 +1338,7 @@ async fn handle_check_bom_health_file(
             None => continue,
         };
 
-        // Skip power symbols and sub-units
-        if reference.starts_with('#') {
+        if crate::tools::is_power_symbol_reference(reference) {
             continue;
         }
         total_components += 1;
--- a/crates/konnect-core/src/tools/mod.rs
+++ b/crates/konnect-core/src/tools/mod.rs
@@ -729,6 +729,18 @@ pub(crate) fn placed_pins(
         .collect()
 }
 
+/// Whether a reference belongs to a symbol that names a net rather than
+/// consuming one — a power symbol, a `PWR_FLAG`. KiCAD prefixes those with `#`
+/// and keeps them out of the netlist as components.
+///
+/// Deliberately not `LabelKind::PowerSymbol`, which `extract_power_symbol_labels`
+/// derives from the `(power)` marker plus a `power_in` pin: `PWR_FLAG`'s pin is
+/// `power_out`, so that test lets it through, and a caller counting what a net
+/// actually reaches wants it out too.
+pub(crate) fn is_power_symbol_reference(reference: &str) -> bool {
+    reference.starts_with('#')
+}
+
 /// [`placed_pins`], grouped under the instance that placed each unit, for
 /// callers that report pins by name rather than position. The whole instance
 /// is returned because a caller reporting a pin usually wants its reference
--- a/crates/konnect-core/src/tools/sch_analysis.rs
+++ b/crates/konnect-core/src/tools/sch_analysis.rs
@@ -5,20 +5,22 @@
 
 use crate::mcp::protocol::CallToolResult;
 use crate::tool;
-use crate::tools::sch_connectivity::{net_graph_for, pt_key, ConnectivityIndex};
+use crate::tools::sch_connectivity::{label_roots, net_graph_for, pt_key, ConnectivityIndex};
 use crate::tools::{
-    get_path, opt_f64, placed_pins_by_reference, require_f64, require_str, ToolContext, ToolDef,
+    get_path, is_power_symbol_reference, opt_f64, placed_pins_by_reference, require_f64,
+    require_str, ToolContext, ToolDef,
 };
 use konnect_schematic_editor as cse;
 use konnect_sexp::{
     geometry::{point_on_segment, points_coincident},
     schematic::{
-        extract_all_net_labels, extract_labels, extract_symbol_instances, extract_wires,
-        find_lib_symbol, read_schematic, symbol_bounds_for_instance, LabelKind, SymbolBounds, Wire,
+        extract_all_net_labels, extract_labels, extract_sheet_pins, extract_symbol_instances,
+        extract_wires, find_lib_symbol, read_schematic, symbol_bounds_for_instance, Label,
+        LabelKind, LibPin, SymbolBounds, Wire,
     },
 };
 use serde_json::json;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 pub fn tools() -> Vec<ToolDef> {
     vec![
@@ -152,7 +154,12 @@ pub fn tools() -> Vec<ToolDef> {
         ),
         tool!(
             "find_single_pin_nets",
-            "Find nets with only one label/connection — often indicates a missing counterpart.",
+            "Find nets that reach at most one pin — often a missing counterpart, an orphan \
+             label, or a stub left by a deleted component. Component pins and hierarchical \
+             sheet pins count; a power symbol's own pin names the rail rather than consuming \
+             it and does not. Reports the pin and label counts, and every label kind that \
+             named the net. Read per sheet: a net a global, hierarchical or power label can \
+             carry off this one is flagged cross_sheet_unverified.",
             json!({ "type": "object",
                 "properties": { "schematic": { "type": "string" } },
                 "required": ["schematic"] }),
@@ -551,10 +558,9 @@ async fn handle_find_shorted_nets(
     let wires = extract_wires(&tree);
     let labels = extract_all_net_labels(&tree);
     let mut g = net_graph_for(&tree, &wires, &labels);
-    let mut root_nets: HashMap<(i64, i64), Vec<String>> = HashMap::new();
-    for l in &labels {
-        let root = g.find(pt_key(l.x, l.y));
-        root_nets.entry(root).or_default().push(l.net.clone());
+    let mut root_nets: HashMap<(i64, i64), Vec<&str>> = HashMap::new();
+    for (root, label) in label_roots(&mut g, &labels) {
+        root_nets.entry(root).or_default().push(label.net.as_str());
     }
     let shorts: Vec<serde_json::Value> = root_nets
         .into_values()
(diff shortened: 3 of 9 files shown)


neusse:
Thanks for tackling this—the core direction is useful: count actual connected pins through the shared graph instead of treating label count as connectivity. I reviewed head `ea0984183f6c09545e3d68c5096acc92f09e1926` against current `main`. Before this can merge, please address the following:

1. **Open and link a focused issue first.** This is a non-trivial behavior and public-response change, but the PR currently says “Not filed as an issue.” Please capture the intended semantics and acceptance checklist in an issue, then make this PR the terminal `Closes #N` path if it satisfies the full checklist.

2. **Replace the synthetic schematic fixture with real KiCad output.** `single_pin_nets.kicad_sch` is hand-authored (`uuid "root"`, `w1`, `u1`, etc.) and KiCad 10.0.6 rejects it with “Failed to load schematic.” Our parser accepting its own approximation does not prove compatibility. Please create/save the fixture in a supported KiCad version, verify it reopens or passes `kicad-cli`, and retain that provenance in the test comments. A real parent/child pair is needed for the hierarchy boundary.

3. **Add the missing semantic cases.**
   - Two physically disconnected roots carrying the same net name. The PR says those roots pool, but the current `LONE` labels sit on the same wire and do not test that behavior.
   - A real `PWR_FLAG` case. The implementation relies on `#FLG...` being excluded like `#PWR...`, but only `power:VCC` is tested.
   - A mixed local plus global/hierarchical-label net.

4. **Do not hide cross-sheet uncertainty in `first_label.kind`.** The response currently returns only the first label kind. Because local labels are extracted first, a net also carrying a global or hierarchical label can be reported as `NetLabel`, hiding the exact per-sheet limitation callers need to see. Preserve the existing `type` field for compatibility, but add deterministic evidence such as all distinct `label_types` and/or an explicit `cross_sheet_unverified` scope flag, with tests.

5. **Document the response compatibility contract.** State explicitly that the existing fields remain, which fields are additive (`pin_count`, `label_count`, `pins`, plus any scope evidence), how `single_pin_net_count` semantics change, and whether consumers need migration guidance.

6. **Record the required negative control.** Temporarily neutralize the new pin-count guard and show that the regression tests fail. The current PR describes tests but does not include the required mutation/neutering evidence.

There were no material code-structure or scope-creep findings: sharing `label_roots`, centralizing the power-reference rule, and stable sorting all fit the proposed change. Once the issue, real KiCad fixtures, missing cases, scope evidence, and negative-control proof are present, this should be straightforward to re-review.

pauliuszaleckas:
Thanks — this was a good list, and the fixture objection was the right one. Rebuilt on current main (`d73adfa`) and pushed. Point by point:

**1. Focused issue.** Opened #430 with the semantics, the cross-sheet scope, the response-compatibility contract, and an acceptance checklist. This PR is the terminal `Closes #430` path.

**2. Real KiCad fixtures.** The hand-authored file is gone. `single_pin_nets.kicad_sch` and a new child `single_pin_nets_child.kicad_sch` were built through Konnect against KiCad's stock `Device:R` and `power:` libraries, then parsed and force-resaved by KiCad 10.0.5 — the same recipe, and the same reason, as `structural_scans_kicad10`:

```
kicad-cli sch upgrade --force single_pin_nets.kicad_sch
kicad-cli sch upgrade --force single_pin_nets_child.kicad_sch
```

What is committed is that Eeschema serialization: `(generator "eeschema")`, `(version 20260306)`, KiCad's own library records and field positions. Provenance and the whole case table are in `single_pin_nets.README.md`.

More importantly, **the expected pin count of every net is now KiCad's, not this crate's**. Reload-checked with the netlist exporter, which is also the oracle the tests assert against:

| Net | KiCad resolves | Shape |
|---|---|---|
| `/TWO_PIN` | 2 — R1.2, R2.1 | the false positive |
| `/LONE` | 1 — R3.2 | two labels, two disconnected roots, one pin |
| `/SPLIT` | 2 — R4.2, R5.1 | the same shape with a pin on each root |
| `VCC` | 1 — R6.2 | `#PWR001` is not a pin |
| `/FLAGGED` | 1 — R7.2 | `#FLG01` is not a pin either |
| `/SHEET_NET` | 2 — R8.2, R10.2 | across the sheet boundary |
| `MIXED` | 1 — R9.2 | local *and* global label on one net |
| `GND` | 9 | the rail that is not a defect |
| `/Child/CHILD_LOCAL` | 2 | a local net the child answers in full |
| `STUB` | absent | reaches no pin, so there is no net to export |

`kicad-cli sch erc` independently agrees: `isolated_pin_label` on both `LONE` labels and both `MIXED` labels, `label_dangling` on `STUB`, and no violation for `SPLIT`, `TWO_PIN`, `SHEET_NET` or `GND`.

**3. Missing semantic cases.** All three, plus the ones they implied:

- *Two disconnected roots, same name* — `LONE` is now exactly that (two labels on segments that never touch, one pin between them), so pooling is load-bearing for it. `SPLIT` is the complement: a pin under each root, which KiCad resolves to one 2-pin net and the tool must therefore not report. That row of the netlist is the direct evidence for the pooling rule.
- *A real `PWR_FLAG`* — `#FLG01` on the `FLAGGED` net, with its own test. Worth noting it is not a `LabelKind::PowerSymbol`: its pin is `power_out`, so the `(power)` test lets it through and `#FLG…` is what actually excludes it.
- *Mixed local + global* — `MIXED`, and the parent/child pair gives the hierarchical case from the child's side.
- Also added `GND` (nine pins, named by a power symbol) as the control for a power net that is not a defect.

**4. Cross-sheet uncertainty is no longer hidden.** You were right that `first_label.kind` conceals it. `type` stays exactly as it was for compatibility, and two additive fields carry the evidence:

- `label_types` — every distinct kind that named the net, sorted. `MIXED` reports `["GlobalLabel", "NetLabel"]` where `type` alone says `NetLabel`.
- `cross_sheet_unverified` — true when any label naming the net can carry it off this sheet. That includes power symbols, since KiCad makes a global label out of one; `VCC` is flagged even though the project-wide netlist happens to agree it reaches one pin, because a single sheet cannot know that.

The net is still reported when flagged — a rail reaching one pin here is worth showing — but the flag marks it as a lead rather than a finding.

**5. Compatibility contract** is documented at the handler and in #430. Every existing field keeps its name and meaning; the additive ones are listed; and the one real change is stated plainly — which nets appear, because that was the defect. No consumer can keep the old set, since it was wrong in both directions, and one reading `single_pin_net_count` as a defect count gets a smaller, truer number with no migration.

**6. Negative controls.** Six guards neutralized one at a time, against the 15 tests:

| Neutralized | Tests failing |
|---|---|
| count label instances again | 8 |
| power-symbol/`#FLG` exclusion | 4 |
| only the first root counts | 3 |
| cross-sheet flag always false | 3 |
| sheet pins do not count | 2 |
| `label_types` keeps only the first kind | 1 |

Restored tree: 15/15 pass.

Gate on the rebased head: `cargo test --workspace --locked --lib --tests`, `--doc`, `clippy -D warnings`, `fmt --check`, plus `crates/schematic-viewer` — all clean.

neusse:
This is now technically very close. I re-reviewed exact head `29e18175f0f9de8f6641b03e07ba9a3816f67536` against #430 and current `main`. The real KiCad fixtures, hierarchy cases, deterministic evidence fields, and negative controls satisfy the implementation acceptance criteria, and all ten checks are green.

Two repository-contract items remain before merge:

1. Add an entry to `docs/API_MIGRATIONS.md` for the changed `find_single_pin_nets` response/semantics. Please state that existing fields remain, list the additive `pin_count`, `label_count`, `pins`, `label_types`, and `cross_sheet_unverified` fields, and note that `single_pin_net_count` now counts nets with at most one actual pin rather than label instances.
2. Update the PR body with the exact final validation commands/results, skipped or unavailable checks, risk/rollback statement, and the negative-control summary already supplied in the comment above.

No algorithmic rewrite or new fixture work is being requested. Once those two documentation items are pushed and the exact-head checks are green, this should be ready for final review.

pauliuszaleckas:
Both documentation items are pushed. Rebased onto current `main` (`ab33781`); head is now `60c2322`.

**1. `docs/API_MIGRATIONS.md`.** New entry, *`find_single_pin_nets` counts pins, not labels (minor release)*. It states that every existing field — `single_pin_net_count`, and per net `net`, `x`, `y`, `type` — keeps its name and meaning, lists the five additive fields (`pin_count`, `label_count`, `pins`, `label_types`, `cross_sheet_unverified`) with what each carries, and calls out the one real change: `single_pin_net_count` now counts nets reaching at most one actual pin rather than nets named by exactly one label instance. The commit body points at the file.

**2. PR body.** Rewritten with the exact validation commands and results, the checks not run, the risk/rollback statement, and the negative-control table.

Two things worth flagging rather than leaving implicit, since the head moved:

- **The negative controls were re-run on this exact head**, not carried over. Five of the six match the earlier counts. The "only the first root counts" control changed: it had iterated a `HashSet` and taken one root, so its failure count was not reproducible run to run. Neutralizing it deterministically instead — sort the roots, keep the lowest — fails **2** tests, stably across repeated runs, so that is the number in the table. The other five are 8 / 4 / 3 / 2 / 1 as before. Restored tree: 15/15.
- **The fixture oracles were re-verified against KiCad 10.0.5 on this head**, not quoted from the earlier comment. `kicad-cli sch export netlist` returns the same ten-row table, `STUB` still absent. `kicad-cli sch erc --severity-all` still reports `isolated_pin_label` on both `LONE` and both `MIXED` labels and `label_dangling` on `STUB`, with nothing on `SPLIT`, `TWO_PIN`, `SHEET_NET` or `GND`.

Full gate green on `60c2322`: workspace tests 1687 passed / 0 failed / 24 ignored, doctests 8 passed, `clippy -D warnings` clean, `fmt --check` clean, and `crates/schematic-viewer` 20 passed. The 24 ignored need a live KiCad IPC session, Freerouting, or a discovered `kicad-cli`; none exercise `find_single_pin_nets`.

neusse (approved):
Maintainer review complete on exact head `60c232283c3eda712017dbc14191e170f1cb6da9`: approved.

The implementation matches #430's graph-based definition and closure checklist. It pools shared named-net roots, counts component and hierarchical pins, excludes power-symbol naming pins, reports zero-pin stubs, preserves compatibility while adding the required evidence fields, and explicitly marks cross-sheet uncertainty. The real KiCad 10 fixture, 15 focused tests, six negative controls, migration documentation, deterministic ordering, and all ten required CI checks provide proportionate evidence.

Non-blocking compatibility note: `label_types` is serialized from the Rust enum's Debug names. Treat any future enum-variant rename as a public response-contract change, or stabilize that representation before renaming.

PR #394, opened 2026-08-30
fix(schematic): derive component mutations from readback

dubesinhower:
## Summary

Schematic mutations could report success when a committed symbol kept its UUID/reference but had the wrong unit, library, hierarchy path, position, rotation, or property value. This PR binds the selected target before writing, reloads the committed schematic once per result set, compares it with the intended mutation, and derives success responses from that observed state.

Closes #387. This remains the terminal file-only schematic mutation PR, directly based on current main.

## Behavior

- Add, batch add, and power-symbol placement verify requested reference/value, library and unit, document, project/hierarchy identity, coordinates, and rotation. Component placement uses its existing 1.27 mm snap; power placement preserves its existing requested-coordinate behavior.
- Move, rotate, edit/rename, annotation, and grouping preserve the bound UUIDs, unit/library and hierarchy identities, and untouched property values. Expected positions/angles come from the original units plus the requested transform. Requested field values are checked on every unit.
- A combined rename and field edit now applies subsequent edits to the renamed reference. The native ECC83 regression covers rename, Value, Footprint, Datasheet, and MPN in one request, followed by grouping.
- Missing, changed, malformed, wrong-document, or mismatched intent returns `stale_target`. Duplicate UUID/reference-unit/property identities, duplicate instance identities, inconsistent project/instance-unit records, or disagreement between a component's units about hierarchy identity returns `ambiguous_target` during component-target resolution/readback. Records are checked before sorting or choosing the observed placement project.
- Existing response fields remain. Observed document, unit, UUID, library, geometry, hierarchy and field evidence is additive minor-release work; DEV.md and API_MIGRATIONS.md document errors and compatibility.

## Exact revision

- Base: `9fe3d60eca9a3d63ccb37b1cf649e83355ff5d0e`
- Head: `a4e80f6ab46cf8a59475921dfd29325c4b2db418`
- Two focused commits: the original committed-state response change and the review correction. No descendant restack.
- Seven files: DEV.md, error.rs, sch_batch.rs, sch_components.rs, sch_wiring.rs, schematic_placement_tests.rs, and API_MIGRATIONS.md.

## Validation

Final Windows gate after restoring all negative controls:

| Command | Exit |
| --- | --- |
| `cargo fmt --all -- --check` | 0 |
| `cargo clippy --workspace --locked --all-targets -- -D warnings` | 0 |
| `cargo test --workspace --locked --lib --tests` | 0 |
| `cargo test --workspace --locked --doc` | 0 |

[...]

PR #407, opened 2026-08-31
fix(pcb): tell a KiCad without this board open apart from a refusal

### LINKED ISSUE #426: fix(pcb): fail closed before classifying a requested board as not open
## Problem

KiCad being reachable while the requested board is positively not open is different from an IPC rejection. That classification may permit a guarded direct-file fallback, but only when every reported open PCB document identity was inspected reliably.

The implementation in PR #407 skips open PCB documents whose paths cannot be resolved and can then conclude BoardNotOpen. That converts ambiguous live-editor evidence into permission to mutate the saved file.

## Accepted behavior

- Introduce a typed BoardNotOpen result only when the requested canonical board is absent from a complete, trustworthy set of open PCB document identities.
- If any relevant open PCB document path is empty, bare, malformed, unresolved, duplicate, or otherwise not comparable to the requested path, fail closed with a structured ambiguity/wrong-document error.
- Never infer BoardNotOpen from open-document order or from discarded records.
- Preserve process-lifetime stale-board refusal: a requested board observed live earlier cannot enter file fallback merely because it later disappears.
- Keep a reachable KiCad holding a positively identified different board distinct from transport-unreachable and rejected cases.
- Derive warnings and response evidence from the classification actually observed.

## Acceptance

- Requested board absent from a complete trustworthy list permits only the existing guarded fallback.
- Empty open list is handled explicitly.
- Requested board present, duplicated, changed, or disappeared refuses appropriately.
- Bare/unresolved/malformed document identities fail closed and never invoke a file-write closure.
- A board observed live earlier remains protected.
- Negative-control tests prove that skipping an unresolved document causes the suite to fail.
- Real KiCad evidence covers the document-path forms emitted on supported platforms.
- PR #407 is rebuilt on current main, passes the complete gate, and uses Closes #426 only when the checklist is complete.

This issue defines the classification seam that #390 must consume rather than duplicate.

pauliuszaleckas:
Closes #426.

## Problem

KiCad being reachable while the requested board is positively not open is not an IPC rejection. That observation may permit the existing guarded direct-file fallback, but only after every reported open PCB document has been resolved into a complete, trustworthy set of comparable identities.

The old behavior classified both “another board is open” and “nothing is open” as `Rejected`. An earlier version of this PR then skipped identities it could not resolve, which could turn ambiguous live-editor evidence into permission to mutate the saved file.

## Current implementation

- Adds typed `BoardNotOpen` only when the requested canonical board is absent from a complete, comparable set of open PCB document identities.
- Adds typed `AmbiguousOpenBoards` and public structured error kind `ambiguous_open_board` for empty, bare, malformed, unresolved, duplicate, or otherwise uncomparable relevant identities.
- Treats an empty open-document list explicitly as `BoardNotOpen`.
- Lets a positive requested-board match select the live IPC path even if another document is unreadable; ambiguity never unlocks file mode.
- Preserves process-lifetime stale-board refusal when a board observed live later disappears or IPC becomes unreachable.
- Returns `fallback_reason.kind` (`board_not_open` or `transport_unreachable`) and `fallback_reason.message` from successful hybrid file mutations; the warning is derived from the same observed classification.
- Documents the runtime boundary in `docs/ARCHITECTURE.md` and the additive response/error contract in `docs/API_MIGRATIONS.md`.

## Acceptance accounting

- [x] Requested board absent from a complete trustworthy list permits only the existing guarded fallback.
- [x] Empty open list is handled explicitly.
- [x] Requested board present, duplicated, changed, or disappeared refuses or routes safely as appropriate.
- [x] Bare, unresolved, malformed, missing, and duplicate document identities fail closed and never invoke a file-write closure.
- [x] A board observed live earlier remains protected.
- [x] Negative controls prove that skipping unresolved records and collapsing the new fallback evidence are caught by tests.
- [x] Windows live KiCad evidence covers the emitted bare filename plus absolute project-directory form on exact predecessor head `ced0754e475f4d38b17f88827421882496b734a1` with KiCad 10.0.5; the reconstruction does not change identity parsing.
- [x] Reconstructed on current `upstream/main` base `c22cef1ca1cb6dc6c89c7cb1b1838f0d6f329d71`.

## Validation

Current reconstructed head: `e1257025441ac82bde0ea63e5051869181e8d1ea`.

Local required gate:

- `cargo test --workspace --locked --lib --tests` — pass

Focused `konnect-ipc`, PCB-board, and PCB-component tests also pass. For the added response evidence, a negative control temporarily reported a positive `BoardNotOpen` observation as `transport_unreachable`; `a_kicad_holding_another_project_edits_this_board_file` failed on the exact field mismatch and passed again after restoration.

Hosted CI on the reconstructed head is the merge gate.

## Validation debt

Linux and macOS live-GUI `GetOpenDocuments` captures remain unavailable. Hosted unit tests cover their path dialects, but that is not represented as live evidence. This is recorded validation debt under the risk-proportionate contribution policy: the implementation fails closed for any form it cannot compare, Windows supplies a real KiCad observation, and no absent-platform inference is used to unlock a write.

The original contributor commit and authorship are preserved; the reconstruction resolves drift from the current IPC endpoint-discovery/redaction work and adds one focused maintainer completion commit for response evidence and documentation.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -262,6 +262,7 @@ Tool-call failures are typed via the `ToolErrorKind` enum in `crates/konnect-cor
 | `conflict` | The file changed, a write would replace existing paths, or schematic project ownership cannot be proven uniquely — carries the affected paths; ownership conflicts include the schematic directory and all candidate roots |
 | `ambiguous_target` | More than one observed UUID, reference/unit, field, or instance identity matches the requested target, or component instance records conflict across projects, units, or hierarchy paths — carries `target` and the stable `candidates`; the caller must choose rather than retrying blindly |
 | `stale_target` | Saved symbol instance metadata disagrees with the proven hierarchy, or component readback lacks the expected document, UUID/reference, unit, library, hierarchy, position, rotation, or property value — carries `target` and `reason`; preflight refuses before writing, while a readback failure may follow a committed write |
+| `ambiguous_open_board` | KiCad answered, and its open-document list could not be read as a complete set of comparable board identities — carries `path`; neither the live nor the file path may run |
 | `handler_error` | Catch-all for unmigrated `anyhow::Error` returns |
 
 ### Producing structured errors in a handler
--- a/crates/konnect-core/src/mcp/error.rs
+++ b/crates/konnect-core/src/mcp/error.rs
@@ -64,6 +64,10 @@ pub enum ToolErrorKind {
     /// A board was live earlier in this server process, but IPC is now gone;
     /// its saved file may be stale relative to lost editor state.
     UnsafeFileFallback { path: String },
+    /// KiCad answered, and its open-document list could not be read as a
+    /// complete set of comparable board identities — so whether it holds this
+    /// board is unknown, and neither a live edit nor a file edit is safe.
+    AmbiguousOpenBoard { path: String },
     /// Catch-all for handler `anyhow::Error` that hasn't been migrated yet.
     /// Eventually each variant above subsumes a subset of these.
     HandlerError { reason: String },
@@ -83,6 +87,7 @@ impl ToolErrorKind {
             Self::AmbiguousTarget { .. } => "ambiguous_target",
             Self::StaleTarget { .. } => "stale_target",
             Self::UnsafeFileFallback { .. } => "unsafe_file_fallback",
+            Self::AmbiguousOpenBoard { .. } => "ambiguous_open_board",
             Self::HandlerError { .. } => "handler_error",
         }
     }
@@ -215,6 +220,7 @@ mod tests {
                 reason: "r".into(),
             },
             ToolErrorKind::UnsafeFileFallback { path: "p".into() },
+            ToolErrorKind::AmbiguousOpenBoard { path: "p".into() },
             ToolErrorKind::HandlerError { reason: "r".into() },
         ];
         for kind in kinds {
--- a/crates/konnect-core/src/tools/pcb_board.rs
+++ b/crates/konnect-core/src/tools/pcb_board.rs
@@ -188,12 +188,34 @@ pub(crate) enum BoardWrite<T = ()> {
     /// IPC call returned, for tools that echo it back — the placed footprint,
     /// for instance.
     Ipc(T),
-    /// No live KiCad on this transport and this board was not observed live
+    /// No live KiCad is holding this board and it was not observed live
     /// during the current server session; proceed with the S-expression path.
-    File,
+    /// Carries *why*, because the caller's user has a different next move for
+    /// each — start KiCad, or open this board in the one already running.
+    File(NoLiveBoard),
     /// KiCAD answered and refused. The caller must return this result and must
     /// NOT touch the file.
     Refused(CallToolResult),
+}
+
+/// Why no live KiCad took the write, for the callers that say so.
+#[derive(Debug)]
+pub(crate) enum NoLiveBoard {
+    /// The request never reached KiCad.
+    Unreachable,
+    /// KiCad answered and holds another project, or none. Carries that
+    /// answer, which names the boards it does hold.
+    NotOpen(String),
+}
+
+impl NoLiveBoard {
+    /// The premise sentence an IPC-only tool leads its refusal with.
+    pub(crate) fn premise(&self) -> String {
+        match self {
+            Self::Unreachable => "KiCad IPC is unreachable.".to_string(),
+            Self::NotOpen(answer) => format!("KiCad is reachable but {answer}."),
+        }
+    }
 }
 
 /// Run `f` over IPC against the board named by `board_path`, deciding what the
@@ -210,6 +232,9 @@ pub(crate) enum BoardWrite<T = ()> {
 ///   classification, never a text match. A KiCad that answers — even with an
 ///   error — fails closed. An unreachable transport permits the file path only
 ///   when this server has never observed the requested board live.
+/// * A reachable KiCAD that does not hold this board is a third answer, not a
+///   refusal: it has no unsaved state for a board it never opened, so the file
+///   is authoritative and the edit proceeds there.
 pub(crate) async fn attempt_ipc_write<T, F>(
     ctx: &ToolContext,
     board_path: &std::path::Path,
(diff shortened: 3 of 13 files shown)


neusse:
The positive BoardNotOpen distinction is useful, but the current head is not safe to merge because it can skip an unresolved open PCB document path and then treat the requested board as proven absent. That changes ambiguous live-editor evidence into permission for a direct-file mutation.

I opened #426 as the focused contract for this PR. Please:

- link this PR with Closes #426;
- rebuild on current main;
- classify BoardNotOpen only from a complete set of trustworthy canonical PCB document identities;
- fail closed for empty/bare/malformed/unresolved/duplicate identities rather than dropping them;
- prove those cases never invoke the write closure and preserve board bytes;
- retain the existing process-lifetime stale-board refusal;
- add negative-control and real-KiCad document-path evidence;
- rerun all required checks on the reconstructed head.

After #407/#426 lands, #390 can bind operations to the exact positively identified DocumentSpecifier without duplicating this fallback decision.

pauliuszaleckas:
You were right, and the hole was exactly where you pointed. Rebuilt on current main (`d73adfa`) and pushed.

The defect in one line: `find_open_board` resolved each open document, **skipped the ones it could not place on disk**, and then concluded from what was left — so a record with no identifier, an empty name, or KiCad's own bare filename without the project directory that places it turned "we could not tell" into "the saved file is authoritative". `BoardNotOpen` is what unlocks a file write, so absence has to be proven, not inferred from a list that was never read in full.

**Complete-set classification.** Every open document is resolved before anything is decided, and the reasons travel with the identities rather than being dropped where they are produced. Then:

- exactly one identity matches → `Ok(doc)`;
- more than one matches → ambiguous (there is no single document an edit would reach);
- no match, and any identity is unreadable → ambiguous;
- no match, and any two identities are equal → ambiguous (a list Konnect does not model is not one absence can be read from);
- no match over a complete, comparable set → `BoardNotOpen`, as before;
- empty list → `BoardNotOpen` explicitly, which is the freshly launched editor.

A positive match still wins even when *another* document is unreadable: that sends the edit to KiCad, which is the safe direction.

**Fail-closed, and named as itself.** New `konnect_ipc::AmbiguousOpenBoards` marker riding the error chain, a fourth `IpcFailure::Ambiguous` variant, and a structured `ambiguous_open_board` `ToolErrorKind` carrying the path. Deliberately *not* `Rejected`: KiCad declined nothing, and reporting a refusal it never made is the same misreading this PR set out to end. There is a test asserting the message does not say "rejected".

**What counts as an identity.** KiCad's contract is a bare `board_filename` plus `ProjectSpecifier.path` — its own proto documents `board_filename` as "a PCB with a given filename, e.g. `board.kicad_pcb`". So a bare name *with* a project directory is the ordinary case and resolves fine; a bare name *without* one, an empty name, a missing identifier, or a non-board identifier are the shapes that fail closed. Paths that exist are canonicalized; a path that does not exist is normalized lexically (a board deleted under an open editor still compares equal to itself); any other resolution failure — a permission denied on a parent, a symlink loop — fails closed rather than resolving to "different".

Worth flagging: the mocks were sending `project: None` with a bare filename, which is the one form KiCad does not emit. They now send KiCad's shape, which is both more faithful and what made 14 of them meaningful again.

**Tests, including that the write closure never runs.** 11 new tests in `open_document_ambiguity_tests` covering: an unplaceable document; no identifier; an empty filename; one unplaceable document beside a readable one; a duplicated document; the requested board reported twice; a positive match not blocked by another unreadable document; the empty list; the file-only guard reaching the same verdict; ambiguity not reported as a rejection; and a previously-live board staying refused. Each write test asserts the closure was never entered and re-reads the board bytes to prove the file is untouched. Plus 8 unit tests on the identity function itself (including symlink equality, which is what keeps a board KiCad *does* hold from being reported absent).

Process-lifetime stale-board refusal is unchanged and has its own test under ambiguity.

**Negative controls** — each guard neutralized once:

| Neutralized | Tests failing |
|---|---|
| skip the unresolvable records (the original defect) | 6 |
| bare filename resolves as a relative path again | 4 + 2 unit |
| duplicates no longer block a not-open verdict | 1 |
| first match wins whatever else the list holds | 1 |
| ambiguity classified as a rejection | 7 |

**Real-KiCad evidence — one caveat, stated plainly.** I added an `#[ignore]`d test, `real_kicad_open_documents_resolve_to_comparable_paths` in `live_kicad_tools.rs`, that prints every field of every open `DocumentSpecifier` and then asserts the two properties the gate rests on: a live KiCad's list resolves in full, and a board that is genuinely not open is reported as `BoardNotOpen` rather than as an ambiguity. **I have not executed it** — this machine has no display and no Xvfb, and `e2e-kicad.yml` runs kicad-cli only, no GUI session, so it is not covered there either. The shapes the gate refuses were derived from the vendored proto's own contract, not from a live capture. If you would rather that evidence exist before merge, I can either run it against a KiCad on a machine with a display or add a GUI session to the e2e workflow — say which and I will do it.

`Closes #426` is in the PR body. Gate on the rebased head — workspace tests, doctests, `clippy -D warnings`, `fmt --check`, and `crates/schematic-viewer` — all clean. `DEV.md`'s error-kind table has the new row.

pauliuszaleckas:
Follow-up: CI caught a genuine defect in the above that my local gate could not, and #429 landed while this was in flight — both are now fixed and all ten checks are green on `ced0754`.

1. **Windows.** `Path::is_absolute` is what decides whether a project directory can place the bare board filename KiCad sends, and a POSIX-rooted path like `/work/controller` is *not* absolute on Windows — it is relative to the current drive. The rule is right and stays strict; my fixtures were the ones speaking only one dialect. They now pick an absolute path for the platform running them. Production was never affected: KiCad on Windows reports drive-qualified paths.

2. **Rebased onto `9fe3d60`** after #429 merged. Its three `verified_trace_delete_*` tests used the old bare-filename-with-`project: None` mock, so they now use KiCad's document shape like the rest of the file.

Worth saying plainly: I reported the gate clean before pushing, and it was — on Linux. The Windows arm of the matrix is what found this, which is the argument for the three-OS matrix rather than for my local run.

dubesinhower:
@neusse @pauliuszaleckas, I ran the previously unexecuted live document-path test on Windows against the exact PR head `ced0754e475f4d38b17f88827421882496b734a1` (parent `9fe3d60eca9a3d63ccb37b1cf649e83355ff5d0e`). No PR source changes were needed.

Environment: native Windows x86_64, KiCad PCB Editor 10.0.5 (`pcbnew.exe` file version 10.0.5.50609), Rust 1.96.0. The disposable project and board names contain spaces. I created the disposable project through Konnect, opened the board in the real PCB Editor, and saved it through live IPC before testing.

Executed:

```text
cargo test -p konnect --locked --test live_kicad_tools real_kicad_open_documents_resolve_to_comparable_paths -- --ignored --nocapture
```

Observed `GetOpenDocuments` record (only the fixture directory is redacted):

```text
type=3
identifier=Some(BoardFilename("PR407 Evidence.kicad_pcb"))
project=Some(ProjectSpecifier {
    name: "PR407 Evidence",
    path: "<absolute Windows fixture directory>"
})
```

This is live evidence for **bare board filename plus absolute project directory**, not a fabricated absolute filename mock.

Results:

- The exact open board resolved successfully through `find_open_board`.
- The deliberately absent sibling `konnect-not-open-probe.kicad_pcb` classified as typed `IpcFailure::BoardNotOpen`, not ambiguity.
- The same test passed again with forward slashes and a `/./` segment in the requested path, while KiCad retained its native document spelling.
- Both runs exited 0: one passed, zero failed, zero ignored in each run.
- Board SHA-256 was unchanged before/after both read-only runs: `EF1AE5C9BFF611056CA81ABD25C1D9773F866266659D7F5AC41D518814FC1FAD`.
- The tested checkout is clean. All ten hosted checks pass on `ced0754`.

This supplies the Windows live-path observation requested by #426. **Linux and macOS live document-path evidence remains BLOCKED on this host**; this result does not satisfy that cross-platform requirement by inference. It also does not replace the mock/no-write/negative-control coverage for malformed or duplicate records that real KiCad should not emit.

Please retain those remaining platform items in acceptance accounting before closing #426. Our dependent #390 remains held until #394 and corrected #407 are merged.

neusse:
The fail-closed classifier and Windows live evidence on exact head `ced0754e475f4d38b17f88827421882496b734a1` are strong. The complete-set classification, malformed/duplicate identity tests, no-write proofs, process-lifetime stale-board refusal, and ten green checks address most of #426.

Three acceptance items remain:

1. Preserve the observed classification reason in ordinary file-write results. The current fallback warning collapses the result into a generic “transport unreachable or board closed” message, and `get_component_pads` reports only `source:file`. Callers need evidence distinguishing positive `BoardNotOpen` from unreachable IPC and ambiguity.
2. Update `docs/ARCHITECTURE.md` to state that file fallback is allowed for positively established `BoardNotOpen`, not only unreachable IPC, and add the public `ambiguous_open_board` response/error to `docs/API_MIGRATIONS.md`.
3. Update the PR body to the current design, tests, negative controls, exact base/head, Windows live result, and the still-BLOCKED Linux/macOS live evidence.

Linux/macOS live evidence remains an acceptance gap and must not be inferred from hosted unit-test success. If those platforms cannot be obtained immediately, keep #426 open and make the PR explicitly partial rather than closing the issue. Please do not restack dependent #390 yet.

neusse:
Reconstructed on current main and published at exact head \1257025441ac82bde0ea63e5051869181e8d1ea\. The original contributor commit/authorship is preserved. The follow-up commit supplies the remaining classification-specific \allback_reason\ evidence and warning text, updates the architecture/API migration docs, and adds a regression assertion with a proven negative control. All four required local gates pass. Windows live evidence remains applicable; Linux/macOS live-GUI captures are explicitly recorded as validation debt rather than inferred from hosted CI. The PR is mergeable and the new ten-check run is in progress.

PR #416, opened 2026-09-02
fix(placement): stop scoring connector filter caps as decoupling defects

### LINKED ISSUE #411: score_placement's decoupling check flags a correctly-placed connector/interface cap as too far from the nearest IC
## Summary

`score_placement`'s decoupling heuristic measures distance from every cap on a shared net to the *nearest IC* on that net, and penalizes anything past a fixed 2.5 mm threshold. It has no way to recognize a cap that is deliberately placed on a connector's own pins for cable/EMI filtering rather than for decoupling a nearby chip — and dings it as a placement defect.

## Environment

- Konnect 0.10.0

## Repro

Board has `U1` (an MCU module) and `J2` (a 4-pin cable connector carrying VCC/GND/RX/TX to an off-board sensor over ~10 m of cable). `C1` (100 nF) was placed directly on J2's VCC/GND pins as cable-entry filtering, per standard practice for a long unshielded run:

```
get_component_pads(board="...", reference="C1")
→ {"pads":[{"net":"VCC","number":"1","x":127.0,"y":65.45},
           {"net":"GND","number":"2","x":127.0,"y":63.55}]}

get_component_pads(board="...", reference="J2")
→ pad "1" net "VCC" at (123.7, 65.475)   # ~3.7 mm from C1 — the cap's actual target
```

```
score_placement(board="...")
→ {"deductions":[{"detail":"C1 (100nF, limit 2.5 mm) is 3.8 mm from nearest shared-net IC U1",
                   "kind":"decoupling","points":15,"references":["C1"]}], ...}
```

## Expected

A cap sitting directly on (or within a pad-width or two of) a connector's own net pins, on the connector's own net, is recognized as intentional interface/EMI filtering and excluded from the "too far from nearest IC" check — or the check at minimum considers the *nearest component on the net* generally, not only ICs, before penalizing.

## Actual

The deduction fires based solely on distance-to-nearest-IC, with no check for whether a closer, non-IC target (the connector itself) explains the placement.

## Impact

Produced a false-positive design defect in the placement score for a cap that was correctly placed exactly where it needed to be for its actual purpose, potentially misleading anyone using the score as a go/no-go signal.

## Affected versions

tonydzi:
Closes #411

**1. Problem and scope.**

`score_placement`'s decoupling heuristic measures every family-classified capacitor to the nearest **IC** (`U*`) it shares a net with, and deducts 15 points if that distance exceeds the value family's limit. It has no notion of a capacitor deliberately placed **on a connector's own pins** for cable/EMI filtering. On the board in #411 — an MCU `U1` and a 4-pin cable header `J2` — a 100 nF cap sitting on J2's VCC/GND pins is reported as a placement defect naming an IC it was never meant to bypass:

```
{"kind":"decoupling","points":15,"references":["C1"],
 "detail":"C1 (100nF, limit 2.5 mm) is 3.8 mm from nearest shared-net IC U1"}
```

Scope is `score_placement`'s decoupling check only. Weights, verdict logic, hard failures, the connector-edge check, and `place_decoupling_caps` are untouched.

**2. Root cause and chosen design.**

The check asked one question — "how far is this cap from its IC?" — and treated the answer as the whole story. Before deducting, it now asks what the cap is actually sitting on. A capacitor is interface/EMI filtering for a cable, and is exempted, when **all three** hold:

1. it is within its **own family limit** of the **courtyard bbox** of a connector (`J*`) — the same limit and the same courtyard-edge metric the IC rule uses, so the exemption can never be looser than the rule it waives;
2. that connector carries **every** named net the cap carries — a two-terminal filter cap belongs to a cable only if both its terminals do. One shared net is not enough: board-wide `GND` alone would let any cap that drifted near any connector claim the waiver;
3. the cap is on the connector's **face**, unless the connector's pads carrying those nets reach **both** copper faces — a plated through-hole pin, not an SMD one.

Condition 3 exists because a blanket same-side rule would reintroduce the exact false positive this issue is about: @ncolomer's board has a `B.Cu` cap filtering an `F.Cu` through-hole header, and that cap is electrically on those pins. Reading a pad's copper stack needs one additive field on `PadSite` in `konnect-sexp` (`layer_side` records where a *footprint* sits and cannot say whether its copper reaches the far face), so this PR is two commits: the sexp field, then the placement change.

The waiver is **reported rather than silent**: exempted caps appear in a new `interface_filter_caps` array (`reference`, `value`, `connector`, `connector_distance_mm`, `limit_mm`), because a deduction that vanishes without a trace is indistinguishable from a broken checker — the same reason this tool already reports `outline_missing` and passing `connector_edges` explicitly.

Alternatives rejected: exempting whenever a connector is merely *nearer* than the closest IC (unbounded — a cap abandoned in a corner 20 mm from everything qualifies for being 19 mm from a connector); a separate, larger connector threshold (a second magic number with nothing behind its value); footprint/value heuristics (this module identifies by net pairing, never by reference guessing).

Implementation reuses existing helpers rather than duplicating them: the IC and connector lookups share `nearest_shared_net`, the point-to-courtyard-bbox formula moved verbatim into `distance_to_bbox`, and `footprints_with_prefix` wraps the existing `ref_prefix` idiom.

**3. Documentation.**

The new response field is documented as public API, per review:

- `tool-directory.md` — `score_placement`'s row names `interface_filter_caps` and says it is evidence, not a separate score.
- `crates/konnect/assets/skills/kicad-pcb/SKILL.md` — an agent reading a score now knows a filter cap is judged against its connector rather than the nearest IC, and that entries in the array are answered checks, not defects to "fix" by dragging a cap toward an IC.
- `docs/API_MIGRATIONS.md` — additive field, minor-release impact, and the exact predicate (every named net; courtyard-edge distance inside the cap's own family limit; same face unless the connector's pads carrying those nets reach both faces).

`interface_filter_caps` is also registered in the `NOT_TOOLS` list of `crates/konnect/tests/asset_references.rs`: the shipped-docs guard reads a backticked identifier in prose as a tool name, and this one is a response field.

**4. Compatibility and migration.**

Additive. No existing response field changes shape or meaning. Scores only ever move **up**, and only for caps that satisfy all three conditions — a board with no such cap scores exactly as before (the repository's own fixture still scores 70 with the same two caps named).

`uncoupled_caps` is intentionally unchanged: a cap sharing no net with any IC never reaches the distance rule, so "nothing to decouple" remains the honest report for it rather than being relabelled a waived deduction. That decision is pinned by its own test.

Known limits, stated rather than hidden: only the `J*` prefix counts as a connector (consistent with the existing connector-edge check, so a board naming connectors `CN1` is unaffected); `place_decoupling_caps` will still plan a move for an exempted cap if a user explicitly asks it to; and a two-terminal cap whose nets are, say, `VCC`+`GND` is exempted by a connector carrying both, which for a power header is the intended reading.

**5. Tests run.**

Nine tests in `tools::placement::tests` cover this change. All are derived from the existing KiCad-authored `placement_fixture.kicad_pcb` by the string surgery this module already uses, with every distance hand-computed from the fixture's own coordinates in the doc comments:

- `a_cap_on_a_shared_net_connector_is_interface_filtering_not_a_decoupling_defect` — the #411 repro: C1 moved onto J1's pins (1.61 mm from J1's courtyard, 5.3 mm from U1). Was 70 with C1 named; now 85 with only C2 named and C1 reported as an interface filter.
- `a_cap_far_from_both_its_ic_and_the_connector_still_deducts` — 3.61 mm from J1, 3.3 mm from U1: still a defect, score stays 70. The exemption is not an off-switch.
- `a_cap_beside_a_connector_it_shares_no_net_with_still_deducts` — same 1.61 mm geometry, J1's pads retagged so the two are net-disjoint: still deducts. Proximity alone is not the exemption.
- `a_connector_cap_sharing_no_net_with_any_ic_is_still_reported_uncoupled` — pins the `uncoupled_caps` decision.
- `a_cap_within_both_limits_is_neither_a_violation_nor_a_reported_waiver` — pins that `interface_filter_caps` means "a deduction was waived", not "a cap is near a connector".
- `the_exemption_uses_the_caps_own_family_limit_not_the_100nf_one` — C1 retagged to `1uF` (5.0 mm family) at 3.609 mm from J1 and 5.778 mm from U1, so the exemption fires at a distance the 100 nF family would reject.
- `a_cap_on_the_far_face_of_a_through_hole_connector_is_still_interface_filtering` — @ncolomer's topology, and the guard against a blanket same-side rule.
- `a_cap_on_the_far_face_of_a_surface_mount_connector_is_not_interface_filtering` — the same board with J1's pads retagged `*.Cu` → `F.Cu` deducts again (85 → 70), so the through-hole carve-out is not a hole.
- `a_cap_sharing_only_ground_with_the_connector_still_deducts` — 85 → 70, pinning the every-net rule.

Each was shown **failing first** on the commit before it. Eleven mutations of the first round were caught one at a time (threshold widened ×3; comparison flipped; distance gate deleted; connector limit hard-coded to `2.5`; connector limit zeroed; prefix ignored; `J`→`U`; shared-net requirement dropped on the connector path; shared-net requirement dropped globally; exemption ordered before the IC-distance check; full revert), and a six-way grid on the tightening round kills every way of loosening either new condition.

Measured on the exact head of this branch, macOS, today: `cargo test --workspace --locked --lib --tests` → **29 suites, 1492 passed, 0 failed, 20 pre-existing ignored**. The docs guard `-p konnect --test asset_references` is 12 passed (it fails without the `NOT_TOOLS` entry above, which is how that entry was found rather than guessed).

Not covered, stated plainly rather than left for you to find: the 10uF family (only 100 nF and 1uF are exercised); `decoupling_limit_mm`'s own string parsing (`µ`/`u`, trailing `F`, `1000n`); and boards with more than one connector, so "nearest shared-net connector" is only ever a one-element minimum.

**6. Risk and rollback.**

Low. No file format, IPC, packaging or release surface is touched. The only behavioural risk is a false exemption — a genuinely misplaced cap that happens to sit within its family limit of a connector carrying all of its nets, on a face its copper reaches. That is the definition of being on its pins. Rollback is reverting these commits: no state, no persisted data, no migration.

@ncolomer's measurement on the real board is in #411: C1's centre is **1.1 mm** from J2's courtyard edge (the issue's 3.7 mm was to pad 1's centre), both of C1's nets are on J2, and J2 is a through-hole JST whose pads report `*.Cu`. All three conditions hold, so the exemption fires on the board that reported this.

Authored by Mycroft, the synthetic co-founder at Anton Dzyatkovsky's lab (autonomous mode; named responsible person: Anton Dziatkovskii). The test runs above were re-executed on this head before submission.

### DIFF
--- a/crates/konnect-core/src/tools/placement.rs
+++ b/crates/konnect-core/src/tools/placement.rs
@@ -1416,14 +1416,18 @@ struct InterfaceFilterCap {
 /// violation — it is reported separately as uncoupled, because "nothing to
 /// decouple" and "too far from what it decouples" are different findings.
 ///
-/// A cap sitting on the pins of a connector (J*) it shares a net with is
+/// A cap sitting on the pins of a connector (J*) whose nets it belongs to is
 /// interface/EMI filtering for that cable, not a bypass cap that drifted away
 /// from an IC on the same rail, so it is exempted from the distance rule
 /// (#411). The exemption is deliberately narrow — the cap must be within the
 /// *same* family limit of the connector's courtyard that the rule would demand
-/// of an IC, so a cap far from everything still deducts. Exemptions are
-/// returned as the third list rather than dropped, so the response says the
-/// deduction was waived instead of going quiet about a check that ran.
+/// of an IC, the connector must carry *every* named net the cap has, and where
+/// the two sit on opposite faces the connector's copper must actually reach
+/// the cap's face (see `nearest_interface_connector`). So a cap far from
+/// everything still deducts, and so does one that merely shares ground with a
+/// connector it happens to sit beside. Exemptions are returned as the third
+/// list rather than dropped, so the response says the deduction was waived
+/// instead of going quiet about a check that ran.
 ///
 /// `uncoupled_caps` is intentionally untouched by the exemption: a cap that
 /// shares no net with any IC never reached the distance rule in the first
@@ -1440,14 +1444,29 @@ fn decoupling_check(
     // Net set per reference, from the connectivity index — keyed by name, so
     // both file formats resolve identically.
     let mut nets_by_ref: HashMap<&str, BTreeSet<&str>> = HashMap::new();
+    // The subset of those nets a part carries on copper that reaches both
+    // faces of the board — what decides whether a part on the far face is
+    // electrically on these pins or only above them.
+    let mut through_nets_by_ref: HashMap<&str, BTreeSet<&str>> = HashMap::new();
     for net in index.nets() {
         for pad in index.pads_of_net(net) {
             nets_by_ref
                 .entry(pad.reference.as_str())
                 .or_default()
                 .insert(net);
+            if pad.reaches_both_faces {
+                through_nets_by_ref
+                    .entry(pad.reference.as_str())
+                    .or_default()
+                    .insert(net);
+            }
         }
     }
+
+    let side_by_ref: HashMap<&str, Side> = items
+        .iter()
+        .filter_map(|c| Some((c.reference.as_deref()?, c.layer_side)))
+        .collect();
 
     let ics = footprints_with_prefix(items, "U");
     let connectors = footprints_with_prefix(items, "J");
@@ -1481,8 +1500,16 @@ fn decoupling_check(
                 // shared-net connector's own pins is filtering that cable
                 // (#411). Same limit, measured against the connector, so this
                 // cannot swallow a cap that is far from everything.
-                match nearest_shared_net(&connectors, cap_nets, &nets_by_ref, center)
-                    .filter(|(_, connector_distance_mm)| *connector_distance_mm <= limit_mm)
+                match nearest_interface_connector(
+                    &connectors,
+                    cap_nets,
+                    side_by_ref.get(reference).copied(),
+                    &nets_by_ref,
+                    &through_nets_by_ref,
+                    &side_by_ref,
+                    center,
+                )
+                .filter(|(_, connector_distance_mm)| *connector_distance_mm <= limit_mm)
                 {
                     Some((connector, connector_distance_mm)) => {
                         interface_filters.push(InterfaceFilterCap {
@@ -1540,6 +1567,50 @@ fn nearest_shared_net<'a, 'n>(
                 (Some(a), Some(b)) => !a.is_disjoint(b),
                 _ => false,
             }
+        })
+        .map(|(reference, bbox)| (*reference, distance_to_bbox(center, *bbox)))
+        .min_by(|(_, a), (_, b)| a.total_cmp(b))
+}
+
+/// The nearest connector this capacitor is filtering for, and how far its
+/// courtyard bbox is from the cap's center. Same metric as the IC rule, two
+/// stricter pairing conditions, both from review of #416:
+///
+/// * The connector must carry **every** named net the cap has, not merely one.
+///   A two-terminal filter cap belongs to a cable only if both its terminals
+///   do; accepting a single overlap would let any cap that drifted near any
+///   connector claim the exemption on board-wide ground alone.
+/// * A cap on the opposite face qualifies only where the connector's copper is
+///   actually on the cap's face: the pads carrying those nets must reach both
+///   faces. Board side alone cannot answer this. A plated through-hole pin is
+///   as available from below as from above — the case @ncolomer measured on the
+///   #411 board, where a B.Cu cap filters an F.Cu THT header — while an SMD
+///   header's copper is not, however close the two look in XY.
+fn nearest_interface_connector<'a, 'n>(
+    connectors: &[(&'a str, Bbox)],
+    cap_nets: Option<&BTreeSet<&'n str>>,
+    cap_side: Option<Side>,
+    nets_by_ref: &HashMap<&str, BTreeSet<&'n str>>,
+    through_nets_by_ref: &HashMap<&str, BTreeSet<&'n str>>,
+    side_by_ref: &HashMap<&str, Side>,
+    center: (f64, f64),
+) -> Option<(&'a str, f64)> {
+    let cap_nets = cap_nets?;
+    connectors
+        .iter()
+        .filter(|(reference, _)| {
+            let carries_every_net = nets_by_ref
+                .get(*reference)
+                .is_some_and(|connector_nets| cap_nets.is_subset(connector_nets));
+            if !carries_every_net {
+                return false;
+            }
+            if side_by_ref.get(*reference).copied() == cap_side {
+                return true;
+            }
+            through_nets_by_ref
+                .get(*reference)
+                .is_some_and(|through_nets| cap_nets.is_subset(through_nets))
         })
         .map(|(reference, bbox)| (*reference, distance_to_bbox(center, *bbox)))
         .min_by(|(_, a), (_, b)| a.total_cmp(b))
@@ -1800,16 +1871,10 @@ mod tests {
         fixture.replace(from, to)
     }
 
-    /// Rewrite text inside ONE footprint's block, leaving every other footprint
-    /// alone. Line-scoped like `lock_footprint`: find the block enclosing the
-    /// Reference property, stop at that footprint's own closing paren, so a
-    /// string several parts share is only rewritten here.
-    fn rewrite_inside_footprint(
-        content: &str,
-        reference: &str,
-        rewrites: &[(String, String)],
-    ) -> String {
-        let lines: Vec<&str> = content.lines().collect();
+    /// The line range of ONE footprint's block: from its `(footprint …)` header
+    /// down to that footprint's own closing paren. Shared by the rewriters so
+    /// they cannot disagree about where a footprint ends.
+    fn footprint_block_range(lines: &[&str], reference: &str) -> (usize, usize) {
         let marker = format!(r#"(property "Reference" "{reference}""#);
         let r = lines
             .iter()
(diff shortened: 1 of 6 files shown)


neusse:
This is focused and all ten checks are green, but it is not yet ready for the merge queue.

Please make the issue accounting explicit by adding Closes #411, then obtain the reporter's fixture/confirmation requested in the PR body. The key measurement must be the connector courtyard-edge distance used by the implementation, not only pad-center distance from the report.

Also keep the release impact explicit: interface_filter_caps is a new public response field, so this belongs in minor-version planning even though the scoring correction itself is narrow. After the reporter evidence is attached, update from current main, rerun the exact-head checks, and record the final base/head. We can then review the one-file diff promptly.

tonydzi:
Both safety decisions are implemented, the issue accounting is explicit, and the branch is on current `main`. Taking your four points in order.

**Issue accounting.** `Closes #411` is now the first line of the PR body.

**Reporter evidence - and the number you asked for.** @ncolomer measured the board against the predicate rather than the build, which is the part that mattered: the metric this check actually uses is courtyard-edge, not pad-center. C1's center is at (127.0, 64.5); J2's courtyard in board space is X:[120.4, 125.9], Y:[63.025, 73.925]. C1's Y falls inside that span, so the gap is pure horizontal - **1.1 mm**, against the issue's 3.7 mm pad-center figure and a 2.5 mm limit. Both of C1's nets (VCC, GND) are on J2's pins, so it satisfies the stricter predicate too, not just the old one.

**Every named net, not one.** Implemented as you described: the connector must carry every named net the capacitor has. `a_cap_sharing_only_ground_with_the_connector_still_deducts` pins it - J1's VCC pad retagged to `/SIG_IN`, geometry untouched at 1.61 mm, so only the net rule can decide. It was green at 85 before the change and requires 70 now.

**Board side - and why a blanket same-side rule would have been wrong.** This is @ncolomer's finding, not mine, and it changed the design. That board is a live instance of exactly the risk you flagged: C1 on B.Cu, J2 on F.Cu. But J2 is a through-hole header whose pads are `(layers "*.Cu" "*.Mask")` - its copper is on both faces, so a cap underneath is electrically on those pins. An unconditional same-side requirement would have turned that board from correctly-exempted back into a false positive, reintroducing #411 for a very common topology.

So the rule is: same face, or the connector's pads carrying the cap's nets reach both faces. `layer_side` cannot answer that - it records where a footprint sits, not where its copper is - so `PadSite` gained a `reaches_both_faces` field read from the pad's own `(layers …)` stack. That is what makes the diff two files instead of one; it is an additive field in its own commit, so the placement change still reads on its own. Two tests, identical boards except J1's pads: `*.Cu` keeps the exemption, `F.Cu` (an SMD header) loses it and C1 deducts again at 5.3 mm.

**The two negative tests were red first.** On the previous commit with the tests already in place, both reported `left: Number(85), right: 70` with C1 still in `interface_filter_caps`. The through-hole test passes before and after by construction - it exists to keep the carve-out from being deleted, and the mutation grid is what proves it does that.

Six mutations, each applied alone, each caught:

| mutation | test that went red |
| --- | --- |
| every-net → any-shared-net | `..._sharing_only_ground_...` |
| side check deleted | `..._surface_mount_connector_is_not_interface_filtering` |
| through-hole carve-out deleted | `..._through_hole_connector_is_still_interface_filtering` |
| every pad claims to reach both faces | `..._surface_mount_...` |
| no pad reaches both faces | `..._through_hole_...` |
| `*.Cu` wildcard ignored | `..._through_hole_...` |

**Release impact, stated as you asked.** `interface_filter_caps` is a new public response field on `score_placement`, so this belongs in minor-version planning regardless of how narrow the scoring correction is. Nothing existing changes shape or meaning and scores only ever move up. The offer from the original PR body still stands: if you would rather keep the response surface unchanged, the array can go and the fix does not depend on it.

**Base and head, on the exact tree that was checked.** Rebased onto `main` at `9fe3d60`; head `efc63e2`.

Full checklist rerun on that head: `cargo test --workspace --locked --lib --tests` 29 suites, 1492 passed, 0 failed, 20 pre-existing ignored; `cargo test --workspace --locked --doc` 8 passed, 0 failed; `cargo clippy --workspace --locked --all-targets -- -D warnings` clean; `cargo fmt --all -- --check` clean. All 10 GitHub checks green on `efc63e2`. The totals moved from last round's 1399/18 because the base did, not because anything here changed them.

Known limits unchanged from the first round, plus one new one worth naming: the through-hole allowance keys on the pads that carry the capacitor's nets, so a connector whose signal pins are plated but whose mounting pads are not is judged on the signal pins - which is the intent, but it does mean the field is per-pad and not a property of the connector as a whole.

neusse:
The implementation and evidence are now convincing on exact head `efc63e26d0d8177d148e11ca99424b5b68fe5359`: the every-named-net rule, same-face/through-hole exception, reporter's 1.1 mm courtyard-edge measurement, negative controls, issue closure path, and ten green checks are all present.

The remaining blocker is documentation of the new public response surface. Please update:

- `tool-directory.md` for `score_placement.interface_filter_caps`, including what qualifies and that this is evidence rather than a separate score;
- the bundled `crates/konnect/assets/skills/kicad-pcb/SKILL.md` guidance so an agent knows how to interpret that field;
- `docs/API_MIGRATIONS.md` with the additive field and minor-version impact.

Please also make the final PR body internally consistent with the implemented every-net and through-hole behavior; remove any earlier wording that still describes a one-shared-net or ground-only exemption. No further algorithmic change is requested. After those documentation edits and green exact-head checks, this should be ready for final review.

tonydzi:
Docs are in, and the PR body now matches the implemented behaviour. Head is `7b4d0f62b8bc319dd82e867083e78b5b54e408c7`, base `9fe3d6098` (current `main`, 0 behind).

**Documentation, as asked:**

- `tool-directory.md` — `score_placement`'s row names `interface_filter_caps`, says what qualifies, and says it is evidence rather than a separate score.
- `crates/konnect/assets/skills/kicad-pcb/SKILL.md` — two places: the placement-strategy list now distinguishes cable/EMI filter caps from decoupling caps, and the `score_placement` entry tells an agent that entries in `interface_filter_caps` are answered checks, not defects to "fix" by dragging a cap toward an IC.
- `docs/API_MIGRATIONS.md` — a new **Unreleased (minor release)** section with the additive field, its five keys, and the exact predicate: every named net, courtyard-edge distance inside the cap's own family limit, same face unless the connector's pads carrying those nets reach both copper faces.

**PR body:** rewritten. The one-shared-net wording is gone from §2 (replaced by the three explicit conditions), and two stale "known limits" went with it — the old body still said a cap sharing only GND would be exempted and that the check ignores `layer_side`, both of which the tightening commit made false. §5 now lists the nine tests by their real names, including the two negative ones from the tightening round.

**One thing the docs edit broke, and how it was found.** `tool-directory.md` and the SKILL are both shipped assets, and `crates/konnect/tests/asset_references.rs` reads any backticked identifier in them as a tool name:

```
shipped docs instruct tools that do not exist:
  skills/kicad-pcb/SKILL.md:149: `interface_filter_caps` reads like a tool but is in no toolset
```

So documenting a new response field requires registering it as not-a-tool. Added `interface_filter_caps` to `NOT_TOOLS` with the same one-line justification the neighbouring `files_generated` and `sheet_instance_path` entries carry. That guard is exactly the check the test was written for, so it is worth noting it did its job here rather than quietly passing.

**Exact-head checks, run on this branch today (macOS):**

```
cargo test --workspace --locked --lib --tests   29 suites, 1492 passed, 0 failed, 20 pre-existing ignored
cargo test -p konnect --test asset_references    12 passed   (1 failed before the NOT_TOOLS entry)
cargo test -p konnect-core placement             69 passed
```

No algorithmic change in this round — the only non-documentation line is the `NOT_TOOLS` entry that the documentation itself requires. CI has the push; happy to squash the docs commit into the placement commit before merge if you would rather the branch land as two commits instead of three.

neusse (approved):
Reviewed exact head 7b4d0f62b8bc319dd82e867083e78b5b54e408c7 against #411 and current main. Both specification and repository-standards reviews pass. The previously requested public API, tool-directory, bundled-skill, migration, and PR-body documentation is complete. The every-net, courtyard-edge, same-face/THT exception, reporter evidence, negative controls, and closing linkage are present. All ten required checks are green on this head. Ready for the maintainer merge gate.

neusse (approved):
Revalidated updated exact head 1c224c16f8acc557241d57e6a97dcbbac16cd59a. The only head change is the conflict-free merge of upstream main at 5a59f3cf52e2661a75ce3db0f2139a87ab3f23c9; its tree exactly matches the previously reviewed #416 content combined with current main. The focused six-file diff and Closes #411 linkage are unchanged. All ten required checks passed on this exact head. Ready for the maintainer merge gate.

PR #421, opened 2026-09-04
feat(config): report which configuration file configured the running process

### LINKED ISSUE #419: Config search order is first-match-wins, documented as a plain list, and nothing reports which file was used
# Config search order is first-match-wins, documented as a plain list, and nothing reports which file was used

## Summary

`Config::load` stops at the **first** config file that exists. The documentation
presents the five locations as an ordered list without saying that, which reads
as layered precedence per key. In the standard plugin layout the
`settings.json` beside the plugin is present, so the platform config directory is
**never read** — silently,
with no warning, and with no way to ask the server which file it actually used.

I lost about an hour to this: a key was moved to the platform config file, the
server was restarted, and nothing changed. There was no signal distinguishing
"the setting is wrong" from "that file is not being read at all".

## The code

`crates/konnect/src/config.rs`:

```rust
let mut config = None;
for path in &config_paths {
    if path.exists() {
        config = Some(Self::load_from(path)?);
        break;                       // <-- first existing file wins; no merge
    }
}
```

`load_from` deserializes a whole `Config` from that one file, so every key not
present in the winning file falls back to its serde default — it does not fall
through to the next location.

## What the docs say

`docs/KICAD_INTEGRATION.md`, "Configuration":

> `crates/konnect/src/config.rs` searches, in order:
> 1. `konnect.toml` in the working directory;
> 2. `settings.json` in the working directory;
> 3. `settings.json` beside the executable;
> 4. `settings.json` one directory above the executable;
> 5. the platform configuration directory.

Nothing there says only the first match is read. "Searches, in order" is the
same phrasing used by tools that merge with precedence, so the natural reading
is the wrong one.

## Consequence in the standard plugin layout

Slot 4 is `settings.json` one directory above the executable. That is the
plugin layout — `<plugin_dir>/bin/konnect` with `<plugin_dir>/settings.json`
beside it — which `exe_relative_settings_paths` exists specifically to cover; its
own comment says so. Whenever the server is installed that way and a
`settings.json` is present, slot 5 — `~/Library/Application Support/konnect/config.toml`
on macOS — can never be reached.

On this machine that file exists and contains:

```toml
kicad_cli = "/Applications/KiCad/KiCad.app/Contents/MacOS/kicad-cli"
kicad_binary = "/Applications/KiCad/KiCad.app/Contents/MacOS/kicad"
ipc_address = "ipc:///tmp/kicad/api.sock"
```

Every one of those is dead. It happens to be harmless because slot 4 sets the
same values, but nothing would have said otherwise if it did not.

This also means **a key cannot be moved out of the plugin directory to survive a
plugin update** — the documented "platform configuration directory" is
unreachable while the plugin file exists. `--config <path>` is the only route,
and it has to carry the entire config because only one file is ever read.

Related hazard from the same list: slots 1 and 2 are relative to the **working
directory**. A `settings.json` or `konnect.toml` that happens to sit in a project
root would silently take over the whole server config.

## Nothing reports the choice

`server_stats` is the natural place to look and does not carry it:

```json
{"started_at_ms":…,"uptime_ms":…,"total_calls":13,"error_calls":0,
 "log_path":"~/Library/Application Support/konnect/logs/calls.jsonl",
 "per_tool":{…}}
```

Startup logs nothing either. The one config path that *does* appear in the log
comes from the `load_user_config` **tool**, and it names a different file for a
different purpose:

```
INFO konnect_core::tools::config: [BETA] Loading user config path=~/Library/Application Support/konnect/config.json
```

So a user grepping the log for "config" is pointed at `config.json` — which is
not the file that configures the server, and is not even in the search list
(slot 5 is `config.toml`). That actively misleads.

## Suggested fix

1. **Say it in the docs**: the first existing file wins; the search does not
   merge, and later locations are unreachable once an earlier one exists.
2. **Report the resolved path.** Either a field on `server_stats`
   (`config_path`, plus perhaps `config_candidates_skipped`), or an INFO line at
   startup naming the file that won. The server knows this and currently exposes
   it nowhere; one field turns an hour of bisecting into one call.
3. Optional: warn when a later-slot config file exists but was skipped. That is
   exactly the silent case above.

Point 2 matches the evidence rule in `GOVERNANCE.md` — a response field derived
from the result rather than echoed from the request. Here there is no field at
all for a fact the server definitely knows.

## Environment

- Konnect 0.11.0, **built from source** and installed into the KiCad plugin
  directory (`<plugin_dir>/bin/konnect`), macOS, KiCad 10.0.5
- Config in use: `settings.json` one directory above the executable (slot 4)

Happy to open a PR for the `server_stats` field and the doc correction if the
shape above is the one you want.

triglav-modular:
## Summary

Konnect loads the **first existing** configuration file from its search list and ignores the rest, but nothing reported which one won. A user whose settings appear to have no effect cannot distinguish a shadowed file from a malformed one, and in the standard plugin layout a `settings.json` beside the binary makes the platform config directory permanently unreachable with no indication at all.

`get_installation_info` now reports the configuration that actually started the process, and the docs state the first-match-no-merge policy outright.

Issue: #419

Closes #419

## Approach

An immutable `ConfigResolution` is captured once during startup and reported read-only as an additive `configuration` block:

```json
"configuration": {
  "source": "explicit_path | search_path | defaults | unavailable",
  "selected_path": "/absolute/path/to/settings.json",
  "search_policy": "first_existing_no_merge",
  "skipped_existing_paths": ["/absolute/path/to/later/config.toml"]
}
```

Captured at startup rather than recomputed on call: re-running the search when the tool is invoked would describe the filesystem as it is *then*, and could name a higher-priority file created after launch — the opposite of the question being asked. A protocol test pins that.

Candidate selection is extracted into `select_config_candidate(&[PathBuf])` so precedence is testable against a supplied list; the real list depends on the working directory and `current_exe()`, neither of which a test can change without affecting the whole process.

Startup also emits one INFO record naming the selection, for when no tool call is possible yet — not one warning per shadowed file, since first-match is the intended policy.

**Three choices that differ from the outline in the issue, all easy to change:**

1. **`ConfigResolution` lives in `konnect-core`, not beside `Config` in the `konnect` crate.** `get_installation_info` is serialized in core and the dependency runs `konnect → konnect-core` only, so a record defined next to `Config` is invisible to the serializer. The alternative — passing a pre-built JSON value out of the binary crate — would put this response's shape somewhere no other field of `get_installation_info` lives.
2. **The constructor is `McpHandler::new_with_config_resolution`**, where the issue suggested `new_with_runtime_provenance`. Named after the type so one noun spans the JSON key, the type, the field and the constructor. Happy to rename if the other reads better.
3. **`search_policy` is a module constant, not a field on the record.** It describes the build, not the individual resolution. The public JSON is identical either way.

**The `ffi.rs` caveat is deliberately not fixed here.** `kicad_plugin_init` uses `unwrap_or_default()` on both config loads, so a malformed explicit `--config` silently starts on defaults. That is a separate defect and is left for its own issue to keep this PR to one reviewable outcome. What this PR does guarantee is that provenance never *lies* about it: that path reports the real resolution when the load succeeds and `unavailable` when it fails, never a fabricated clean `defaults`. (While implementing, I also noticed `ffi.rs` never calls `apply_env_fallbacks`, so its `--config` branch skips it while `main.rs` does it for the same case — relevant on that path specifically, since it is the KiCad plugin route and #39 exists because KiCad launches with `--config` and `KICAD_API_SOCKET` together. Behaviour unchanged here; I will file both symptoms together.)

`Config::load()` is removed rather than kept as a wrapper: `mod config` is private in both `main.rs` and `lib.rs`, so it was unreachable outside the crate, and both call sites now take the resolution.

## Branch and dependencies

- Base branch: `main` (branched from `a51b856`)
- Depends on: nothing
- Series order: standalone, not part of a series
- Unique commits/acceptance criteria owned: one commit; the whole of #419's provenance and documentation outcome

## Compatibility and safety

**This is an additive public response change, not a no-op for compatibility.** `get_installation_info` gains a `configuration` object; no existing field is renamed, removed or changed. Per GOVERNANCE.md's version rule a changed response shape is a **minor**, and the release notes should mention the new block.

Behaviour is otherwise unchanged, and this is asserted rather than assumed — the first existing candidate still wins, a malformed selected file is still a startup error rather than permission to fall through, no candidates still means defaults, `--config` still bypasses discovery, and the environment fallback (#39) still applies after file or default resolution.

No file or IPC mutations. The block carries **paths only** — never configuration values, file contents or IPC credentials — verified by a test that puts a token in `ipc_address` and asserts it does not appear.

No dependencies added; `Cargo.lock`, `Cargo.toml`, `registry.rs` and `tool-directory.md` are untouched, and no tools were added or removed, so `fix-doc-counts` does not apply.

## Validation

```
cargo fmt --all -- --check                                     → 0
cargo test --workspace --locked --lib --tests                  → 0  (29 suites)
cargo test --workspace --locked --doc                          → 0
cargo clippy --workspace --locked --all-targets -- -D warnings  → 0
```

Acceptance matrix from the issue, all covered: no files exist → `defaults`/null/empty; first missing, second exists → the second; first and third exist → first selected, third reported skipped; malformed selected file → startup error, no fall-through; valid explicit `--config` → `explicit_path` with the automatic list not consulted; malformed explicit `--config` → startup error, no fall-back; blank IPC plus environment socket → same source and path, env fallback still applied.

Protocol-level tests spawn the real binary and prove the returned absolute path is the file that controlled startup, that a shadowed later candidate is named, and that creating a higher-priority file *after* launch does not change the answer.

**Per the evidence rule, every new guard was neutered and watched to fail.** Discarding the resolution in the handler fails all provenance tests (and they report `unavailable`, not a fabricated `defaults`). Making the malformed-file loads `unwrap_or_default()` fails both startup-failure tests. All pass again on restore.

**Existing fields verified unchanged**, as requested: I built a binary from `a51b856`, ran both through the same scenario and diffed the whole `get_installation_info` response. Keys added: exactly `configuration` and its four children. Keys removed: none. The only differing shared values were `build.commit` and `runtime.executable_path`, which differ because they are two different binaries.

Also exercised by hand against the real binary, including the case that motivates the issue: with `<plugin>/bin/konnect` and `<plugin>/settings.json`, the response names that `settings.json` as selected and reports the platform `config.toml` as shadowed.

**Not run, and why:** the Windows `APPDATA` branch of `dirs_config_path` (developed and tested on macOS; it is exercised only on Windows CI). `source: "unavailable"` is not reachable through the binary — it belongs to the cdylib entry point, so it is covered by unit tests and by the neutering run rather than end-to-end. No real-KiCad, viewer or packaging checks were relevant to this change.

### DIFF
--- a/crates/konnect-core/src/config_resolution.rs
+++ b/crates/konnect-core/src/config_resolution.rs
@@ -1,0 +1,190 @@
+//! Which configuration file configured the running process.
+//!
+//! Konnect selects the **first existing** server configuration file from its
+//! search list and loads only that one; later files are not merged in. Nothing
+//! in the protocol reported which file won, so a user whose settings appeared to
+//! be ignored had no way to tell a shadowed file from a malformed one (#419).
+//!
+//! This type records that decision once, at startup, and
+//! `get_installation_info` reports it. It is deliberately captured rather than
+//! recomputed on demand: re-running the search when the tool is called would
+//! describe the filesystem as it is *now*, which may name a higher-priority file
+//! created after launch — the opposite of the question being asked.
+//!
+//! The record carries paths only. Configuration values, file contents and IPC
+//! credentials are never included.
+
+use std::path::{Path, PathBuf};
+
+/// How the running process obtained its configuration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConfigSource {
+    /// An explicit `--config <path>` was given; the search list was not consulted.
+    ExplicitPath,
+    /// A file from the automatic search list was selected.
+    SearchPath,
+    /// No candidate existed, so built-in defaults are in use.
+    Defaults,
+    /// The resolution was not recorded by this entry point. Reported rather than
+    /// guessed, so a caller never reads a fabricated `defaults` for a load whose
+    /// outcome is genuinely unknown.
+    Unavailable,
+}
+
+impl ConfigSource {
+    /// Stable public string, as reported in `get_installation_info`.
+    pub fn as_str(self) -> &'static str {
+        match self {
+            ConfigSource::ExplicitPath => "explicit_path",
+            ConfigSource::SearchPath => "search_path",
+            ConfigSource::Defaults => "defaults",
+            ConfigSource::Unavailable => "unavailable",
+        }
+    }
+}
+
+/// The startup configuration decision, captured once and reported read-only.
+#[derive(Debug, Clone)]
+pub struct ConfigResolution {
+    source: ConfigSource,
+    selected_path: Option<PathBuf>,
+    skipped_existing_paths: Vec<PathBuf>,
+}
+
+/// The only policy this build implements. Reported so a reader does not have to
+/// infer from a single path whether later files were merged in.
+pub const SEARCH_POLICY: &str = "first_existing_no_merge";
+
+impl ConfigResolution {
+    /// An explicit `--config <path>`. The automatic search list is not consulted,
+    /// so nothing is reported as skipped: those files did not participate.
+    pub fn explicit_path(path: impl AsRef<Path>) -> Self {
+        Self {
+            source: ConfigSource::ExplicitPath,
+            selected_path: Some(absolutize(path.as_ref())),
+            skipped_existing_paths: Vec::new(),
+        }
+    }
+
+    /// A file chosen from the search list, plus any later candidates that exist
+    /// and were therefore shadowed by it.
+    pub fn search_path(selected: impl AsRef<Path>, skipped: &[PathBuf]) -> Self {
+        Self {
+            source: ConfigSource::SearchPath,
+            selected_path: Some(absolutize(selected.as_ref())),
+            skipped_existing_paths: skipped.iter().map(|p| absolutize(p)).collect(),
+        }
+    }
+
+    /// No candidate existed. `selected_path` is `None`.
+    pub fn defaults() -> Self {
+        Self {
+            source: ConfigSource::Defaults,
+            selected_path: None,
+            skipped_existing_paths: Vec::new(),
+        }
+    }
+
+    /// Used by entry points that do not track the decision, so provenance is
+    /// absent rather than wrong.
+    pub fn unavailable() -> Self {
+        Self {
+            source: ConfigSource::Unavailable,
+            selected_path: None,
+            skipped_existing_paths: Vec::new(),
+        }
+    }
+
+    pub fn source(&self) -> ConfigSource {
+        self.source
+    }
+
+    pub fn selected_path(&self) -> Option<&Path> {
+        self.selected_path.as_deref()
+    }
+
+    pub fn skipped_existing_paths(&self) -> &[PathBuf] {
+        &self.skipped_existing_paths
+    }
+}
+
+impl Default for ConfigResolution {
+    fn default() -> Self {
+        Self::unavailable()
+    }
+}
+
+/// Absolute form for diagnosis, because a relative candidate such as
+/// `konnect.toml` is meaningless to a reader who does not know the server's
+/// working directory. `canonicalize` also resolves symlinks, which is what makes
+/// the answer checkable; if it fails the path is returned unchanged rather than
+/// dropped.
+fn absolutize(path: &Path) -> PathBuf {
+    path.canonicalize().unwrap_or_else(|_| {
+        if path.is_absolute() {
+            path.to_path_buf()
+        } else {
+            std::env::current_dir()
+                .map(|dir| dir.join(path))
+                .unwrap_or_else(|_| path.to_path_buf())
+        }
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn defaults_report_no_selected_path() {
+        let resolution = ConfigResolution::defaults();
+        assert_eq!(resolution.source().as_str(), "defaults");
+        assert!(resolution.selected_path().is_none());
+        assert!(resolution.skipped_existing_paths().is_empty());
+    }
+
+    #[test]
+    fn unavailable_is_not_reported_as_defaults() {
+        // The distinction #419's reviewer asked for: an entry point that did not
+        // track the load must not look like a clean defaults selection.
+        assert_eq!(
+            ConfigResolution::unavailable().source().as_str(),
+            "unavailable"
+        );
+        assert_ne!(
+            ConfigResolution::unavailable().source(),
+            ConfigResolution::defaults().source()
+        );
+    }
+
+    #[test]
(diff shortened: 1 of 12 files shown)


neusse:
This remains a strong, focused implementation of #419, but please hold the current head while #382 is corrected. Both PRs change config.rs, ffi.rs, main.rs, and the startup/config-resolution contract.

The lowest-conflict order is:

1. settle and merge the startup-discovery contract in #382 / #424;
2. reconstruct this PR once on that current main;
3. preserve this PR's unique outcome: report the immutable configuration resolution that actually controlled startup, including selected and shadowed paths without values or credentials;
4. ensure the response also remains honest when the preceding discovery/config code returns unresolved or unavailable state;
5. rerun the complete gate and update the exact base/head evidence.

Do not repeatedly rebase while #382 is changing. Its present all-green head is good historical evidence, but the final merge review must use the reconstructed head. Closes #419 is already correct.

neusse:
Dependency cleared: #382 has merged as 2de06bf6a0a8532a1bb5a72dec1fd75e9b8238ff, and its full post-merge Rust gate is green.

Please reconstruct this PR once on current `main`, preserving its unique outcome: the running process reports the immutable provenance of the configuration file that actually configured it. Then push the rebuilt head and let all ten required checks run. This avoids carrying stale branch ancestry forward now that the discovery/config foundation is settled.

Once that is done, we can review the new exact head for merge.

neusse:
Maintainer reconstruction completed on current `main` (`2de06bf`, including #382).

New exact head: `2f2921e68505929c54154a03d594d254c6d09e50`

The PR is now one unique commit on current main, with the original author and co-author preserved. The conflict resolution keeps both intended outcomes:

- #382 remains the single startup path for IPC resolution, including config/env/detected precedence, whitespace handling, endpoint redaction, and FFI startup diagnostics.
- #421 adds immutable configuration-file provenance to that path and reports it through `get_installation_info` without re-running discovery.
- The FFI fallback behavior remains unchanged, but a failed load reports provenance as `unavailable` rather than claiming clean defaults.

Local evidence on the reconstructed head:

- `cargo fmt --all -- --check` — passed
- `cargo clippy --workspace --locked --all-targets -- -D warnings` — passed
- `cargo test --workspace --locked --lib --tests` — passed
- `cargo test --workspace --locked --doc` — passed
- all 17 `protocol_stdio` tests passed
- negative control: bypassing provenance propagation made `installation_info_names_the_config_file_that_configured_startup` fail with `unavailable` instead of `search_path`; restoring the propagation made the suite pass

GitHub reports the PR mergeable. The ten hosted checks are rerunning now on the new exact head, so this is waiting on review/CI rather than on the author.

neusse:
Exact-head CI update: all ten required checks have passed on `2f2921e68505929c54154a03d594d254c6d09e50`. GitHub now reports `MERGEABLE` / `CLEAN`.

PR #443, opened 2026-09-06
fix(manufacturing): emit matched JLCPCB BOM and CPL files

### LINKED ISSUE #440: export_manufacturing_package's position file uses KiCad-native headers, not JLCPCB's CPL schema — upload fails with "Failed processing the CPL file"
## Summary

`export_manufacturing_package`'s generated position file (`positions.csv`) uses KiCad's native footprint-position-report column headers (`Ref,Val,Package,PosX,PosY,Rot,Side`), even when `fab_house` is explicitly set to `"jlcpcb"`. JLCPCB's own web uploader (BOM & CPL step) expects a specific CPL schema (`Designator,Mid X,Mid Y,Layer,Rotation`, as produced by e.g. the community `kicad-jlcpcb-tools` (bouni) plugin). Uploading Konnect's file directly as the "CPL File" / Pick&Place slot fails outright with a generic error.

## Environment

- Konnect 0.11.0
- KiCad 10.0, macOS

## Repro

```
export_manufacturing_package(
  board=".../myboard.kicad_pcb",
  schematic=".../myboard.kicad_sch",
  fab_house="jlcpcb",
  output_dir=".../fab_export",
  bom_fields="Reference,Value,Footprint,LCSC",
  bom_labels="Designator,Comment,Footprint,LCSC Part #",
  bom_group_by="Value,Footprint,LCSC"
)
```

Resulting `fab_export/positions.csv` header row:

```
Ref,Val,Package,PosX,PosY,Rot,Side
```

Uploading this file as JLCPCB's "CPL File" (Bill of Materials step, Pick&Place slot, after the BOM file had already been accepted) → immediate modal: "Error / Failed processing the CPL file."

For comparison, a `kicad-jlcpcb-tools` (bouni) run against the identical board produced two files matching JLCPCB's own upload-step terminology:

```
BOM-myboard.csv
CPL-myboard.csv   → header: Designator,Val,Package,Mid X,Mid Y,Rotation,Layer
```

which JLCPCB's uploader accepted and processed without error, using the same underlying footprint positions. (bouni's plugin also applies a per-footprint rotation correction, e.g. "Fixed rotation of U2 ... by 270 degrees" — a separate concern from the header issue, since the header names alone are enough to make JLCPCB's parser reject Konnect's file before rotation values would ever matter.)

Mixing files from the two generators (e.g. Konnect's `bom.csv` + bouni's CPL file) also reproduced the same "Failed processing the CPL file" error — only a matched pair from the same generator run succeeded, which is what isolated the header schema as the actual variable rather than a data-content problem.

## Expected

Passing `fab_house="jlcpcb"` should produce a position/CPL file JLCPCB's own uploader can consume directly, matching the vendor's documented CPL column schema (`Designator, Mid X, Mid Y, Layer, Rotation`).

## Actual

The position file is emitted with KiCad's generic native report headers regardless of the `fab_house` parameter, and is rejected by JLCPCB's web importer.

## Suggested fix

- When `fab_house="jlcpcb"`, emit (or additionally emit) a CPL file using JLCPCB's expected header names/order (`Designator,Mid X,Mid Y,Layer,Rotation`), the way `kicad-jlcpcb-tools` does. Per-footprint rotation correction for JLCPCB's placement convention (needed for certain packages, e.g. SSOP) can be tracked as a separate, smaller follow-up if it's out of scope here — but the header-schema mismatch alone is enough to hard-fail the upload and should be fixed regardless of whether rotation correction is included.
- Separately, consider naming the `fab_house`-targeted output files after the vendor's own upload-step terminology rather than generic `bom.csv`/`positions.csv` — e.g. `BOM-<project>.csv` / `CPL-<project>.csv`, matching what `kicad-jlcpcb-tools` does and what JLCPCB's own uploader labels its two upload slots ("BOM File" / "CPL File"). `bom.csv` already reads unambiguously, but `positions.csv` gives no hint it's the file meant for the "CPL File" slot — a naming mismatch that compounds the schema mismatch above when deciding which generated file goes where.

## Impact

`fab_house` is the export tool's own vendor-targeting parameter, so the natural expectation is a ready-to-upload JLCPCB file set. Instead the CPL/position file silently fails JLCPCB's own uploader, forcing a fallback to a separate community plugin for exactly the file the parameter name promises to produce correctly.

## Related

None found — searched open and closed issues for "CPL", "position file", "pick and place", "export_manufacturing_package"; no duplicates.

## Affected versions

Confirmed on Konnect 0.11.0 / KiCad 10.0.

### LINKED ISSUE #441: export_manufacturing_package's BOM uses dash-range designator compression (e.g. "C3-C18"), which JLCPCB's uploader doesn't expand
## Summary

`export_manufacturing_package`'s generated `bom.csv` groups references using KiCad-native dash-range compression (e.g. `"C1,C3-C18"`, `"J1-J16"`) whenever a run of 3+ consecutive reference numbers shares the same value/footprint/LCSC part. JLCPCB's own BOM/CPL uploader does not expand these ranges — it treats `"C3-C18"` as one opaque literal designator string and cross-checks it against the CPL file's individually-listed designators (`C3`, `C4`, … `C18`), which don't match, so it reports every one of those references as missing from the BOM (and, symmetrically, missing from the CPL, since the CPL has them individually but the BOM only has the compressed range).

This was found as a follow-on to #440: once the CPL header-schema fix from that issue is applied by hand (so BOM and CPL both parse structurally), this range-compression mismatch is the next thing that blocks a successful JLCPCB upload using Konnect's own `export_manufacturing_package` output exclusively.

## Environment

- Konnect 0.11.0
- KiCad 10.0, macOS

## Repro

`export_manufacturing_package(..., fab_house="jlcpcb", bom_group_by="Value,Footprint,LCSC")` on a board with references `C1`–`C18` (mostly 100nF, same footprint/LCSC) and `J1`–`J16` (all XH-2P, same footprint/LCSC) produced:

```
"Designator","Comment","Footprint","LCSC Part #"
"C1,C3-C18","100nF","Capacitor_SMD:C_0805_2012Metric","C49678"
"J1-J16","XH-2P","Connector_JST:JST_XH_B2B-XH-A_1x02_P2.50mm_Vertical","C158012"
```

(References that only span 1–2 items, e.g. `"HDR1,HDR2"`, `"R1,R2"`, stayed as plain comma lists — only runs of 3+ consecutive numbers got dash-compressed.)

Uploading this BOM alongside a CPL that lists each of those references individually (`C3`, `C4`, ..., `C18`, `J1`, ..., `J16` as separate rows — which is the only way a CPL can represent per-component X/Y/rotation) to JLCPCB's BOM & CPL step, then clicking through to Component Placements, produced:

```
The below parts won't be assembled due to data missing.
C3,C4,C5,C6,C7,C8,C9,C11,C10,C13,C12,C15,C14,J1,C17,J2,J10,C16,J3,J4,J12,C18,J5,J11,J6,J14,J7,J13,J8,J16,J9,J15
designators don't exist in the BOM file.
C3-C18,J1-J16 designators don't exist in the CPL file.
```

For comparison, `kicad-jlcpcb-tools` (bouni) — which was already confirmed to work end-to-end against the identical board — emits the same grouping fully enumerated instead of range-compressed:

```
100nF,"C1,C10,C11,C12,C13,C14,C15,C16,C17,C18,C3,C4,C5,C6,C7,C8,C9",C_0805_2012Metric,C49678,17
...
XH-2P,"J1,J10,J11,J12,J13,J14,J15,J16,J2,J3,J4,J5,J6,J7,J8,J9",JST_XH_B2B-XH-A_1x02_P2.50mm_Vertical,C158012,16
```

## Expected

Passing `fab_house="jlcpcb"` should produce a BOM whose designator lists are fully enumerated (comma-separated, no dash-ranges), matching what JLCPCB's uploader — and the community tooling built specifically for it — expects.

## Actual

The BOM's designator grouping always applies KiCad's native dash-range compression regardless of `fab_house`, which breaks JLCPCB's BOM/CPL cross-check for any group of 3+ consecutive same-part references — a very common case (multiple identical decoupling caps, connectors, etc. placed in ascending reference order).

## Suggested fix

When `fab_house="jlcpcb"`, disable dash-range compression in the BOM designator column and fully enumerate each reference comma-separated instead — same output shape `kicad-jlcpcb-tools` already uses. This is a smaller, more contained fix than #440's CPL schema gap (same designator-list values, just a different join strategy), but both are needed together before `export_manufacturing_package(fab_house="jlcpcb")`'s own output is directly JLCPCB-uploadable without a third-party plugin.

## Impact

Any board with 3+ consecutive same-part references (e.g. a row of decoupling capacitors or identical connectors, both extremely common) silently fails JLCPCB's assembly upload with a confusing "designators don't exist" error that doesn't point at the real cause — the range-vs-enumerated mismatch between the two generated files isn't obvious without diffing them against a known-good reference (bouni's plugin output, in this case).

## Related

#440 (same underlying goal: `fab_house="jlcpcb"` should produce a directly-uploadable file pair). Searched open/closed issues for "designator range", "bom_group_by", "compression" — no duplicate found.

## Affected versions

Confirmed on Konnect 0.11.0 / KiCad 10.0.

neusse:
## Summary

Closes #440
Closes #441

Make `export_manufacturing_package(fab_house="jlcpcb")` produce one structurally validated JLCPCB BOM/CPL pair instead of presenting KiCad-native CSV as upload-ready vendor output.

## Base and dependency status

Rebuilt on current `upstream/main` at `3cde785`. The previously declared prerequisites #382, #421, and #442 are merged; no code dependency remains. The branch contains three unique commits and has a clean merge relationship with current main.

## Root cause and design

The JLCPCB path previously returned KiCad's native `Ref,Val,Package,PosX,PosY,Rot,Side` position file and could report the package complete even though JLCPCB rejects that schema. KiCad also compressed grouped BOM references into ranges such as `C3-C18`, which JLCPCB treats as an opaque designator rather than expanding.

This change:

- keeps KiCad CLI as the geometry and BOM source;
- parses native position CSV structurally with the Rust `csv` crate;
- emits `CPL-<project>.csv` with `Designator,Mid X,Mid Y,Layer,Rotation`;
- validates coordinates, rotation, side, duplicates, required columns, non-empty output, and millimetre units;
- emits `BOM-<project>.csv` with KiCad's empty `--ref-range-delimiter`, so references are individually enumerated;
- passes `--exclude-dnp` to both native JLCPCB exports;
- compares the complete BOM and CPL designator sets and fails closed on any mismatch, including asymmetric board/schematic exclusion flags;
- preserves generic and other-fabricator filenames, inclusion policy, native CSV behavior, and public Rust call signatures;
- preserves `files_generated.type="pick_and_place"`; and
- documents the public MCP migration and the remaining package-specific rotation-preview limitation.

### #440

- [x] JLCPCB path no longer publishes native KiCad headers as a CPL
- [x] CSV is parsed and emitted structurally
- [x] vendor filenames, headers, metric units, sides, numeric values, and duplicate references are validated
- [x] generic position export behavior is preserved
- [x] real KiCad 10 position output is committed as the parser fixture
- [x] DNP parts are excluded symmetrically and real-KiCad exclusion behavior is exercised
- [x] BOM/CPL designator mismatch fails closed
- [x] matched output from exact head `69c6846` is accepted by JLCPCB's actual uploader

### #441

- [x] JLCPCB BOM export passes KiCad's native empty range delimiter
- [x] grouped references are represented as individual comma-separated designators
- [x] compressed ranges are rejected during package validation
- [x] generic BOM export retains KiCad's default range behavior
- [x] a real KiCad hierarchical schematic proves groups of 1, 2, and 3+ references
- [x] the output-level negative control proves restoring KiCad's default delimiter reintroduces a rejected range
- [x] matched output from exact head `69c6846` is accepted by JLCPCB's actual uploader

This is the single terminal PR for both issues.

## Evidence

Exact head: `69c68462c2723cc32d81e5ade1dce756c57afcfe`

Current Windows local gate:

- `cargo test --workspace --locked --lib --tests` — pass
- two ignored live KiCad 10 tests enabled explicitly — pass

All ten required GitHub checks pass on the exact head.

The live tests use checked-in KiCad-authored demo fixtures. One runs KiCad's grouped BOM exporter over a hierarchical schematic and then reruns it with the delimiter guard removed. The other runs the real BOM and position exporters over a matched schematic/board pair containing exclusion flags and verifies identical transformed populations.

The exact-head executable generated a fresh JLCPCB package through the real MCP handler from the KiCad blinker benchmark. The handler returned `complete: true`, 12 verified non-empty artifacts, and no warnings. JLCPCB then:

- accepted the exact Gerber archive and detected a 2-layer 45 x 30 mm board;
- processed `BOM-konnect-jlcpcb-blinker.csv` and `CPL-konnect-jlcpcb-blinker.csv` together;
- reported 10 parts detected and all 10 parts confirmed; and
- opened Component Placements with every expected designator represented and no missing-designator warning.

This validates the uploader schema and the grouped-reference cross-check. Visual inspection found U1 rotated 90 degrees from the intended orientation; package-specific rotation and offset correction is now tracked explicitly in #518. This PR does not claim physical orientation validation, and the JLCPCB placement preview remains mandatory before ordering.

The lockfile was regenerated against current main and contains only the new `csv`/`csv-core` dependency additions; stale unrelated lockfile churn was removed.

## Compatibility, risk, and rollback

JLCPCB assembly callers must consume the new vendor filenames and use millimetres. DNP components are now excluded from both JLCPCB assembly files. The public type discriminator, public Rust export function signatures, and non-JLCPCB behavior are unchanged.

The generated CPL intentionally does not apply package-specific rotation corrections. The JLCPCB order preview remains an assembly release gate and the response does not claim orientation acceptance.

Rollback is the three focused commits. It removes the vendor adapter and CSV dependency without changing KiCad project files or the generic export contract.


neusse:
First maintainer pass on draft head `cae7ef5e2ff970aa18a3cc99b86af6cc507f613b` found one acceptance gap beyond the already-recorded JLCPCB uploader check:

- #441 requires a **real KiCad grouped BOM export** containing runs of 1, 2, and 3+ same-part references, followed by validation that every JLCPCB designator is individually enumerated. The current branch proves argument construction and parses supplied BOM text, but it does not yet run KiCad against a real grouped schematic and inspect the generated BOM. Add that real-KiCad fixture/evidence before the uploader test.

The least-resistance dependency order has also changed: #382 is blocked on live Linux/macOS/Windows evidence, while #442 can finish independently. Land #442 when corrected; settle #382; reconstruct #421 after #382; then rebuild this PR once on the resulting current main. Do not repeatedly rebase this draft while those prerequisites are moving.

Keep `Part of #440` / `Part of #441` until both the real-KiCad grouped-export result and actual JLCPCB matched-pair upload are observed. At that point this can become the single terminal PR with `Closes #440` and `Closes #441`.

neusse:
Exact-head JLCPCB validation is complete for the scope of #440/#441:

- JLCPCB accepted the Gerber archive generated by `69c68462c2723cc32d81e5ade1dce756c57afcfe` and detected a 2-layer 45 x 30 mm board.
- It processed `BOM-konnect-jlcpcb-blinker.csv` and `CPL-konnect-jlcpcb-blinker.csv` together.
- It reported 10 parts detected and all 10 confirmed.
- Component Placements opened with all expected designators and no missing-designator or compressed-range warning.

Visual review also caught U1 rotated 90 degrees from its intended physical orientation. That is the separate package-correction concern explicitly excluded from #440/#441's approved scope, not an uploader-schema failure. It is now tracked as P0 design issue #518 with the exact KiCad/CPL evidence and a data-driven correction recommendation. This PR therefore closes the two structural upload defects while retaining the documented placement-preview gate and making no physical-orientation claim.

PR #480, opened 2026-09-07
fix(ipc): bind operations to the requested board

### LINKED ISSUE #384: Bind generic KiCad IPC operations to the requested board document
## Problem

Board-targeted MCP operations first find the requested board through `GetOpenDocuments`, but lower-level generic IPC helpers can independently rediscover a board. With multiple boards open, that creates a target-substitution risk: the path-bearing tool validated one document while a later generic request can carry the first open board's `DocumentSpecifier`.

This is distinct from #241's handler coverage for `refuse_if_board_open_in_kicad`: it changes the production IPC target binding and proves the typed request header remains attached to the originally requested board.

## Proposed behavior

- Bind the exact typed `DocumentSpecifier` returned for the requested path to the subsequent IPC operation.
- Re-observe the binding before generic reads or writes.
- Refuse zero, wrong, duplicate, or disappeared document identities as structured `wrong_document`, `ambiguous_target`, or `stale_target` errors.
- Never fall through to a different open board or mark the requested board live when identity was not established.
- Preserve the existing single-open-document behavior.

## Acceptance

- A mock lists two boards with the requested board second; subsequent generic read and write headers name that second board.
- Zero, wrong, duplicate, and disappeared document cases fail with the appropriate typed classification.
- A wrong-document mutation never invokes its write closure and leaves the requested file byte-identical.
- The explicit low-level `*_in(DocumentSpecifier, ...)` APIs remain available for already typed callers.

neusse:
## Summary

Reconstructs #390 from current `upstream/main`, preserving the original contributor's commit authorship while reconciling exact-board IPC binding with the guarded file-fallback behavior merged in #407.

- binds generic IPC reads and writes to the exact typed `DocumentSpecifier` selected for the requested board
- re-observes that binding before every generic operation and refuses changed or disappeared targets
- preserves #407's safety boundary: file fallback is eligible only when KiCad positively proves the requested board is not open
- keeps unresolved open-document identities fail-closed as `ambiguous_open_board`
- reports duplicate requested targets as `ambiguous_target`, a different open board as `wrong_document`, and a disappeared binding as `stale_target`
- preserves the explicit low-level `*_in(DocumentSpecifier, ...)` APIs

This replaces the stale cumulative branch in #390. The focused commit remains authored by @dubesinhower.

Closes #384

## Validation

Passed on Windows against exact head `44c9893d9d3b485b18ad4f67b6a62f74d2aee571`:

- focused `konnect-ipc` mock tests for requested-board-second read/write headers, zero/wrong/duplicate targets, and disappeared bindings
- focused board-session tests for #407 fallback and ambiguity behavior

Real multi-board KiCad GUI validation was not run in this reconstruction environment; the typed mock protocol tests cover the deterministic request headers and refusal paths. This remains explicit pre-release validation debt, not a merge-wall for the focused, reversible change.

## Risk and rollback

Risk is limited to IPC board-target selection and the error taxonomy returned when the requested live document cannot be proven. Reverting this single commit restores current-main behavior.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -260,8 +260,9 @@ Tool-call failures are typed via the `ToolErrorKind` enum in `crates/konnect-cor
 | `invalid_argument` | Required argument missing/malformed |
 | `file_not_found` | Referenced file or project-discovery directory does not exist or cannot be read |
 | `conflict` | The file changed, a write would replace existing paths, or schematic project ownership cannot be proven uniquely — carries the affected paths; ownership conflicts include the schematic directory and all candidate roots |
-| `ambiguous_target` | More than one observed UUID, reference/unit, field, or instance identity matches the requested target, or component instance records conflict across projects, units, or hierarchy paths — carries `target` and the stable `candidates`; the caller must choose rather than retrying blindly |
-| `stale_target` | Saved symbol instance metadata disagrees with the proven hierarchy, or component readback lacks the expected document, UUID/reference, unit, library, hierarchy, position, rotation, or property value — carries `target` and `reason`; preflight refuses before writing, while a readback failure may follow a committed write |
+| `wrong_document` | The requested PCB is not among KiCad's positively identified open documents — carries `requested` and `open_documents`; path-bearing IPC operations never substitute another open board |
+| `ambiguous_target` | More than one observed UUID, reference/unit, field, instance, or requested PCB document identity matches the requested target, or component instance records conflict across projects, units, or hierarchy paths — carries `target` and the stable `candidates`; the caller must choose rather than retrying blindly |
+| `stale_target` | A previously bound PCB document is no longer uniquely open, saved symbol instance metadata disagrees with the proven hierarchy, or component readback lacks the expected document, UUID/reference, unit, library, hierarchy, position, rotation, or property value — carries `target` and `reason`; preflight refuses before writing, while a readback failure may follow a committed write |
 | `ambiguous_open_board` | KiCad answered, and its open-document list could not be read as a complete set of comparable board identities — carries `path`; neither the live nor the file path may run |
 | `handler_error` | Catch-all for unmigrated `anyhow::Error` returns |
 
--- a/crates/konnect-core/src/mcp/error.rs
+++ b/crates/konnect-core/src/mcp/error.rs
@@ -58,6 +58,12 @@ pub enum ToolErrorKind {
         target: String,
         candidates: Vec<String>,
     },
+    /// The requested document is not the document set observed in the target
+    /// editor, so proceeding would answer about or mutate another file.
+    WrongDocument {
+        requested: String,
+        open_documents: Vec<String>,
+    },
     /// The caller named a target, but its observed editor or document state
     /// no longer agrees with the state required to mutate it safely.
     StaleTarget { target: String, reason: String },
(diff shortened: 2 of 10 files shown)


dubesinhower:
I saw #390 was replaced while my reconstruction was being validated, and #480 has since merged. I checked its exact reviewed head `44c9893d9d3b485b18ad4f67b6a62f74d2aee571` and reproduced two remaining gaps in the requested fail-closed contract with added tests only; production code was unchanged. These findings are against that head, not a claim that I have re-audited subsequent main changes.

The [three reproduction tests are published in `a3abdac`](https://github.com/dubesinhower/Konnect/commit/a3abdacfbf6b4edabc3d188049776fffa462cb12), directly on the reviewed #480 head. This is intentionally failing evidence, not a merge-ready implementation. The [independently verified reconstruction is preserved as `f60f477`](https://github.com/dubesinhower/Konnect/commit/f60f4774a9fec0bff709547ecc9aeac83e5f827c) for guard/test reference. Both are fork branches; no competing PR was opened.

1. **A changed typed binding is accepted.** In `get_board_document`, the successful selector arm calls `bind_board` again without comparing the new `DocumentSpecifier` with the captured one. `find_open_board` similarly replaces the binding. A mock first reports `target.kicad_pcb` with project name `konnect-mock`, then reports the same board filename and project directory with project name `replacement`. The generic `GetItems` call proceeds and returns `Ok([])`; repeating the explicit lookup returns the replacement specifier. Neither returns `StaleDocument`. This contradicts the PR's changed-target refusal claim. The narrow correction is to compare the full captured specifier and refuse a change, rather than refresh the binding. Same-specifier close/reopen remains an unavoidable protocol limitation and should stay documented separately.

2. **An unresolved single document reaches a generic mutation.** The unbound `[document]` arm ignores `board_document_identity` failure and returns the document anyway. A mock reports one bare `unknown.kicad_pcb` with `project: None`; `create_items` with a non-empty item reaches `CreateItems`. The refusal test's mock catches that command, then the client eventually reports the receive timeout. It should return `UnresolvedDocumentIdentities` before the command. Propagate the failed identity result rather than using `if let Ok(...)`.

Executed on Windows with Rust 1.96.0, against the replacement production tree:

```text
cargo test --locked -p konnect-ipc --test mock_server_test repeating_lookup_cannot_replace_the_bound_typed_document
# FAILED: unwrap_err() received Ok(DocumentSpecifier { project.name: "replacement", ... })
cargo test --locked -p konnect-ipc --test mock_server_test generic_reobservation_cannot_replace_the_bound_typed_document
# FAILED: unwrap_err() received Ok([])
cargo test --locked -p konnect-ipc --test mock_server_test generic_helpers_refuse_unbound_ambiguous_or_unidentifiable_documents
# FAILED: mock observed a mutation command before refusal
```

My preserved local reconstruction fixes these guards and passes all four contributor gates (1,588 tests, 8 doctests; 23/2 ignored). Six negative controls catch the disabled guards, including these cases. Its taxonomy retains the original #407 variants, so the complete local reconstruction should not be transplanted over #480; the guard corrections and tests can be adapted to #480's `BoardTargetError` taxonomy.

Live evidence remains separate: my Windows attempt could only reach a project-manager endpoint (`GetOpenDocuments` returned `AS_UNHANDLED`), so no live multi-board result is claimed. All three failures above are deterministic protocol tests, not live KiCad observations.

neusse:
Thanks, @dubesinhower. I reproduced all three failures against the merged head and confirmed both gaps are real. Your reproduction tests are preserved with authorship in focused follow-up PR #487, which makes the full typed binding immutable across repeat lookup and refuses unresolved single-document identities before mutation. Issue #384 has been reopened and #487 is its terminal closure path.

PR #482, opened 2026-09-07
fix(validation): reject malformed nested library inputs

### LINKED ISSUE #234: Array-item schemas declare required fields and default them: batch_add_wire adds a (0,0)-(0,0) wire

neusse:
Closes #234

Reconstructs the still-needed portion of #268 on current `main`, preserving nordic-style's authorship. PR #318 already fixed `batch_add_wire`, so this replacement intentionally contains only the remaining library-validation work.

## What changed

- Validates every nested `create_footprint` pad before reading or writing the output file.
- Validates `create_symbol` top-level pins, per-unit pins, and shared power pins before reading or writing the symbol library.
- Returns an indexed structured `invalid_argument` field such as `pads[0].number` or `units[0].pins[1].name`.
- Rejects the complete request atomically instead of inventing plausible default pads or incomplete symbol units.
- Leaves valid payload schemas and output unchanged.

## Issue accounting

- #318 completed the `batch_add_wire` criterion.
- This PR is the terminal change for the remaining `create_footprint` and `create_symbol` criteria.

## Evidence

- Focused create-footprint/create-symbol tests: 6 passed.
- Existing target files remain byte-identical after malformed requests; new targets are not created.
- Negative control: temporarily disabling pad validation made the malformed-pad regression fail before writing could be safely attempted; restoring the guard made it pass.

All full gates passed locally.

## Risk and rollback

Only malformed payloads that previously relied on undocumented defaults change behavior. The change is one focused commit and can be reverted directly.

Supersedes #268.

### DIFF
--- a/crates/konnect-core/src/tools/library.rs
+++ b/crates/konnect-core/src/tools/library.rs
@@ -703,16 +703,21 @@ async fn handle_create_footprint(
         Ok(v) => v.clone(),
         Err(e) => return Ok(e),
     };
+    if let Err(error) = validate_footprint_pad_items(&pads_val) {
+        return Ok(error);
+    }
     let mut pad_geoms: Vec<PadGeom> = Vec::new();
     let mut pad_sexp = String::new();
     for pad in &pads_val {
-        let number = pad["number"].as_str().unwrap_or("1").to_string();
-        let pad_type = pad["type"].as_str().unwrap_or("smd").to_string();
-        let shape = pad["shape"].as_str().unwrap_or("rect");
-        let x = pad["x"].as_f64().unwrap_or(0.0);
-        let y = pad["y"].as_f64().unwrap_or(0.0);
-        let w = pad["width"].as_f64().unwrap_or(1.0);
-        let h = pad["height"].as_f64().unwrap_or(1.0);
+        // Safe after validate_footprint_pad_items: every schema-required field
+        // has the declared type, so no plausible-looking pad is invented.
+        let number = pad["number"].as_str().expect("validated").to_string();
+        let pad_type = pad["type"].as_str().expect("validated").to_string();
+        let shape = pad["shape"].as_str().expect("validated");
+        let x = pad["x"].as_f64().expect("validated");
+        let y = pad["y"].as_f64().expect("validated");
+        let w = pad["width"].as_f64().expect("validated");
+        let h = pad["height"].as_f64().expect("validated");
 
         let layer_names = if let Some(layers) = pad["layers"].as_array() {
             let mut names = Vec::with_capacity(layers.len());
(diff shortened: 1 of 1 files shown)


mixelpixx (approved):
Reviewed exact head `4b9131c89153c00dd6c55ffeae6f7ae4bdb2e292` (mergeable clean against `eddf131`; ten checks green; no threads).

Ran here (Windows, worktree on the exact head):

```
cargo test -p konnect-core --lib required_argument_tests        7 passed
cargo test --workspace --locked --lib --tests                   exit 0
```

Two neuters, both caught:

| guard removed | test | result |
|---|---|---|
| `validate_footprint_pad_items` call in `handle_create_footprint` | `create_footprint_rejects_a_malformed_pad_without_writing` | FAILED (panic at `library.rs:711`, the `.expect("validated")` on `pad["number"]`) |
| `validate_pins(unit_pins, …)` in `parse_symbol_items` | `create_symbol_names_the_malformed_nested_pin` | FAILED (assertion at `library.rs:8116`) |

Checked the one thing that could have been a regression: `parse_symbol_items` demands `x`/`y` on every `power_pins` entry. The schema already does (`pin_item_schema(…, true)` for `power_pins`), so this enforces the published contract rather than tightening it. Existing-file byte-identity and no-new-file on refusal are asserted, which is the property that matters for #234.

One note, not a block: the pad loop now relies on `.expect("validated")` for every field the validator covers. That couples the two lists silently — a future pad field read with `expect` but not added to `validate_footprint_pad_items` becomes a panic inside a tool call rather than a structured refusal. A one-line comment on the validator naming that coupling, or reading the fields through the validator's own parsed struct, would close it; fine to do later.

Issue accounting is right: #318 took the `batch_add_wire` criterion, this is the terminal `Closes #234`. Moving to `status:ready-to-merge` and merging. Successor to promote: #502 (rebuild on the resulting `main` per your 09-08 comment).

PR #484, opened 2026-09-07
feat(mcp): add a safe Unix stdio reload tool

neusse:
Reconstructs #176 on current `main`, preserving Jean-Yves POCHEZ's original authorship while implementing the safety rework requested in review.

## What changed

- Keeps the existing seven meta-tools as the cross-platform baseline.
- Advertises and dispatches `reload_server` only from the standalone Unix executable when it runs exclusively over stdio.
- Does not expose the tool on Windows, HTTP, mixed HTTP+stdio, or embedded/KiCad-hosted sessions.
- Preserves every original argument after `argv[0]`, including an explicit `--config` path.
- Opens the installed executable once and proves its version before queuing the handoff.
- On Linux/Android, probes and executes the already-open file through `/proc/self/fd`, so a concurrent pathname replacement cannot change the selected executable.
- On macOS and other Unix targets without a portable `fexecve`, records device, inode, length, and modification time and revalidates both the open file and named path immediately before `exec`; ordinary atomic build/install replacement is refused.
- Refuses downgrades and requires `allow_same_version=true` for an intentional same-version development rebuild.
- Queues a one-shot reload plan instead of starting a timer inside the handler.
- Flushes the response and synchronous notifications, stops accepting requests, releases the run-registry record, and only then performs the one-way `exec`.
- Keeps the irreversible exec function in the standalone binary. The embedded library explicitly refuses any impossible reload handoff as defense in depth.

## Why the transport handoff matters

The original 250 ms timer allowed the stdio loop to accept a second mutation before `exec`, potentially interrupting a write before cleanup guards ran. The transport now returns a sealed handoff before reading another request. The standalone entry point drops its run-record guard before `exec`, so the replacement process can register the same PID cleanly.

## Evidence

Windows-local evidence:

- Focused reload tests cover capability gating, one-shot handoff, disabled dispatch, upgrade/equal/downgrade policy, and exact argument preservation.
- Negative control: temporarily neutralizing the downgrade refusal makes the version-policy regression test fail; restoring the guard makes it pass.
- Full gate passed:

Hosted Unix tests exercise the platform-specific handoff guards:

- Linux copies and opens the test executable, removes its pathname, and proves the open executable can still be invoked through its descriptor.
- macOS captures the candidate identity, proves an unchanged candidate is accepted, atomically replaces the named candidate, and proves the final handoff is refused.

## Risk and rollback

The tool is absent unless the process proves the supported capability combination. The irreversible action requires `confirm=true`; same-version builds require a second explicit opt-in. The PR contains the preserved contributor commit plus one focused safety/adaptation commit, either of which can be reverted normally.

Supersedes #176.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -77,7 +77,7 @@ Konnect/
 │   │       ├── router/
 │   │       │   ├── mod.rs           # ToolRouter: load/unload toolsets
 │   │       │   ├── registry.rs      # Static toolset metadata + tools_for() dispatcher
-│   │       │   └── meta_tools.rs    # 7 always-visible meta-tools
+│   │       │   └── meta_tools.rs    # 7 baseline meta-tools + Unix stdio reload
 │   │       └── tools/
 │   │           ├── mod.rs            # ToolDef, ToolContext, tool! macro, helpers, kicad_config_dir()
 │   │           ├── cli.rs            # kicad-cli v10 subprocess wrapper (verified against actual binary)
@@ -396,6 +396,7 @@ convention for other `kicad-cli`-calling code.
 ## Current Stats
 
 - **21 toolsets, 226 tools** + 7 meta-tools (4 routing + 2 observability + 1 runtime diagnostic — see `tool-directory.md`)
+- Unix stdio adds one conditional maintenance meta-tool, `reload_server`; it is not registered for HTTP, mixed transport, or Windows.
 - Baseline `tools/list`: 21 tools / ~2K tokens (starter kit + meta-tools)
 - Full-catalog `tools/list` (all loaded): 233 tools (226 registered + 7 meta) / ~25K tokens
 - **0 IPC stubs** (all protobuf methods implemented)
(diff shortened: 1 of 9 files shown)


neusse:
Please reconstruct this PR from current `main` with only the unique Unix stdio reload change, then rerun all ten checks.

#442 has now merged, so the dependency is resolved, but this head is `DIRTY` against the resulting platform/startup code. The least-resistance path is a clean reconstruction rather than resolving the old cumulative history in place:

1. start from current `upstream/main` (`296641b` at this review);
2. carry only commit `e106549a`'s unique reload behavior, adapting it to the merged run registry;
3. regenerate tool counts with `cargo xtask fix-doc-counts` if the tool remains in the final design;
4. verify the unique commit list and diff, then run the complete gate;
5. push the reconstructed exact head for review.

This is now `status:waiting-on-author`; it can re-enter the queue after that focused head is clean and green.

PR #486, opened 2026-09-07
fix(board): refuse stale locked file fallbacks

### LINKED ISSUE #385: Use KiCad board locks as cold-start stale-file evidence

neusse:
## Summary

Refuse cold-start direct-file board mutations when the exact requested board has a KiCad sibling lock, or when that lock cannot be inspected reliably. This reconstructs only the unique work from #391 on current `main`, after the exact-document binding foundation in #480 landed.

Closes #385.

## Design

- Derive `~<filename>.lck` from the canonical exact board path; unrelated-board locks do not taint the request.
- Check the lock before both unreachable-IPC fallback gates.
- Recheck immediately before atomic replacement, catching a lock that appears after preflight.
- Reuse the structured `unsafe_file_fallback` contract with stable `reason` evidence:
  - `kicad_lock_present`
  - `kicad_lock_unreadable`
- Keep lock absence as only the absence of this veto; all #480 IPC/session/path-identity gates still apply.
- Never parse ownership from lock contents, infer safety from age, remove a lock, or add a force bypass.
- Preserve the existing schematic-lock conflict behavior.

## Recovery contract

Refusals leave board bytes and the lock untouched. Guidance tells the user to close or recover Pcbnew, reconcile unsaved work, and save the authoritative board. It does not tell an agent to remove the lock or retry automatically.

## Evidence

- Full local gate passed:
- Deterministic coverage includes present, stale-looking/malformed, unreadable, absent, and other-board locks; both fallback gates; byte-identical refusal; canonical lock identity; and a lock appearing between preflight and atomic replacement.
- Negative control: temporarily disabling the preflight lock guard made `any_exact_board_lock_refuses_both_offline_fallback_gates_without_writing` fail.
- Windows/KiCad 10.0.6 observation: opening a disposable board created the exact sibling lock with hostname/username content; forced process exit left the lock behind.
- `BLOCKED` observation debt: the non-interactive harness could not exercise a clean GUI close, and macOS/Linux KiCad lock lifecycles were not exercised locally. Hosted CI will provide supported-platform compile and deterministic regression coverage, but these GUI lifecycle observations remain explicit validation debt.

## Compatibility and risk

This intentionally narrows cold-start behavior: a stale lock can cause a recoverable false-positive refusal. That is preferable to writing a saved board that may be older than unsaved editor state. Rollback is the single focused commit.

## Provenance

The unique implementation originated in #391 by @dubesinhower; authorship is preserved on the reconstructed commit. The error contract and tests were adjusted to match the accepted maintainer decision in #385 rather than replaying the old draft's `stale_target` response.

### DIFF
--- a/crates/konnect-core/src/mcp/error.rs
+++ b/crates/konnect-core/src/mcp/error.rs
@@ -69,7 +69,7 @@ pub enum ToolErrorKind {
     StaleTarget { target: String, reason: String },
     /// A board was live earlier in this server process, but IPC is now gone;
     /// its saved file may be stale relative to lost editor state.
-    UnsafeFileFallback { path: String },
+    UnsafeFileFallback { path: String, reason: String },
     /// KiCad answered, and its open-document list could not be read as a
     /// complete set of comparable board identities — so whether it holds this
     /// board is unknown, and neither a live edit nor a file edit is safe.
@@ -230,7 +230,10 @@ mod tests {
                 target: "p".into(),
                 reason: "r".into(),
             },
-            ToolErrorKind::UnsafeFileFallback { path: "p".into() },
+            ToolErrorKind::UnsafeFileFallback {
+                path: "p".into(),
+                reason: "r".into(),
+            },
             ToolErrorKind::AmbiguousOpenBoard { path: "p".into() },
             ToolErrorKind::HandlerError { reason: "r".into() },
         ];
--- a/crates/konnect-core/src/mcp/handler.rs
+++ b/crates/konnect-core/src/mcp/handler.rs
@@ -368,22 +368,40 @@ impl McpHandler {
                     )
                 }
                 Err(e) if kicad_editor_locked_path(&e).is_some() => {
-                    let path = kicad_editor_locked_path(&e)
-                        .expect("guard matched")
-                        .display()
-                        .to_string();
-                    (
-                        CallToolResult::error_kind(
-                            ToolErrorKind::Conflict {
-                                paths: vec![path.clone()],
-                            },
-                            format!(
-                                "Schematic '{path}' has a KiCad editor lock. Close Eeschema, or resolve a stale lock only after confirming no editor owns the file, then retry."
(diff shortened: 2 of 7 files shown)


neusse:
Merge-order update: PR #487 should land before this lock-safety PR. #487 closes the two remaining exact board-identity gaps found after #480 merged; this PR relies on those identity gates when deciding whether file fallback is safe. Once #487 lands, I will rebase #486 and rerun its gate.

Issue #226, opened 2026-08-16
Datasheet written as "" where KiCad writes "~", and no Description on placed instances


mixelpixx:
Two differences between what `add_schematic_component` writes and what eeschema
writes, found by @triglav-modular while resolving #204. Filing them as reported:
**neither is confirmed to matter**, and the #204 exercise is a good reminder
that a difference from KiCad is not automatically a defect — the Footprint
half of that report turned out to be Konnect matching KiCad exactly.

Both confirmed present on `main`.

### 1. `Datasheet` is written empty, KiCad writes `~`

`crates/konnect-core/src/tools/sch_components.rs:630`:

```rust
.push(positioned("Datasheet", "", x, y, 0.0, true, centred));
```

KiCad writes `(property "Datasheet" "~" …)` on a placed symbol with no
datasheet — verbatim from
`<KiCad>/SharedSupport/template/RaspberryPi-HAT/RaspberryPi-HAT.kicad_sch`.

`~` is KiCad's "nothing here" sentinel in this position, so the question is
whether an empty string reads as *absent* or as *present and blank* to
eeschema, the BOM exporter, and the datasheet-link action. If they differ, ours
is the wrong one. If they don't, this is cosmetic and should be closed as such
rather than changed for symmetry.

### 2. No `Description` property on the placed instance

KiCad emits `(property "Description" "…" …)` on the placed symbol, copied from
the library symbol. Konnect writes no `Description` at all — `grep '"Description"'
sch_components.rs` returns nothing.

Worth knowing what consumes it before acting: if eeschema re-reads the
description from `lib_symbols` on load, its absence on the instance costs
nothing; if the BOM or a search reads the instance copy, a Konnect-authored
sheet has a blank column where an eeschema-authored one does not.

### How to settle both

The same way #204 was settled, which worked well: compare against KiCad's own
files rather than reasoning about them. The 19 template projects under
`<KiCad>/SharedSupport/template/` are eeschema-authored, and
`kicad-cli sch export bom` / `python-bom` will show whether either field
survives into an export.

If the answer is "no consumer cares", close this — a gratuitous change to what
we write is its own risk.

PR #503, opened 2026-09-08
Three placement/library fixes found by driving a real KiCad 10 project

asqude:
These are the three patches a downstream project still carries against 0.11.0
after the rebase absorbed the rest. All three were found by breaking real files
in an eight-sheet KiCad 10 avionics project, not by reading code, and each is
reproducible against a copy of that project.

## 1. Placement writes an empty Footprint and a Value taken from the symbol name

`add_schematic_component` / `batch_place_components` copy Datasheet and
Description onto the instance (#226) but hardcode an **empty Footprint** and
derive Value from the `lib_id`'s symbol name rather than the library symbol's
own `Value` field.

That looks harmless because the sheet still carries the correct definition in
its `lib_symbols` cache — but **`kicad-cli` does not fall back to that cache**.
It emits a bare `(field (name "Footprint"))`, so a part placed by stock 0.11.0
reaches no netlist footprint and no board. Verified by exporting a netlist, not
by inspection.

Fixed by resolving both fields through the `src` resolver already threaded
through the placement path.

## 2. Placement readback fails its own intent after fix 1

0.11.0 verifies every write by reading it back against an intent bound
beforehand. The intent computed Value with its own copy of the
name-derived expression, so fix 1 made **every placement fail its own
readback** with `post-write property Value differs from bound intent`.

Both sides now call `resolved_placement_value()`, so the two cannot drift
again.

## 3. `prune_unused_lib_symbols`

Not present upstream. Orphaned `lib_symbols` cache entries accumulate whenever
a symbol is renamed or swapped — eeschema never prunes them and `kicad-cli`
preserves them verbatim. On one sheet here that was ~1900 dead lines in a
7600-line file.

Placed in the **`sch_batch`** toolset rather than `sch_components`, because
0.11.0 caps a toolset at 20 tools and `sch_components` is already at 20. It
fits better there anyway: it is a bulk delete of cached entries. The operation
is netlist-neutral, which is its acceptance test.

## Testing

`konnect-core`: 1031 passed, 0 failed.

Fixes 1 and 2 were additionally verified end to end by placing parts into a
copy of the downstream project and exporting a `kicad-cli` netlist — the check
that exposed the original bug, since the in-file `lib_symbols` cache hides it.

Happy to split these into separate PRs, rework the toolset placement, or add
tests in whatever style you prefer.

### DIFF
--- a/crates/konnect-core/src/router/registry.rs
+++ b/crates/konnect-core/src/router/registry.rs
@@ -28,7 +28,7 @@ pub static ALL_TOOLSETS: &[ToolsetMeta] = &[
         name: "sch_components",
         description: "Add, edit, move, rotate, and delete schematic symbols, and set the page size",
         category: "schematic",
-        tool_count: 20,
+        tool_count: 21,
     },
     ToolsetMeta {
         name: "sch_wiring",
--- a/crates/konnect-core/src/router/registry.rs
+++ b/crates/konnect-core/src/router/registry.rs
@@ -28,7 +28,7 @@ pub static ALL_TOOLSETS: &[ToolsetMeta] = &[
         name: "sch_components",
         description: "Add, edit, move, rotate, and delete schematic symbols, and set the page size",
         category: "schematic",
-        tool_count: 21,
+        tool_count: 20,
     },
     ToolsetMeta {
         name: "sch_wiring",
@@ -52,7 +52,7 @@ pub static ALL_TOOLSETS: &[ToolsetMeta] = &[
         name: "sch_batch",
         description: "Bulk add, edit, delete, and move schematic elements in one call",
         category: "schematic",
-        tool_count: 12,
+        tool_count: 13,
     },
     ToolsetMeta {
         name: "sch_export",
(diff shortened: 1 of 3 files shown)


mixelpixx:
@asqude — thank you for this; closing it did not make it go away. I reproduced all three claims against current `main` (`1cd4978`) with a release binary and `kicad-cli`, and every one holds. Filed so they can be claimed and landed as independent PRs under the queue rules:

- **#506** — placement writes `Footprint ""` and a name-derived `Value` while the library symbol's own values already sit in the sheet's `lib_symbols` cache (`Regulator_Linear:AMS1117-3.3` → cache says `Package_TO_SOT_SMD:SOT-223-3_TabPin2`, instance says `""`; kicad-cli's netlist then has no `(footprint …)` node). Your second finding — the bound intent re-deriving `Value` with its own copy of the expression — is recorded there as the reason the fix must go through one shared function.
- **#507** — the part that made it worse than you reported: `update_pcb_from_schematic` does `required_value(component, "footprint")` on the netlist, so one footprint-less symbol fails the *entire* sync rather than being reported.
- **#508** — `prune_unused_lib_symbols`, as a `sch_batch` feature with a netlist-neutral acceptance test.

If you want to carry any of them, assign yourself on the issue and open one focused PR per issue from current `upstream/main` (docs/BRANCH_AND_PULL_REQUEST_WORKFLOW.md has the shape). Your authorship on the original diff is noted on #506.

Issue #452, opened 2026-09-07
update_pcb_from_schematic: two footprints with no schematic path collide on identity "/" and block every sync


triglav-modular:
## Summary

A footprint with no `(path …)` — that is, any footprint placed directly on the board and not present in the schematic — is normalised to the schematic identity `/`. Two of them therefore look like duplicates **of each other**, `update_pcb_from_schematic` raises `duplicate_board_identity`, and the conflict blocks the whole sync.

Board-only footprints never have a path. So **two logos, or two fiducials, or two mounting holes, are enough to make every future sync of that board impossible** — including a sync that would otherwise have been a no-op.

Renaming them does not help. This is the part that surprised us: the duplicate-reference diagnostic is only the first of two, and clearing it leaves the blocking one behind.

## Reproduction

v0.11.0, clean fixture, KiCad 10.0, pcbnew open on the board (the tool is live-IPC-only). Board carries `R1` and `R2` from the schematic; the schematic is untouched throughout, so the correct answer in every row below is `noop`.

| board-only footprints | dry run | diagnostics |
|---|---|---|
| one, `REF**` | `noop` | none — `board_only_preserved {planned: 1}` |
| two, both `REF**` | **conflict** | `duplicate_board_reference` **and** `duplicate_board_identity` |
| two, renamed `LOGO1` / `LOGO2` | **conflict** | `duplicate_board_identity` alone |

The blocking diagnostic reads:

```json
{"code": "duplicate_board_identity",
 "message": "board contains duplicate schematic identity /",
 "reference": "LOGO1"}
```

Apply is refused in both conflict rows, with every counter `applied: 0`.

Verified against the file rather than the return value: `R1` and `R2` each carry a `(path "/<uuid>")` line; **neither board-only footprint has a `(path …)` line at all.**

## The response contradicts itself

In the same payload that calls them a conflict, the tool already classifies them correctly:

```json
"board_only_preserved": {"applied": 0, "planned": 2},
"conflicts":            {"applied": 0, "planned": 2}
```

The preservation logic understands these are board-only items to leave alone. Only the identity grouping treats them as duplicates.

## Workaround, and why it is not a fix

Giving each pathless footprint a unique synthetic path from inside pcbnew —

```python
fp.SetPath(pcbnew.KIID_PATH("/" + str(uuid.uuid4())))
```

[...]

PR #504, opened 2026-09-08
fix(pcb): read a pathless board footprint as having no identity

triglav-modular:
## Summary

Part of #452 — the **identity** half. The `REF**` reference half is not in this diff (see "Not in this PR"), so merging this will not close the issue while half its Ask is outstanding.

A footprint placed directly on the board — logo, fiducial, mounting hole — has no schematic symbol. KiCad's IPC layer reports that as a *present but empty* `SheetPath`, and `snapshot_board` rendered it through `sheet_path_string`, which turns an empty path into `/`. Every such footprint therefore arrived carrying the same synthetic identity, `plan_sync` read two of them as duplicates, and `duplicate_board_identity` blocked the sync permanently — `update_pcb_from_schematic` returns before apply on a conflict and takes no override. Renaming does not help: the collision is on identity, not reference.

## Approach

The planner was already right. `symbol_path` is already `Option<String>`, the duplicate-identity grouping is already guarded by `if let Some(path)`, and three sites already read `None` as "board-only". **The reader was turning absence into presence.**

`board_symbol_path` maps a `SheetPath` with no non-empty KIID to `None`. A real path is unchanged. An empty segment *inside* a real path is deliberately left alone — KiCad does not emit that shape, and inventing a meaning would be guessing.

**Extraction for testability.** The defect lived in the IPC-to-planner conversion, unreachable from a test: `snapshot_board` takes a live `KiCadIpcClient`, and the planner's own tests build `BoardFootprint` directly, so a regression test written that way passes `None` straight in and **passes with the bug in place**. It is now `board_footprint_from_instance(&FootprintInstance) -> Result<BoardFootprint>`. The move is otherwise byte-faithful — the only changed lines are the guard and the `Ok(…)` wrapper.

### Behaviour change beyond the reported defect

Two guards test `symbol_path.is_none()`, so while every board-only footprint wore `/` they were unreachable for exactly the footprints this issue is about. Reading absence correctly wakes both, on boards that never hit the duplicate-identity bug:

| scenario | before | after |
|---|---|---|
| schematic names `R1`; board has an unlinked footprint named `R1` | `conflict` / `reference_identity_conflict`, nothing written | `ready`, one update — adopted, identity written onto it |
| schematic gains a component matching an unlinked footprint's library id **and** value | `ready`, one add | `conflict` / `reference_only_rename_ambiguous` |
| `on_board=no` instance already on the board | `conflict` | unchanged |

**Row two means this fix can turn a `ready` plan into a `conflict`** — the opposite direction from its purpose. I believe it is the better behaviour: the old path silently added a second identical footprint beside the unlinked one, and this is the `possible_renames` scan finally working. But it belongs in the release notes.

The captured fixture sharpens this. KiCad leaves `not_in_schematic` **false** on a board-only mounting hole (marking it `exclude_from_position_files` and `exclude_from_bill_of_materials` instead), and that flag is the precondition for row two — so ordinary board graphics are *exposed* to this, not exempt.

### Not in this PR

The `REF**` half — whether a default library reference is an identity — is a separate claim, agreed on the issue as a follow-up. **With this PR alone, two footprints still named `REF**` continue to raise `duplicate_board_reference`**, so the reported board needs its graphics renamed before it syncs. Renamed board-only footprints now plan `noop` with `board_only_preserved`, which was impossible before at any name.

## Branch and dependencies

Base current `main` (`d4d1dda`); no dependencies; four commits, all unique to this PR. `pcb_sync.rs` is touched by no other open PR.

## Compatibility and safety

No public API, schema, tool or argument change; no tool added or removed, so counts are untouched. Behavioural impact only, in the direction the fields already documented: a board-only footprint is now counted in `board_only_preserved` and no longer produces a spurious `duplicate_board_identity`. `symbol_path` was already `Option`; this makes the value honest. No file or IPC mutation is added — the changed code builds the plan; the apply path is untouched.

## Validation

```
cargo fmt --all -- --check                                    exit 0
cargo test --workspace --locked --lib --tests                 exit 0
cargo test --workspace --locked --doc                         exit 0
cargo clippy --workspace --locked --all-targets -- -D warnings exit 0
```

**Fixture provenance is real KiCad output.** `crates/konnect-core/tests/fixtures/board_only_footprint.ipc.bin` is a `FootprintInstance` captured over IPC from KiCad 10.0.6 for a mounting hole on this repo's `live_ipc.kicad_pcb`; README beside it records provenance, the redaction check and regeneration. Regeneration runs through the ignored live test that asserts the same property it writes, so a capture changing the answer fails rather than silently replacing the fixture.

**Live observation, not deferred.** All four of the template's pathless mounting holes report `Some(SheetPath { path: [] })` — present and empty, `path_human_readable` empty rather than `/`. Checked in as `kicad_reports_an_empty_sheet_path_for_a_board_only_footprint`. The capture also corrected the fixture on first run (`assertion failed: logo.not_in_schematic`), which is the argument for it: the hand-written fixture and its test shared one wrong belief about KiCad.

**Guard neutering.** Restoring `footprint.symbol_path.as_ref().map(sheet_path_string)` fails all four guard tests, each showing the pre-fix value:

```
board_only_footprint_reads_as_no_identity   Some("/") != None
two_board_only_footprints_do_not_collide    1 duplicate_board_identity != 0
an_unlinked_footprint_..._is_adopted        Conflict  != Ready
an_unlinked_lookalike_blocks_...            Ready     != Conflict
```

A fifth test — a real two-KIID path still reads `/sheet-uuid/symbol-uuid` — passes either way by design: it proves the guard is not too broad.

## Review checklist

- [x] Focused diff; no generated output, personal data, or unrelated cleanup.
- [x] Based on current `main`, no conflicts, CI green on this exact head.
- [x] Only this PR's unique commits; no dependencies.
- [x] Names follow `docs/NAMING_CONVENTIONS.md`; no public rename.
- [x] New behaviour and failure paths have regression coverage.
- [x] File mutations: none added. IPC mutations: none added.
- [x] No tools added or removed; counts untouched.


neusse:
Reviewed exact head `07e034c067d175503c8810ce76912b13b72f5d07` against current `main` and issue #452.

**Please replace or prove the provenance of the synthetic IPC fixtures before this enters the merge queue.** `board_only_instance` hand-constructs the exact `FootprintInstance` shape whose interpretation is under test. CONTRIBUTING requires board/footprint fixtures to come from real KiCad output. Use a checked-in, redacted protobuf capture from KiCad IPC (or a demonstrably KiCad-generated fixture) and decode that in the regression test. The fixed build live observation can remain explicit validation debt if unavailable; the deterministic fixture itself still needs real-KiCad provenance.

The declared partial scope and closure accounting are correct: this is `Part of #452`, implements the empty-path identity half, and does not claim to close the still-blocking `REF**` reference half.

One behavior expansion needs to stay prominent in tests and release notes: restoring `None` activates reference adoption and lookalike-ambiguity branches, so some plans change beyond merely suppressing `duplicate_board_identity`. I do not see an implementation error there—the behavior follows the intended absence invariant—but it is broader than the issue primary reproduction.

Once the fixture provenance is corrected and all checks pass on the new exact head, this can return for final review.

triglav-modular:
Thanks — the issue decision answers both my questions. Follow-up as specified: separate PR after this one, narrow rule, no warnings array, `Part of #452`, both acceptance cases proven.

**Fixture provenance: done, and you were right in a way I'd have missed.** Commits `dad4e7a`, `3e3e081`, `835fbe3`.

The tests now decode `crates/konnect-core/tests/fixtures/board_only_footprint.ipc.bin` — a real `FootprintInstance` captured over IPC from KiCad 10.0.6, for a mounting hole on this repo's own `live_ipc.kicad_pcb` that carries no `(path ...)` line. Provenance, redaction check and regeneration are in the README beside it.

It corrected the fixture on its first run:

```
assertion failed: logo.not_in_schematic
```

KiCad leaves `not_in_schematic` **false** on a board-only footprint, marking it `exclude_from_position_files` and `exclude_from_bill_of_materials` instead. My hand-written fixture set it true and the test asserted it — fixture and test sharing one wrong belief, which is the failure your rule describes. I would not have found it by argument.

It also changes the PR's own risk note: that flag gates the rename-ambiguity branch, so ordinary board graphics are *exposed* to the `ready`→`conflict` transition, not exempt. The body's behaviour section is updated.

The other assumption held — `symbol_path` arrives present and empty (`path: []`, `path_human_readable: ""`), on all four of the template's mounting holes.

Regeneration is wired to verification: `KONNECT_CAPTURE_IPC_FIXTURE=1` on the ignored live test `kicad_reports_an_empty_sheet_path_for_a_board_only_footprint`, which asserts the same property it writes, so a capture that changes the answer fails instead of silently replacing the fixture. Neutered check: restoring the old reader fails all four #452 tests, now against real KiCad bytes.

**Behaviour expansion** is in the body's "Behaviour change beyond the reported defect" section and pinned by `issue_452_an_unlinked_footprint_the_schematic_names_is_adopted_not_refused` and `issue_452_an_unlinked_lookalike_blocks_a_new_component_instead_of_duplicating_it`. Say the word if you want it somewhere more prominent.

Two notes:

- The capture sits in `konnect-core/tests/fixtures/` beside the crate that decodes it. Move it if you'd rather — one `include_bytes!` depends on the path.
- `live_kicad_test.rs`'s header recommends `ls -t /tmp/kicad/api-*.sock | head -1`. That picked a stale socket newer by mtime than the live one and gave "Connection refused"; matching `api-<pid>.sock` to a running `pcbnew` is reliable. Happy to fix that comment here or separately.

Last thing, not blocking: `docs/TESTING_AND_RELEASE.md` still routes IPC paths to unit tests plus an ignored live test, reserving real-KiCad fixtures for "formats KiCad owns". This PR does both, so nothing hangs on it — but that line and CONTRIBUTING's bullet disagree, and the next contributor hits the same fork.

neusse (approved):
Reviewed exact head \835fbe3770031e332aa2a07ca9cfc98f34074ff7\ after the fixture-provenance correction. The regression path now decodes a redacted KiCad 10.0.6 IPC capture, documents regeneration, and includes a live observation plus red-before-green guard evidence. The broader adoption and lookalike-conflict consequences are disclosed and pinned by tests. All ten required checks pass on this head. Closure accounting is correct: this remains \Part of #452\; the separate REF** reference half remains outstanding. Ready for the maintainer merge gate.

PR #505, opened 2026-09-08
fix(platform): discover KiCad's socket from metadata, never by connecting

### LINKED ISSUE #498: IPC socket auto-detection wedges KiCad's API server on Linux (no reply to any client until restart)
## Summary

On Linux, when `KICAD_API_SOCKET` is unset, Konnect's IPC address auto-detection appears to leave KiCad's API server permanently stuck. The Ping then fails with `NNG recv failed: Timed out`, and **from that moment on KiCad stops answering every IPC client**, including KiCad's own `kicad-python` (kipy). Only restarting the editor recovers it.

Passing the address explicitly via `KICAD_API_SOCKET` (which skips auto-detection) makes everything work.

## Environment

- Konnect v0.11.1, built from source (`cargo build --release -p konnect`), rustc 1.96.0
- KiCad 10.0.6 (`10.0.6~ubuntu22.04.1`, KiCad PPA), system `libnng1` 1.5.2
- Ubuntu 22.04.5, x86_64
- Board open in standalone `pcbnew`; `api.enable_server = true`

## Reproduction

1. Start `pcbnew <board>` with the API enabled. Confirm `/tmp/kicad/api.sock` exists.
2. Verify KiCad answers, using KiCad's own client:
   ```python
   from kipy import KiCad
   KiCad(socket_path="ipc:///tmp/kicad/api.sock", client_name="probe").ping()   # OK
   ```
   This succeeds repeatedly from separate processes.
3. Run Konnect **without** `KICAD_API_SOCKET` and call `open_project`:
   ```
   WARN konnect::config: KiCad IPC address from auto-detection: ipc:///tmp/kicad/api.sock
   WARN konnect_ipc::client: [BETA] Ping to ipc:///tmp/kicad/api.sock failed: NNG recv failed: Timed out
   ```
   → `ipc_available: false`
4. Re-run step 2. It now **also** times out. KiCad is wedged until `pcbnew` restarts.

With `KICAD_API_SOCKET=ipc:///tmp/kicad/api.sock` set, step 3 instead returns:

```
ipc_available : true
kicad_ui      : true
open_boards   : ['.../StickHub.kicad_pcb']
```

## What I ruled out

- **Not the request payload.** I captured Konnect's bytes with a pynng REP server and replayed them verbatim from pynng to KiCad — KiCad answered normally, twice.
  ```
  kipy    : 0a07120570726f6265 12300a2e747970652e676f6f676c65617069732e636f6d2f6b696170692e636f6d6d6f6e2e636f6d6d616e64732e50696e67
  konnect : 0a0f120d6b6f6e6e6563742d3930313437 1230...(identical Any)
  ```
  Structurally identical; only `client_name` differs.
- **Not concurrency.** A REP probe server counted exactly **one** request from `open_project`'s ping path.
- **Not the bundled nng version.** Konnect statically links nng 1.4.0-rc.0 (via `nng-sys 1.4.0-rc.0`) while KiCad uses 1.5.2 and pynng 1.11.0. I rebuilt Konnect against `nng-sys 0.3.0+1.11.0` — **identical failure**, so this is not a version-pairing problem.
- **A raw stream proxy makes it work.** Putting a plain `AF_UNIX` relay between Konnect and `/tmp/kicad/api.sock` (which also means the address arrives via `KICAD_API_SOCKET` rather than auto-detection) produced correct replies for all three requests:
  ```
  C->K    8B 0053500000300000      # nng REQ0 handshake
  K->C    8B 0053500000310000      # nng REP0 handshake
  C->K   80B ...konnect-97680...   # Ping
  K->C  102B ...                   # reply
  ```
  The wire bytes are the same as the direct case that hangs.

That combination points at the auto-detection path itself rather than at `send_command`. `crates/konnect-ipc` carries a "bounded probing of KiCad's unix socket, and the euid it must be owned by" (socket2/libc) — a bare `connect()` that opens and drops a connection without completing nng's handshake looks like a plausible way to block KiCad's single-threaded REP loop, though I did not confirm that specific line.

## Impact

Out of the box on Linux (no `KICAD_API_SOCKET` in the environment, which is the normal case when the MCP server is launched by an AI client rather than by KiCad), every live-PCB tool fails **and** the user's KiCad session is left unusable for any API client until restarted. The failure is silent from KiCad's side — no error dialog, no log.

## Suggested fix

Make the probe non-destructive: either detect the socket by path/permission checks only (`stat`, ownership) without connecting, or perform the liveness check with a real nng REQ0 `Ping` and a short timeout so the handshake always completes.

## Workaround

Set `KICAD_API_SOCKET=ipc:///tmp/kicad/api.sock` in the MCP server environment (or `ipc_address` in `settings.json`).

tonydzi:
I am an AI agent (Claude), working with disclosure on Anton Dzyatkovsky's behalf.

Fixes #498, on the implementation order in the triage comment.

## The probe was the wedge

`socket::is_listening` decided liveness by dialling: it opened a raw `AF_UNIX` stream and took `connect_timeout` success as proof. Against KiCad's NNG `REP` endpoint that is a connection which never completes NNG's handshake, and afterwards KiCad answers nobody — its own `kipy` included — until `pcbnew` restarts.

The reporter measured that from outside the process. Here it is from this side of the socket, red on `d4d1dda` before the change: a candidate is bound by a listener the test owns, discovery runs, and the listener's accept queue is inspected.

```
thread 'socket::tests::detection_never_connects_to_a_candidate' panicked at
crates/konnect-ipc/src/socket.rs:333:22:
detection opened a stream connection to the candidate; against KiCad's NNG endpoint
that is the wedge in #498
```

That assertion is the whole fix expressed as a test: it fails for exactly one reason, and no rewording of the probe can satisfy it except not connecting.

## What changed

Discovery now answers from metadata: the path exists, it is a socket rather than some other file, and its owner is this euid. Nothing is dialled. `is_listening` is gone; `is_adoptable` is what replaces it, named for what it can actually prove.

Point by point against your list:

1. Existence, socket type and same-euid ownership only — `is_adoptable_owned_by` does three metadata reads and returns.
2. No second handshake is added. The bounded NNG `Ping` through `KiCadClient` remains the only liveness check, and the only code in this crate that speaks to the endpoint.
3. Ordering and precedence are untouched. `candidates_in` still yields the temp-dir socket first with the macOS `/tmp` fallback second, `Config::resolve_ipc_address_with` still takes `KICAD_API_SOCKET` before detection, and `discovery_picks_the_first_adoptable_candidate` plus the existing config tests pin both.
4. The regression test is `detection_never_connects_to_a_candidate`, asserting from the listener's side rather than by inspecting our own code. The live Linux confirmation is not mine to give — see below.
5. Stale-socket behavior is documented where the address is resolved, on `detect_ipc_address` and on `is_adoptable`, and asserted by a test rather than left in prose.

## The cost, named

Metadata cannot distinguish a socket KiCad left behind from one it is still serving, so discovery will now adopt a stale `api.sock`. `a_socket_left_behind_by_a_closed_kicad_is_still_adopted` asserts that on purpose; it is the inverted twin of the test that used to assert the opposite.

What the user sees in that case is the `Ping` failing within its own bound and the session falling back to file editing — the same outcome as a KiCad that was never running, reached without touching the editor. Rejecting the stale file would cost a connect, and the connect is the bug.

## Mutations killed

Each mutation was applied to the merged state and the suite re-run; each is caught by one named test, so the tests are pinned to mechanisms rather than to a shared marker.

```
drop the socket-type check (existence only)  -> an_owned_regular_file_is_not_adopted FAILED
drop the ownership gate                      -> a_live_socket_that_fails_the_ownership_check_is_not_adopted FAILED
restore the connect probe (the bug itself)   -> detection_never_connects_to_a_candidate FAILED
                                                a_socket_left_behind_by_a_closed_kicad_is_still_adopted FAILED
baseline restored                            -> 11 passed
```

## Also removed

`PROBE_TIMEOUT` and `a_saturated_listener_gives_up_within_the_bound` go with the connect. That test guarded a real hazard — a blocking `connect()` on a saturated `AF_UNIX` backlog waits forever, before tracing is up — but the hazard existed only because discovery connected. Keeping the test would mean keeping the thing it guards.

`socket2` was pulled in for that probe alone and is now unused, so it leaves `konnect-ipc`'s Unix dependencies and `Cargo.lock`. `libc` stays, for `geteuid`.

## Compatibility and rollback

No public surface moves: `detect_ipc_address` and `candidate_socket_paths` keep their signatures and meaning, and the renamed function was private. The Windows arm still detects nothing, with the same reasoning; only its doc changes, because there is no filesystem entry to inspect rather than nothing to connect to. Config files, addresses and on-disk formats are untouched.

Blast radius is address resolution when `KICAD_API_SOCKET` is unset. Rollback is a plain revert of this branch — no migration, nothing written to disk to undo.

## Local gate

macOS 26.3.1 x86_64, rustc 1.96.0, on the head of this branch:

```
cargo test --workspace --locked --lib --tests                     1655 passed, 0 failed
cargo test -p konnect-ipc --lib socket::                          11 passed, 0 failed
```

## What I could not verify

I have no Linux machine with KiCad 10.0.6, so your point 4's live check — auto-detect, then repeated Konnect and `kipy` pings against a KiCad that stays responsive — is untested by me. It stays with the reporter, whose environment you already identified as the right place for it. I would rather say that plainly than let a green macOS gate imply a confirmation it does not carry.

The Windows named-pipe question is also untouched here: discovery still declines to guess NNG's pipe mapping, which is the pre-existing behavior and a separate change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2521,7 +2521,6 @@ dependencies = [
  "prost-types",
  "serde",
  "serde_json",
- "socket2",
  "tempfile",
  "thiserror 1.0.69",
  "tokio",
--- a/crates/konnect-ipc/Cargo.toml
+++ b/crates/konnect-ipc/Cargo.toml
@@ -20,9 +20,8 @@ nng.workspace = true
 prost.workspace = true
 prost-types.workspace = true
 
-# Bounded probing of KiCad's unix socket, and the euid it must be owned by.
+# The euid a discovered socket must be owned by.
 [target.'cfg(unix)'.dependencies]
-socket2 = "0.6"
 libc = "0.2"
 
 [build-dependencies]
--- a/crates/konnect-ipc/src/socket.rs
+++ b/crates/konnect-ipc/src/socket.rs
@@ -3,26 +3,15 @@
 //! KiCad exports `KICAD_API_SOCKET` only to plugins it launches itself, so a
 //! standalone server started by an MCP client sees nothing and every IPC call
 //! fails as unconfigured. KiCad's own default is predictable —
-//! `<temp dir>/kicad/api.sock` — so probe it before giving up.
+//! `<temp dir>/kicad/api.sock` — so look there before giving up.
+//!
+//! Looking is all this module does. It reads the filesystem's own metadata and
+//! never opens a connection: what sits at that path is KiCad's NNG endpoint,
+//! and dialling it outside the NNG protocol is what wedged the editor in #498.
+//! Whether the endpoint answers is decided later, by the bounded NNG `Ping` in
+//! [`crate::client`].
 
 use std::path::{Path, PathBuf};
-
-/// How long to wait for a candidate socket to accept the probe.
-///
-/// A live local listener accepts immediately, so this bounds only the
-/// pathological case: a KiCad that is still bound but has stopped accepting,
-/// whose backlog is full. A *blocking* `connect()` on an `AF_UNIX` stream
-/// waits for a slot there — it does not fail fast the way TCP does — and
-/// against a queue nothing ever drains it waits forever. This probe runs from
-/// `Config::load_resolved` before tracing is initialized, so that would hang
-/// the server before one line reached stderr.
-///
-/// `connect_timeout` is therefore what does the work: it connects
-/// non-blocking, and Linux answers a full backlog immediately (`POLLHUP`,
-/// measured at ~16 µs). The duration is the remaining bound, for a listener
-/// that neither accepts nor hangs up.
-#[cfg(unix)]
-const PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(250);
 
 /// Socket paths KiCad may have created on this platform, most likely first.
 pub fn candidate_socket_paths() -> Vec<PathBuf> {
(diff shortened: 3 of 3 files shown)


neusse (approved):
Reviewed exact head \7e2d787603406e6def19fed28f432d77696a55b3\ against current \main\ and #498. The diff is focused: it removes the raw AF_UNIX connect probe and its sole dependency, preserves candidate ordering and explicit-address precedence, and proves from the listener side that discovery never connects. All ten required checks pass on this head. The terminal \Fixes #498\ reference is correct. The reporter's KiCad 10.0.6/Linux repeat-ping observation remains explicit validation debt, not a merge wall: the deterministic guard directly removes the destructive operation, hosted Linux CI is green, and rollback is a plain revert. Ready for the maintainer merge gate.

PR #509, opened 2026-09-09
fix(pcb): scope duplicate_board_reference to references the schematic uses

### LINKED ISSUE #452: update_pcb_from_schematic: two footprints with no schematic path collide on identity "/" and block every sync

triglav-modular:
## Summary

Closes #452 — the **reference** half, and the last one outstanding. #504 landed the identity half as `c62bc5a`, and the agreement on the issue was that the follow-up stays `Part of` unless #504 had already merged and the follow-up satisfies the final acceptance item. It has and it does, both cases pinned by tests below. If you read the remaining scope differently, say so and I will downgrade to `Part of` rather than have the issue close early.

A footprint placed directly on the board keeps its library's reference, and for KiCad's stock libraries that is `REF**`. Two of them are therefore both `REF**`, which is normal. `plan_sync` called that `duplicate_board_reference`, and since any non-empty `diagnostics` makes the plan a `Conflict`, the whole sync was refused — the remaining reason the reported board would not sync without renaming its graphics first.

## Approach

A duplicate board reference needs an unambiguous adoption target **only when the schematic export actually names that reference**. `board_by_reference` is consulted at exactly four sites after it is built, and every one looks up a reference that came from the export — `skipped.reference` once, `component.reference` three times. A reference the export never names is never looked up, so which of two duplicates `insert` kept is unobservable. The condition becomes "duplicate **and** the reference is in `design.components` or `design.skipped`": one `HashSet`, one added clause. The map is still built from every footprint — only the diagnostic is scoped, since suppressing the `insert` would change which footprint an unrelated adoption resolves to.

**It keys on the export's reference set, not on the footprint's kind.** A blanket "board-only footprints cannot collide" exemption would look equivalent and be wrong: the reference-only adoption path matches a schematic component to a *pathless* board footprint by reference, so with two candidates it would pick whichever the map kept and write the schematic identity onto it. That case still conflicts.

**No `warnings` array.** `SyncDiagnostic` has no severity field and `plan_sync` treats any non-empty `diagnostics` as `Conflict`, so "duplicate, but harmless" would mean a new response field — a minor bump for what is otherwise a patch. Settled that way on the issue.

### Behaviour change

This only ever *removes* a diagnostic, so it can turn a `conflict` into `ready` or `noop` and never the reverse — the opposite direction from #504, which could newly block and said so. "It only removes a push" would not establish that alone: if `plan_sync` short-circuited on a non-empty `diagnostics`, suppressing this one would let later loops run for the first time and raise theirs. It does not. `diagnostics` is read only at the two `!diagnostics.is_empty()` checks at the end, after both loops have run, with no early return — so suppression removes one element from an otherwise identical set.

| scenario | before | after |
|---|---|---|
| two board-only `REF**` graphics, schematic names no `REF**`, otherwise in sync | `conflict`, nothing written | `noop`, both preserved |
| the same board with one component out of date | `conflict`, the update unreachable | `ready`, one update, both graphics preserved |
| two pathless footprints sharing a reference the schematic **does** use | `conflict` | unchanged |
| a reference used only by an `on_board=no` instance | `conflict` | unchanged |
| a duplicate pair where one carries a schematic path the export names | `conflict`, the identity match unreachable | `ready`, the path-matched one adopted, the other preserved |

The last row is the limit of "un-diagnosed footprints are left alone": they are never *selected by reference*, which is not the same as never written to. A footprint never back-annotated carries an identity and still wears `REF**`, and selection by path is a separate, unambiguous route the duplicate reference plays no part in.

## Branch and dependencies

Base current `main` (`296641b`); no dependencies; two commits, both unique to this PR. `pcb_sync.rs` is touched by no other open PR. Builds on #504, already merged and an ancestor of this base.

## Compatibility and safety

No public API, schema, tool, argument or response-shape change; no field or tool added or removed, so documented counts are untouched. One diagnostic is raised in a strictly narrower set of cases, in the direction of unblocking. No file or IPC mutation is added — the changed code builds the plan; the apply path is untouched.

## Validation

```
cargo fmt --all -- --check                                    exit 0
cargo test --workspace --locked --lib --tests                 exit 0
cargo test --workspace --locked --doc                         exit 0
cargo clippy --workspace --locked --all-targets -- -D warnings exit 0
```

All pre-existing `pcb_sync` tests pass unchanged, including `reference_only_possible_rename_is_a_conflict`.

**Fixture provenance is real KiCad output.** `board_only_shared_reference.kicad_pcb` was written by KiCad through the PCB editor's scripting API — `NewBoard`, `FootprintLoad` twice against the stock `MountingHole.pretty`, `SaveBoard` — and `REF**` is the string that library file already carries. `board_only_shared_reference.ipc.bin` is a `FootprintInstance` captured over IPC from KiCad 10.0.6 on that board; its README records provenance, redaction, regeneration, and the one field the tests vary. It is a new board rather than two more footprints on `live_ipc.kicad_pcb`, which is deliberately legacy format (`version 20241229`) and would be upgraded in place by any save from KiCad 10.

**On `ki_fp_filters`, since the fixture rule names it:** this board has none — and neither does `live_ipc.kicad_pcb`, the board this repository ships and the source of the capture accepted in #504. Across the tracked `.kicad_pcb` fixtures it appears only on boards whose footprints carry `(path ...)`: it comes from the schematic symbol a footprint was placed for, so its absence on a mounting hole is correct, not a sign of a synthetic file.

**Regeneration is wired to verification.** The ignored live test asserts what the capture encodes — shared reference `REF**`, IPC returning as many footprints under it as the file carries, each with a present-but-empty `SheetPath` — and rewrites it only under `KONNECT_CAPTURE_IPC_FIXTURE=1`, so a regeneration that changes the answer fails rather than silently replacing the fixture. No CI job opens KiCad, so the board's own side is checked without it by a test deliberately not `#[ignore]`d.

**Guard neutering, twice, because the guards protect different things.** Deleting the whole clause:

```
issue_452_repeated_ref_star_graphics_are_not_a_duplicate     1 duplicate_board_reference != 0
issue_452_repeated_ref_star_graphics_no_longer_block_update  Conflict != Ready
issue_452_a_path_matched_footprint_is_still_adopted...       Conflict != Ready
```

Dropping only the `skipped` half of the set:

```
issue_452_a_duplicate_reference_only_an_on_board_no_instance_uses_still_conflicts   not raised
```

**One test is a scope test and survives both by design.** `issue_452_a_duplicate_reference_the_schematic_does_use_is_still_a_conflict` asserts a diagnostic the unconditional version also raises, so neutering cannot fail it: it proves the narrowing did not go too far — the acceptance case about an ambiguous adoption target — and counting it as coverage would overstate the tally. Four guard tests, one scope test.

## Review checklist

- [x] Focused diff; no generated output, personal data, or unrelated cleanup.
- [x] Based on current `main`, no conflicts, CI green on this exact head.
- [x] Only this PR's unique commits; no dependencies.
- [x] Names follow `docs/NAMING_CONVENTIONS.md`; no public rename.
- [x] New behaviour and failure paths have regression coverage.
- [x] File mutations: none added. IPC mutations: none added.
- [x] No tools added or removed; counts untouched.


neusse (approved):
Reviewed exact head `27330474172ecea94f064e3514b6e9ad053aa897` against issue #452, the later maintainer decision, and current repository standards.

Spec: pass. The narrow exported-reference rule implements the final `REF**` half, preserves the ambiguous-adoption conflict for both ordinary and `on_board=no` references, and correctly avoids a response-schema change. The real KiCad board/IPC evidence and guard-neutering tests satisfy the agreed evidence contract. With #504 already merged, `Closes #452` is now correct terminal accounting.

Standards: no hard findings. The repeated pathless-reference extraction in the live/offline tests is a small non-blocking duplication smell; keeping the assertions explicit is acceptable here and does not justify delaying this P0 fix.

All ten required checks pass on this exact head. Ready for the maintainer merge gate.

PR #513, opened 2026-09-10
fix(schematic): resolve nets by identity, not by name

### LINKED ISSUE #510: net_at names a two-name net by HashMap order: the decoupling audit reports the other name as an undecoupled rail
Found while reviewing the shared connectivity model after #336 landed. Verified on `main` at `ab33781`.

## Summary

`NetGraph::net_at` answers "what net is at this point" by scanning `point_nets` for the first entry whose union-find root matches the query point's:

```rust
// crates/konnect-core/src/tools/sch_connectivity.rs:234
pub(crate) fn net_at(&mut self, x: f64, y: f64) -> Option<String> {
    let k = pt_key(x, y);
    self.ensure(k);
    let root = self.find(k);
    let labels: Vec<_> = self.point_nets.clone().into_iter().collect();
    for (lk, net) in labels {
        if self.find(lk) == root {
            return Some(net);
        }
    }
    None
}
```

`point_nets` is a `HashMap`, so "the first entry" is iteration order. When one electrical net carries **two names** — a rail named by a `+3V3` power symbol that also has a label on it, or a local label on a net that also carries a global one — the name that comes back is arbitrary, and it differs between processes.

`collect_power_nets` (`design_review.rs:1547`) meanwhile lists **both** names as rails, because both are labels on the sheet. `capacitor_nets` resolves each capacitor pin through `net_at`, so every cap on that one net lands under whichever name won the toss, and the other rail is reported as having no decoupling at all:

```rust
// design_review.rs:434
let power_nets = collect_power_nets(&labels);          // both names
let (cap_nets, bulk_cap_nets) = capacitor_nets(&mut graph, &placed);  // one name
…
if !cap_nets.contains(net.as_str()) {
    // severity: "error" — "Power rail '…' has no decoupling capacitors"
```

## Reproduction

#336's decoupling test sheet with one label added to the `+3V3` rail, three consecutive runs of the same binary:

```
run 1  error: Power rail '+3V3' has no decoupling capacitors
run 2  error: Power rail '+3V3' has no decoupling capacitors
run 3  error: Power rail 'VCC'  has no decoupling capacitors
```

C2 (10 µF) is on that net in all three. Within a run the answer is stable — `point_nets` is never mutated after seeding, so one graph gives one answer — which is why the existing tests do not catch it: it takes a second name on a net, and it moves between processes rather than within one.

## Blast radius

The false `error`-severity finding on a correctly decoupled rail is the visible symptom, but the arbitrary naming is `net_at`'s and every caller inherits it.

Same name-as-identity comparison, so same failure:

- `audit_decoupling` — `capacitor_nets` / `test_point_nets` vs `collect_power_nets` (above).
- `audit_connections` — `pull_up_nets` collects the nets a pull-up resistor reaches, then the I2C and reset checks ask whether the pin's `net_at` name is in that set. A rail with two names produces a false "has no pull-up resistor" warning by the same coin toss.

Report the arbitrary pick to the caller as the net's name:

- `sch_analysis`: `get_pin_connections`, `get_component_nets`, `trace_from_point`, `get_connected_items`
- `sch_export`: `export_netlist_summary`

## Fix

The audits compare nets by *name* where they mean *identity*. `power_nets` should be the graph's root for each rail rather than its label text, and `cap_nets` (and `pull_up_nets`, `test_point_nets`) the roots the pins reach — `points_on_net` already walks that relation in the other direction and can anchor it.

Failing that, or in addition, `net_at` should at least choose **deterministically**, following KiCad's own precedence: global label over power symbol over local label over hierarchical label, ties broken by name. (Corrected — the original text had power symbols ranked below local labels; see the comment below for the measurements, and for why root comparison also needs a name-merge pass in the graph.) That alone turns a run-to-run coin toss into one stable wrong-ish answer, and it is what the read-only tools above need regardless of what the audits do.

Worth taking with it: the `self.point_nets.clone()` in that loop copies every key *and* every net-name `String` on every single query, and #336 made `design_review` the heaviest caller in the tree.

A test needs two names on one net — a `+3V3` power symbol plus a `VCC` label wired together, one capacitor on it — and must assert on identity, not on which name is returned, or it will pass by luck the way today's do.

pauliuszaleckas:
Closes #510.

## The bug

One electrical net can carry several names — a rail named by a `+3V3` power symbol that also has a `VCC` label on it, a local label on a net that also carries a global one. `NetGraph::net_at` answered "what net is at this point" by returning the first `point_nets` entry whose root matched, which is `HashMap` iteration order. The name therefore moved between runs of the same binary.

The audits then compared nets by that name where they meant identity. `collect_power_nets` listed **both** names as rails while every capacitor resolved to one of them, so the other was reported as an `error`-severity rail with no decoupling — a false finding on a correctly decoupled rail, landing on a different name run to run:

```
run 1  error: Power rail '+3V3' has no decoupling capacitors
run 2  error: Power rail '+3V3' has no decoupling capacitors
run 3  error: Power rail 'VCC'  has no decoupling capacitors
```

`audit_connections` had it too, via `pull_up_nets`, and `get_connected_items` filtered labels by name.

## The fix

**Identity, from `root_at`.** The audits and `get_connected_items` compare graph roots. One rail per net rather than per name; the capacitor, bulk-capacitor, test-point and pull-up sets are the roots pins reach.

**A root had to mean what KiCad means by a net first.** `seed_net_graph` joined points by geometry alone — wires, junctions, and labels lying on them — so two segments each carrying a `SIG` label were two roots, and every `GND` symbol on a sheet was its own rail. Nothing had depended on that: `points_on_net` pools by name, and `find_single_pin_nets` pools the roots a name reaches. Comparing roots as identity is what makes it matter, and without the merge it breaks the ordinary sheet — a decoupling capacitor is normally drawn on its own stub with its own `+3V3` symbol and shares no wire with the pin it decouples. `merge_named_nets` unions the points that share a name, after the geometry and before the naming.

**`name_of_root` gives KiCad's name,** not the map's first answer. The precedence was measured against KiCad 10.0.5 by netlisting one net carrying each pair, not assumed:

| net carries | KiCad calls it |
|---|---|
| `+3V3` power symbol + `LOCAL_VCC` label | `+3V3` |
| `+3V3` power symbol + `GLOBAL_SYS` global label | `GLOBAL_SYS` |
| `LOCAL_VCC` label + `GLOBAL_SYS` global label | `GLOBAL_SYS` |
| `LOCAL_VCC` label + `HIER_H` hierarchical label | `LOCAL_VCC` |
| `LOCAL_VCC` and `ZZZ` labels; `LOCAL_VCC` and `AAA` | `LOCAL_VCC`; `AAA` |

Global > power symbol > local > hierarchical, ties on the name ascending — `CONNECTION_SUBGRAPH`'s driver priority. A power symbol outranking a local label is why a label cannot rename a rail.

Names now resolve per root once seeding finishes, so `net_at`'s per-query clone of every key and every net-name `String` is gone. Every `NetGraph` mutator is private, so the resolved names cannot go stale behind a later union.

## Also fixed on the way

- **The ground skips in `audit_power_rails` read only the winning name.** A `GND` net that also carries a global label — globals outrank power symbols — escaped the skip and was reported as an undecoupled power rail. A `PowerRail` now carries every name on its net and the skips read all of them.
- **`get_connected_items` keyed everything by name.** On a two-name net it dropped the labels spelling the net its other way, and a net with **no** label contributed no points at all, so the wires and components on it went unreported. It collects roots, filters labels through `label_roots`, and takes points from one `points_on_roots` walk instead of one per name. `nets` still reports names, now sorted rather than in `HashSet` order.

## Behaviour changes visible to callers

- `audit_power_rails`' `power_nets` lists one entry per net instead of one per name, under the name KiCad would print; the `summary` count follows.
- `get_connected_items`' `nets` array is sorted and deduplicated.
- Tools reading `net_at` — `get_pin_net`, `get_component_nets`, `trace_from_point`, `export_netlist_summary` — return a stable name where the answer previously varied between processes.

## Verification

Nine new tests, each confirmed to fail against the code they guard:

- precedence for all five pairs above, twenty graph builds per case so an order-dependent answer cannot pass by luck
- both names of a net share one root
- two segments of one name are one net; two `+3V3` power symbols are one rail
- a capacitor on its own rail stub decouples the pin (1/1 power pins pass, zero findings)
- a rail named twice is audited as one rail, with no duplicate findings
- a pull-up on another segment of the net is found, so a fitted pull-up is not reported missing
- a component on another segment of the same net is reported as connected
- every label on a net is reported, and an unnamed net still reports its wire

Full local gate on this branch against current `main`: `cargo test --workspace --locked --lib --tests`, `--doc`, `cargo clippy --workspace --locked -- -D warnings`, `cargo fmt --all -- --check`, plus `crates/schematic-viewer`'s own tests, since `konnect-sexp` is touched (`LabelKind` gains `Copy`/`Eq`).

## Compatibility

`docs/API_MIGRATIONS.md` carries an entry for this under **Unreleased: schematic nets resolve by identity (minor release)**. Two response shapes change meaning — `audit_power_rails.power_nets` (one entry per net rather than per name) and `get_connected_items.nets` (sorted and deduplicated, with labels and unnamed-net items no longer dropped) — and findings move in the direction of fewer false positives. No tool, argument, or response field is renamed or removed.

`find_shorted_nets`' description is corrected in the tool schema and `tool-directory.md`: with same-named segments joined, two names can be netted together through a name rather than only through a wire path they share.

## Rollback

`git revert` of the single commit. It restores the name-based comparisons and the nondeterministic `net_at`; nothing persisted needs undoing and no caller state migrates either way.

## Known limitation

`find_shorted_nets` still reports two names on one net as a short, which on a rail named by a power symbol and labelled for readability is a false positive. That tool is a deliberate "two names met" report rather than a naming question, so it is left as it is.

### DIFF
--- a/crates/konnect-core/src/tools/design_review.rs
+++ b/crates/konnect-core/src/tools/design_review.rs
@@ -8,7 +8,7 @@
 use super::cli;
 use crate::mcp::protocol::CallToolResult;
 use crate::tool;
-use crate::tools::sch_connectivity::{net_graph_for, NetGraph};
+use crate::tools::sch_connectivity::{net_graph_for, NetGraph, NetRoot};
 use crate::tools::{
     get_path, invalid_arg, placed_pins_by_reference, project_name_for, sch_hierarchy, ToolContext,
     ToolDef,
@@ -23,7 +23,7 @@ use konnect_sexp::{
     },
 };
 use serde_json::{json, Value};
-use std::collections::HashSet;
+use std::collections::{BTreeSet, HashSet};
 use std::path::{Path, PathBuf};
 use tracing::info;
 
@@ -245,7 +245,7 @@ async fn handle_audit_decoupling_file(
     let mut total_power_pins = 0;
 
     // Collect all capacitor references and their net connections
-    let (cap_nets, _) = capacitor_nets(&mut graph, &placed);
+    let (cap_roots, _) = capacitor_nets(&mut graph, &placed);
 
     // For each IC (non-passive, non-connector component), check power pins
     for (inst, pins) in &placed {
@@ -273,17 +273,12 @@ async fn handle_audit_decoupling_file(
             let (px, py) = pin_endpoint(pin, *transform);
 
             // Check if there's a capacitor connected to a net that this pin is on
-            let pin_net = graph.net_at(px, py);
+            let pin_root = graph.root_at(px, py);
 
-            let has_decoupling = if let Some(ref net) = pin_net {
-                cap_nets.contains(net)
-            } else {
-                false
-            };
-
-            if has_decoupling {
+            if cap_roots.contains(&pin_root) {
                 pass_count += 1;
             } else {
+                let pin_net = graph.name_of_root(pin_root);
                 findings.push(AuditFinding {
                     severity: "error",
                     category: "decoupling",
@@ -337,7 +332,7 @@ async fn handle_audit_connections_file(
     let wires = extract_wires(&tree);
     let labels = extract_all_net_labels(&tree);
     let mut graph = net_graph_for(&tree, &wires, &labels);
-    let pull_up_nets = pull_up_nets(&mut graph, &placed);
+    let pull_up_roots = pull_up_nets(&mut graph, &placed);
 
     let mut findings = Vec::new();
 
@@ -352,9 +347,10 @@ async fn handle_audit_connections_file(
             for (pin_name, placed_pin) in [("SDA", named("SDA")), ("SCL", named("SCL"))] {
                 if let Some((pin, transform)) = placed_pin {
                     let (px, py) = pin_endpoint(pin, *transform);
-                    let net = graph.net_at(px, py);
+                    let root = graph.root_at(px, py);
+                    let net = graph.name_of_root(root);
                     if let Some(ref net_name) = net {
-                        if !pull_up_nets.contains(net_name) {
+                        if !pull_up_roots.contains(&root) {
                             findings.push(AuditFinding {
                                 severity: "warning",
                                 category: "connection",
@@ -381,9 +377,10 @@ async fn handle_audit_connections_file(
                 && !name_upper.contains("OUT")
             {
                 let (px, py) = pin_endpoint(pin, *transform);
-                let net = graph.net_at(px, py);
+                let root = graph.root_at(px, py);
+                let net = graph.name_of_root(root);
                 if let Some(ref net_name) = net {
-                    if !pull_up_nets.contains(net_name) {
+                    if !pull_up_roots.contains(&root) {
                         findings.push(AuditFinding {
                             severity: "warning",
                             category: "connection",
@@ -431,17 +428,20 @@ async fn handle_audit_power_rails_file(
     let mut findings = Vec::new();
 
     // Find all power nets (from power symbols and labels)
-    let power_nets = collect_power_nets(&labels);
+    let rails = power_rails(&mut graph, &labels);
 
     // Check each power net for bulk capacitance
-    let (cap_nets, bulk_cap_nets) = capacitor_nets(&mut graph, &placed);
+    let (cap_roots, bulk_cap_roots) = capacitor_nets(&mut graph, &placed);
 
-    for net in &power_nets {
-        if net.to_uppercase().contains("GND") || net.to_uppercase().contains("VSS") {
+    for rail in &rails {
+        let PowerRail {
+            root, name: net, ..
+        } = rail;
+        if rail.any_name_contains("GND") || rail.any_name_contains("VSS") {
             continue; // Ground nets don't need caps
         }
 
-        if !cap_nets.contains(net.as_str()) {
+        if !cap_roots.contains(root) {
             findings.push(AuditFinding {
                 severity: "error",
                 category: "power",
@@ -449,7 +449,7 @@ async fn handle_audit_power_rails_file(
                 issue: format!("Power rail '{}' has no decoupling capacitors", net),
                 recommendation: format!("Add at least one 100nF ceramic cap on the '{}' rail", net),
             });
-        } else if !bulk_cap_nets.contains(net.as_str()) {
+        } else if !bulk_cap_roots.contains(root) {
             findings.push(AuditFinding {
                 severity: "warning",
                 category: "power",
@@ -464,12 +464,15 @@ async fn handle_audit_power_rails_file(
     }
 
     // Check for test points on power rails
-    let test_point_nets = test_point_nets(&mut graph, &placed);
-    for net in &power_nets {
-        if net.to_uppercase().contains("GND") {
+    let test_point_roots = test_point_nets(&mut graph, &placed);
+    for rail in &rails {
+        let PowerRail {
+            root, name: net, ..
+        } = rail;
+        if rail.any_name_contains("GND") {
             continue;
         }
-        if !test_point_nets.contains(net.as_str()) {
+        if !test_point_roots.contains(root) {
             findings.push(AuditFinding {
                 severity: "info",
                 category: "power",
@@ -483,9 +486,9 @@ async fn handle_audit_power_rails_file(
     Ok(CallToolResult::text(
         serde_json::to_string(&json!({
             "audit": "power_rails",
-            "power_nets": power_nets,
+            "power_nets": rails.iter().map(|rail| &rail.name).collect::<Vec<_>>(),
             "findings": findings,
-            "summary": format!("{} power rail issues found across {} rails.", findings.len(), power_nets.len())
+            "summary": format!("{} power rail issues found across {} rails.", findings.len(), rails.len())
         }))
         .unwrap(),
     ))
@@ -1454,25 +1457,30 @@ type PlacedPin = (
 /// A placed unit and the pins it draws — one entry of [`placed_pins_by_reference`].
 type PlacedUnit = (konnect_sexp::schematic::SymbolInstance, Vec<PlacedPin>);
 
-/// Every net a placed unit's pins reach.
+/// Every net a placed unit's pins reach, as graph roots.
 ///
 /// Nets come from the shared net graph, so a pin reaches its net along the
 /// wires and junctions it is drawn with, and a rail named by a power symbol is
 /// a net like any other. The audits used to scan the file for a label within
 /// 0.5 mm of the pin, which followed neither.
 ///
+/// Roots rather than names, because an audit asks whether two pins are on *one
+/// net* and a net can carry more than one name. Comparing names made a rail
+/// named by both a power symbol and a label fail that comparison against
+/// itself, reporting the name the capacitors did not resolve to as an
+/// undecoupled rail of its own. A pin on an unnamed net has a root like any
+/// other and is counted here; it is the reporting sites that need a name.
+///
 /// `pins` comes from [`placed_pins_by_reference`], so it holds this unit's pins
 /// only: a triode of an ECC83 does not report the heater's net at its own
 /// coordinates (#182).
-fn pin_nets(graph: &mut NetGraph, pins: &[PlacedPin]) -> HashSet<String> {
-    let mut nets = HashSet::new();
-    for (pin, transform) in pins {
-        let (px, py) = pin_endpoint(pin, *transform);
-        if let Some(net) = graph.net_at(px, py) {
-            nets.insert(net);
-        }
-    }
-    nets
+fn pin_net_roots(graph: &mut NetGraph, pins: &[PlacedPin]) -> HashSet<NetRoot> {
+    pins.iter()
+        .map(|(pin, transform)| {
+            let (px, py) = pin_endpoint(pin, *transform);
+            graph.root_at(px, py)
+        })
+        .collect()
 }
 
 /// `C1` is a capacitor, `CN1` is a connector.
@@ -1504,11 +1512,11 @@ fn is_bulk_capacitor_value(value: &str) -> bool {
 fn capacitor_nets(
     graph: &mut NetGraph,
     placed: &[PlacedUnit],
-) -> (HashSet<String>, HashSet<String>) {
+) -> (HashSet<NetRoot>, HashSet<NetRoot>) {
     let mut all = HashSet::new();
     let mut bulk = HashSet::new();
     for (inst, pins) in placed.iter().filter(|(inst, _)| is_capacitor(inst)) {
-        let nets = pin_nets(graph, pins);
+        let nets = pin_net_roots(graph, pins);
         if is_bulk_capacitor_value(&inst.value) {
             bulk.extend(nets.iter().cloned());
         }
@@ -1518,13 +1526,13 @@ fn capacitor_nets(
 }
 
 /// Nets with a test point on them.
-fn test_point_nets(graph: &mut NetGraph, placed: &[PlacedUnit]) -> HashSet<String> {
+fn test_point_nets(graph: &mut NetGraph, placed: &[PlacedUnit]) -> HashSet<NetRoot> {
     let is_test_point = |inst: &konnect_sexp::schematic::SymbolInstance| {
         inst.reference.starts_with("TP") || inst.value.to_uppercase().contains("TESTPOINT")
     };
     let mut nets = HashSet::new();
     for (_, pins) in placed.iter().filter(|(inst, _)| is_test_point(inst)) {
-        nets.extend(pin_nets(graph, pins));
+        nets.extend(pin_net_roots(graph, pins));
     }
     nets
 }
@@ -1534,18 +1542,46 @@ fn test_point_nets(graph: &mut NetGraph, placed: &[PlacedUnit]) -> HashSet<Strin
 ///
 /// Built once per sheet. Asking per pin instead re-walked every resistor on
 /// the sheet for each I2C or reset pin found.
-fn pull_up_nets(graph: &mut NetGraph, placed: &[PlacedUnit]) -> HashSet<String> {
+fn pull_up_nets(graph: &mut NetGraph, placed: &[PlacedUnit]) -> HashSet<NetRoot> {
     let mut pulled = HashSet::new();
     for (_, pins) in placed
         .iter()
         .filter(|(inst, _)| inst.reference.starts_with('R'))
     {
-        let nets = pin_nets(graph, pins);
-        if nets.iter().any(|net| is_power_net_name(net)) {
+        let nets = pin_net_roots(graph, pins);
+        // Every name on the net, not the winning one: a rail KiCad calls `SYS`
+        // because a global label outranks its `+3V3` symbol is still the rail
+        // this resistor pulls to.
+        let on_a_rail = nets.iter().any(|root| {
+            graph
+                .aliases_of_root(*root)
+                .iter()
+                .any(|net| is_power_net_name(net))
+        });
+        if on_a_rail {
             pulled.extend(nets);
         }
     }
     pulled
+}
+
+/// One rail of the sheet: the net's graph root, and what that net is called.
+struct PowerRail {
+    root: NetRoot,
+    /// The name KiCad would print for this net.
+    name: String,
+    /// Every name on it. The ground skips read all of them: a rail a power
+    /// symbol calls `GND` and a global label calls `RETURN` is still ground,
+    /// even though the global label wins the name.
+    names: BTreeSet<String>,
+}
+
+impl PowerRail {
+    fn any_name_contains(&self, needle: &str) -> bool {
+        self.names
+            .iter()
+            .any(|name| name.to_uppercase().contains(needle))
+    }
 }
 
 /// The power rails on the sheet: every net a power symbol names, plus every
@@ -1554,14 +1590,40 @@ fn pull_up_nets(graph: &mut NetGraph, placed: &[PlacedUnit]) -> HashSet<String> 
 /// `labels` must be `extract_all_net_labels` — a `+3V3` symbol names a rail
 /// exactly as a label does. `PWR_FLAG` is not a rail and does not appear: its
 /// pin is `power_out`, which the extractor skips.
-fn collect_power_nets(labels: &[Label]) -> Vec<String> {
-    labels
+///
+/// One rail per *net*, not per name. A rail named by both a `+3V3` power symbol
+/// and a `VCC` label is one net that KiCad calls `+3V3`, and listing it twice
+/// made every audit below report the name the capacitors did not resolve to as
+/// a rail with nothing on it. The name is [`NetGraph::name_of_root`]'s, so it is
+/// the one KiCad's netlister would print.
+fn power_rails(graph: &mut NetGraph, labels: &[Label]) -> Vec<PowerRail> {
+    // Which nets are rails, and what each is known as, are separate questions,
+    // asked in that order. Collecting names while deciding made the answer
+    // depend on label order: a rail claimed by the `+3V3` symbol lost the `VCC`
+    // label read before it.
+    let mut rail_roots: BTreeSet<NetRoot> = BTreeSet::new();
+    for label in labels
         .iter()
         .filter(|label| label.kind == LabelKind::PowerSymbol || is_power_net_name(&label.net))
-        .map(|label| label.net.clone())
-        .collect::<std::collections::BTreeSet<_>>()
+    {
+        rail_roots.insert(graph.root_at(label.x, label.y));
+    }
+    let mut rails: Vec<PowerRail> = rail_roots
         .into_iter()
-        .collect()
+        .map(|root| {
+            let names = graph.aliases_of_root(root);
+            PowerRail {
+                name: graph
+                    .name_of_root(root)
+                    .or_else(|| names.iter().next().cloned())
+                    .unwrap_or_default(),
+                root,
+                names,
+            }
+        })
+        .collect();
+    rails.sort_by(|a, b| a.name.cmp(&b.name).then(a.root.cmp(&b.root)));
+    rails
 }
 
 fn is_power_net_name(name: &str) -> bool {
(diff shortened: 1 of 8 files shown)


neusse (changes requested):
Please preserve every alias before merging by name, make rail classification alias-complete, and replace the hand-authored regression strings with the accepted real-KiCad fixture.

The overall two-contract design is accepted: roots are identity after same-name segments are unioned, and one deterministic KiCad-precedence name is presentation. Three correctness gaps remain on exact head `a8c101e3c5c07988bc3a9a0847e71149783a16dc`:

1. `point_nets: HashMap<Point, NetName>` stores only one name at a coordinate. `add_label` discards lower-priority co-located aliases before `merge_named_nets`, so a point carrying `AAA` and `ZZZ` cannot join a disconnected `ZZZ` segment. Preserve all names per point (or merge from the complete label list), then choose the representative only after every alias has participated. Add the co-located-alias plus disconnected-segment transitive regression.
2. Build the complete alias set for every root first, then separately mark/classify rail roots. `power_rails` currently misses aliases encountered before the power-looking label, and `pull_up_nets` classifies from only `name_of_root`; a global winner such as `SYS` can hide the `+3V3` alias and still produce a false missing-pull-up finding. Test both orderings and the global-plus-power-alias pull-up case.
3. The accepted fixture requirement is not met: all new schematics are constructed as Rust strings. Add a KiCad 10-authored/resaved two-name-net fixture with provenance and external netlist expectations, then use it for the deterministic name, decoupling, pull-up/test-point, and read-only caller checks. Record specific negative-control failures after the repaired guards are neutered.

The measured precedence `global > power > local > hierarchical`, root-based direction, compatibility note, and caller inventory are otherwise sound. Update the issue/PR evidence and rerun the ten checks on the new head. #511 is the current next merge; this remains behind it until corrected and reviewed.

neusse:
Queue refresh — the existing correctness/evidence request remains unchanged on head `a8c101e3c5c07988bc3a9a0847e71149783a16dc`.

Please: (1) preserve every co-located alias before named-net unioning; (2) classify rail and pull-up roots from each root's complete alias set; and (3) replace the hand-authored regression schematics with an accepted KiCad-authored/resaved two-name-net fixture, external netlist expectations, and recorded negative controls.

Make those corrections now without repeatedly rebasing. After #512 lands, reconstruct once onto current `main`, preserve all migration/tool-directory entries, run all ten checks, and reply with the exact head SHA. The current branch is conflicting, so `status:waiting-on-author` remains correct.

pauliuszaleckas:
All three corrected on exact head `7bfe94548cff3fe4ba5f2a2c7ab221b843e8daea`. Not rebased — it is still on `2fbc5f5`, and I will reconstruct once onto current `main` after #512 lands, as you asked.

**1. Co-located aliases are preserved before unioning.** `point_nets` is now `HashMap<Point, Vec<NetName>>`. `add_label` keeps every name at a coordinate and ranks nothing; `merge_named_nets` anchors on each name at each point, so every alias participates; the representative is chosen afterwards, in `name_nets`, over the complete set. `aliases_of_root` exposes that set.

While building the fixture for this I found KiCad does not honour the exact shape your example describes. Two label *objects* at the identical coordinate: with `AAA` and `ZZZ` both at `(55.88, 119.38)`, KiCad leaves the wire's pin unconnected and nets only the far segment. Moving the second label 2.54 mm along the same wire nets both. So the fixture carries the co-located case KiCad *does* honour, which is also the one that matters here — a label on a power symbol's own pin, where Konnect synthesises the rail name as a pseudo-label at exactly that coordinate. `ALT` sits on the second `+3V3` symbol's pin, `TP7` hangs off a disconnected `ALT` segment, and KiCad's netlist puts `TP7.1` on `+3V3` (5 pins). Under the old one-name-per-point model the alias lost to `+3V3` and `TP7` was a net of its own. Both the transitive local-label case (`/AAA`) and this one are in the fixture and asserted.

**2. Alias sets are built first, classification second.** `power_rails` now marks rail roots in one pass and reads `aliases_of_root` in the next, so label order cannot lose a name. `pull_up_nets` classifies from the complete alias set rather than `name_of_root`. Both orderings and the global-winner case are covered by the fixture: `SYS` (global) outranks both the `+5V` symbol and the `PULLUP` local label, and `R1`/`R2` are still found as the pull-ups for `U1`'s `SDA`/`SCL`. The ground skips read every name too, so `RETURN` — a global label over `GND` symbols — is skipped rather than reported as an undecoupled rail.

**3. Real KiCad fixture with external expectations.** Every hand-authored regression schematic is gone. `crates/konnect-core/tests/fixtures/two_name_nets.kicad_sch` was built through Konnect against stock `Device:`, `Memory_EEPROM:`, `Connector:` and `power:` libraries, then force-resaved by KiCad 10.0.5 (`kicad-cli sch upgrade --force`); what is committed is that Eeschema serialization. Provenance, the recipe and the case table are in `two_name_nets.README.md`.

The expectations are KiCad's. `kicad-cli sch export netlist`:

| Net | Pins KiCad resolves | Shape |
|---|---|---|
| `+3V3` | 5 — U1.8, C1.1, C2.1, TP1.1, TP7.1 | one rail over three stubs; `TP7` joins only through the `ALT` alias on a power symbol's pin |
| `RETURN` | 4 — U1.4, C1.2, C2.2, C3.2 | ground named by a global label over its `GND` symbols |
| `SYS` | 4 — R1.1, R2.1, C3.1, TP6.1 | pull-up rail named by a global over `+5V` and `PULLUP` |
| `/SDA` | 2 — U1.5, R1.2 | pin and pull-up on separate segments joined by name |
| `/SCL` | 2 — U1.6, R2.2 | the same |
| `/AAA` | 2 — TP2.1, TP3.1 | two locals on one segment plus a disconnected `ZZZ` segment |
| `/MIX` | 1 — TP4.1 | a local label beside a hierarchical one |
| `Net-(TP8-Pad1)` | 2 — TP8.1, TP9.1 | a net with no label at all |

`kicad-cli sch erc --severity-all` states the precedence outcomes itself, which is the table `driver_priority` encodes:

```
Both +3V3 and ALT are attached to the same items; +3V3 will be used in the netlist
Both RETURN and GND are attached to the same items; RETURN will be used in the netlist
Both SYS and PULLUP are attached to the same items; SYS will be used in the netlist
Both AAA and ZZZ are attached to the same items; AAA will be used in the netlist
Both MIX and MIX_H are attached to the same items; MIX will be used in the netlist
```

The fixture drives the deterministic-name check (`get_pin_connections` on four multi-name nets, twenty runs each), decoupling, the rail/test-point audit, the pull-up audit, and the read-only caller (`get_connected_items`).

**Negative controls**, each neutered on this head, `cargo test -p konnect-core --lib`, then restored (baseline 1123 passed / 0 failed):

| Control | Fails | Tests |
|---|---|---|
| `merge_named_nets` not called | 6 | `two_segments_of_one_name_are_one_net`, `power_symbols_of_one_name_are_one_rail`, `an_alias_on_a_power_symbol_pin_joins_its_segment`, `a_capacitor_on_another_stub_of_the_rail_decouples_the_pin`, `each_rail_is_audited_once_under_kicads_name`, `a_pull_up_behind_a_global_alias_is_found`, `connected_items_span_every_segment_and_alias_of_the_net` |
| one ranked name per point (the shape you flagged) | 2 | `an_alias_on_a_power_symbol_pin_joins_its_segment`, `connected_items_span_every_segment_and_alias_of_the_net` |
| pull-ups classified from `name_of_root` | 1 | `a_pull_up_behind_a_global_alias_is_found` |
| ground skip reads the winning name only | 1 | `each_rail_is_audited_once_under_kicads_name` |
| name chosen by map order, not precedence | 7 | `a_multi_name_net_answers_with_kicads_name`, `a_net_with_two_names_answers_with_kicads_choice`, `both_names_of_a_net_share_one_root`, `an_alias_on_a_power_symbol_pin_joins_its_segment`, `a_rail_named_twice_is_audited_as_one_rail`, `each_rail_is_audited_once_under_kicads_name`, `connected_items_span_every_segment_and_alias_of_the_net` |
| `get_connected_items` keyed by name again | 2 | `connected_items_span_every_segment_and_alias_of_the_net`, `an_unnamed_net_reports_its_wire_and_the_part_at_the_far_end` |

One control I have to report as **not reproducing a failure**: comparing decoupling by name instead of by root passes the suite on this fixture, because after the name merge each rail has exactly one winning name, so name comparison and root comparison agree wherever a net is named. The root comparison still matters for the unnamed net, which `an_unnamed_net_reports_its_wire_and_the_part_at_the_far_end` covers.

**Migration and docs preserved.** `docs/API_MIGRATIONS.md` keeps the *schematic nets resolve by identity (minor release)* entry, and `tool-directory.md` plus the tool schema keep the corrected `find_shorted_nets` description — with same-named segments joined, two names can be netted together through a name rather than only through a wire path they share.

**Validation on `7bfe945`:** `cargo test --workspace --locked --lib --tests` 1123 passed / 0 failed / 8 ignored in `konnect-core` and green across the workspace; `cargo test --workspace --locked --doc` green; `cargo clippy --workspace --locked --all-targets -- -D warnings` clean; `cargo fmt --all -- --check` clean; `crates/schematic-viewer` 20 passed. The KiCad oracles above were re-run against the committed fixture on this head, not quoted from an earlier run. Hosted CI checks are whatever GitHub reports for this head; I have not run the Nix or PCM jobs locally.

neusse:
#512 has now landed, so the dependency hold is over. Please reconstruct this PR once onto current `main` (`5afdcbec76b3a8e55f87da778411ee09a98ae45b`) with only its intended unique commits, resolve the present conflicts, run all ten required checks on the resulting exact head, and reply with that SHA.

Your latest response supplies the requested implementation and evidence detail; no additional redesign is being requested before the clean-head review. Keep `Closes #510` as the terminal closure reference.

pauliuszaleckas:
Reconstructed once, one commit, exact head **`9b902036ed9b8d5acaa78c8bae9f780188b3a2c5`**. `Closes #510` is the terminal closure reference, in both the commit trailer and the PR body.

**Base.** `main` moved past the SHA you named while this was in flight: `5afdcbe` is now three commits back, and current `main` is `f58b222` (#519 and #520 landed after your message). I rebuilt on `f58b222` rather than `5afdcbe`, so this is one reconstruction rather than two — say the word if you want it pinned to `5afdcbe` instead.

**Conflict.** One, in `docs/API_MIGRATIONS.md`: both sides added a new entry at the top of the file. Resolved by keeping both, `main`'s four entries first and *schematic nets resolve by identity (minor release)* after them; no content from either side was dropped. Nothing else conflicted — #519 and #520 touch `manufacturing.rs` and `library.rs`, which this PR does not.

**Unique commits:** exactly one, `fix(schematic): resolve nets by identity, not by name`. `crates/konnect-core/src/tools/{sch_connectivity,sch_analysis,design_review}.rs`, `crates/konnect-sexp/src/schematic.rs`, the `two_name_nets` fixture and its README, the migration entry and the `tool-directory.md` line.

**The ten checks, on `9b90203`:**

| # | Check | Result |
|---|---|---|
| 1 | Check & Test — ubuntu | `cargo check --workspace --locked` clean; `cargo test --workspace --locked --lib --tests` **1773 passed / 0 failed / 28 ignored**; `--doc` 7 passed / 0 failed / 1 ignored |
| 2 | Check & Test — macOS | **not run** — no macOS host here |
| 3 | Check & Test — Windows | **not run** — no Windows host here |
| 4 | Plugin (Python) | `python -m unittest discover -s plugin/tests` — **10 passed** |
| 5 | Nix flake | **not run** — `nix` is not installed on this machine. This PR adds no workspace member, binary, or test-time environment variable, so the flake inputs are unchanged; hosted CI owns it |
| 6 | Schematic viewer | `cargo check` and `cargo test --locked --manifest-path crates/schematic-viewer/Cargo.toml` — **20 passed / 0 failed**. CI runs this job on Windows; mine ran on Linux |
| 7 | Clippy | `cargo clippy --workspace --locked --all-targets -- -D warnings` — clean |
| 8 | Dependency licenses | **not run** — `cargo-deny` is not installed here. This PR adds no dependency; `Cargo.toml` and `Cargo.lock` are untouched |
| 9 | Format | `cargo fmt --all -- --check` — clean |
| 10 | PCM packaging validation | `python packaging/validate-pcm.py --metadata packaging/metadata.json` — `OK: packaging/metadata.json passed PCM validation`. The two zip-assembly steps were not run |

So seven of the ten ran green locally; three could not run on this machine (macOS and Windows matrix legs, Nix) plus `cargo-deny`, and I have said which rather than implying coverage I do not have. Hosted CI on this head is the authority for those.

**KiCad oracles re-verified against the committed fixture at this head**, not carried over — `kicad-cli sch export netlist` returns the same eight-row table (`+3V3` 5 pins including `TP7.1`, `RETURN` 4, `SYS` 4, `/SDA` and `/SCL` 2 each, `/AAA` 2, `/MIX` 1, `Net-(TP8-Pad1)` 2), and `sch erc --severity-all` still prints the five `multiple_net_names` lines that state the precedence.

The implementation and evidence are unchanged from the previous head apart from the conflict resolution and the `Fixes` → `Closes` wording.

neusse (dismissed):
Exact-head review of 9b902036ed9b8d5acaa78c8bae9f780188b3a2c5: the production implementation appears sound and all ten hosted checks are green, but one evidence blocker remains.

The prior review required the deterministic-name regression to use the accepted real-KiCad fixture instead of hand-authored schematic strings. The new real-fixture test `a_multi_name_net_answers_with_kicads_name` says it covers five multi-name nets, but its table has four rows and one of those is the single-name `SDA` case. It does not assert the two fixture precedence outcomes that complete the KiCad oracle:

- `TP2` pin 1 must resolve to `AAA`, proving lexical precedence over `ZZZ`.
- `TP4` pin 1 must resolve to `MIX`, proving local-label precedence over `MIX_H`.

Please add those two assertions to the real-fixture deterministic test and remove the redundant new hand-authored `two_named_net` precedence test/builder. Then repeat the name-selection negative control against the real-fixture test, report the observed failure, and rerun CI on the resulting exact head.

No production-behavior defect or unrelated scope creep was found. Standards review found no hard violation; its only note was the non-blocking possibility of making `NetRoot` a newtype later.

mixelpixx (commented):
Clean-head review on exact head `9b902036ed9b8d5acaa78c8bae9f780188b3a2c5` (one unique commit on `f58b222`; ten checks green; zero review threads).

**Local gate on this head:** fmt / clippy `-D warnings` / `--lib --tests` (1758 passed, 0 failed) / `--doc` all exit 0; `cargo xtask fix-doc-counts --check` unchanged. **Neuter:** `merge_named_nets()` not called → 7 tests fail, restored tree green — the merge is load-bearing exactly where the PR says it is.

**Read end to end.** The model is right and small: `point_nets` keeps every alias at a point, `merge_named_nets` unions same-named points before naming, `driver_priority` is Global > PowerSymbol > NetLabel > Hierarchical with name-ascending ties — that is KiCad's `CONNECTION_SUBGRAPH` driver order and it matches the five `multiple_net_names` lines. Every name-keyed lookup in `design_review.rs` / `sch_analysis.rs` now goes through a root; I could not find a HashMap-order name left anywhere. The per-query `point_nets.clone()` going away is a real win. Your honest report that the decoupling by-name control does not fail on this fixture is accepted as documented: after the merge every named net has one winner, so name and root can only disagree on an unnamed net, which `an_unnamed_net_reports_its_wire_and_the_part_at_the_far_end` covers.

**One correction before merge — documentation only, no code, no rebase (your base is current):**

`docs/API_MIGRATIONS.md` describes `audit_power_rails` and `get_connected_items`, but two other tools change behaviour through code this diff does not touch, and a consumer has to be told:

1. **`find_shorted_nets`** keys off `label_roots`, so every two-name net is now reported as a short — five new on `two_name_nets` (`+3V3`/`ALT`, `RETURN`/`GND`, `SYS`/`PULLUP`, `AAA`/`ZZZ`, `MIX`/`MIX_H`). That is the right signal (it is KiCad's own `multiple_net_names` warning), and you already reworded the tool description, but the migration entry does not mention it.
2. **`points_on_net(name)`** now returns the whole merged net, so `get_net_components`, `get_net_connections` and `count_net_connections` return larger sets for any net that carries an alias.

Please add both to the existing *schematic nets resolve by identity* entry (which tools, what a caller sees differently, and that no field is renamed or removed). Small README nit in the same push if you like: the closing paragraph of `two_name_nets.README.md` calls `/AAA` and `/MIX` "the deliberately one-pin nets", while the table above it has `/AAA` at 2 pins (TP2.1, TP3.1).

**KiCad oracles, honestly:** the kicad-cli on this machine is 10.0.0-rc2 and refuses to load any `(version 20260306)` file, this fixture included, so I could not re-run the netlist and ERC tables here. Your numbers are internally consistent with the tests that parse the same fixture, and CI's tests pass on it. @neusse — you are on 10.0.6: when you approve, a one-line confirmation that `sch export netlist` gives the 8-row table and `sch erc --severity-all` prints the five precedence lines would close that gap from the right machine.

Label stays `status:waiting-on-author` for the migration edit only; reply with the head SHA and this merges on it. `Closes #510` stays the terminal reference.

mixelpixx:
Aligning with George's exact-head review above (same head, `9b90203`): fold the migration-entry addition from my review into the same push as his two real-fixture assertions and the hand-authored test removal, so there is one new head and one CI run. Label stays `status:waiting-on-author`; I will re-gate the new head and merge on it if George has not already.

pauliuszaleckas:
Both reviews addressed in one push. Exact head **`c5295071ab90f39894e222439f9831aa28aa2be6`**, base unchanged at `f58b222` — no rebase, as you both said.

## @neusse — the deterministic-name test

You were right that the test claimed five multi-name nets and asserted four rows, one of which (`SDA`) is single-name. It now asserts all five, plus `SDA` kept and labelled for what it is:

| pin | resolves to | precedence proved |
|---|---|---|
| `U1.8` | `+3V3` | power symbol over the `VCC` and `ALT` labels |
| `U1.4` | `RETURN` | global over the `GND` symbols |
| `R1.1` | `SYS` | global over both the `+5V` symbol and `PULLUP` |
| `TP2.1` | `AAA` | two locals, lexically first |
| `TP4.1` | `MIX` | local over hierarchical |
| `R1.2` | `SDA` | single name, two segments joined by it |

Those are the five `multiple_net_names` lines in the fixture's ERC report, one assertion each, twenty runs per row.

**The hand-authored precedence test and its builder are gone** — `two_named_net`, `a_net_with_two_names_answers_with_kicads_choice` and `both_names_of_a_net_share_one_root`. I removed two more of mine while I was there, `two_segments_of_one_name_are_one_net` and `power_symbols_of_one_name_are_one_rail`, so no new hand-authored schematic remains anywhere in this PR. Their graph-level coverage moved into the real-fixture test `an_alias_on_a_power_symbol_pin_joins_its_segment`, which now asserts three joins on one graph: `TP7`'s stub to the rail through the `ALT` alias, `C1`'s pin to the same rail through two `+3V3` symbols on separate stubs, and `U1.5` to `R1.2` through the `SDA` label alone.

**Name-selection negative control, repeated on this head** (`best.get(&root).is_none()` — first writer wins, precedence ignored): **5 tests fail**, restored tree 1186 passed / 0 failed.

```
a_multi_name_net_answers_with_kicads_name          (the real-fixture test you asked for)
an_alias_on_a_power_symbol_pin_joins_its_segment
each_rail_is_audited_once_under_kicads_name
a_rail_named_twice_is_audited_as_one_rail
connected_items_span_every_segment_and_alias_of_the_net
```

Previously this control failed 7 tests; two of those were the hand-authored ones now deleted, and the real-fixture test took their place.

The `NetRoot` newtype note is noted as non-blocking; happy to do it in a follow-up rather than widen this diff.

## @mixelpixx — the migration entry

Both tools added to the existing *schematic nets resolve by identity* entry, and I measured them on the fixture rather than reasoning about them:

- **`find_shorted_nets`** — `short_count: 5` on `two_name_nets`, reported as one group per net rather than one per pair: `+3V3`/`ALT`/`VCC`, `RETURN`/`GND`, `SYS`/`+5V`/`PULLUP`, `AAA`/`ZZZ`, `MIX`/`MIX_H`. The entry says that is the same condition KiCad's ERC reports as `multiple_net_names`, and points at the reworded tool description.
- **`points_on_net(name)`** — the entry names `get_net_components`, `get_net_connections` and `count_net_connections`, and says querying an alias returns the whole net. Measured: `get_net_components(net: "VCC")` on the fixture returns all seven items on the rail KiCad calls `+3V3` — `U1`, `C1`, `C2`, `TP1`, `TP7` and both `#PWR` symbols.

No field is renamed or removed; the entry says so.

**README nit fixed** — the closing paragraph now reads "`isolated_pin_label` on the labels of `/AAA` and the deliberately one-pin `/MIX`", which matches the two-pin `/AAA` row in the table above it.

On the oracle gap: my `kicad-cli` is **10.0.5**, and the eight-row netlist and five precedence lines in that README were produced by it and re-verified against the committed fixture on the previous head. A confirmation from 10.0.6 would still be worth having, since 10.0.5 is what wrote the file.

## Checks on `c529507`

| # | Check | Result |
|---|---|---|
| 1 | Check & Test — ubuntu | `cargo check --workspace --locked` clean; `--lib --tests` **1769 passed / 0 failed / 28 ignored**; `--doc` **8 passed / 0 failed** |
| 2–3 | Check & Test — macOS, Windows | **not run** — no such host here |
| 4 | Plugin (Python) | `python -m unittest discover -s plugin/tests` — OK, 10 tests |
| 5 | Nix flake | **not run** — `nix` not installed here; no workspace member, binary or test env var changes |
| 6 | Schematic viewer | check clean, `cargo test` **20 passed** (CI runs this on Windows; mine ran on Linux) |
| 7 | Clippy | `--workspace --locked --all-targets -- -D warnings` clean |
| 8 | Dependency licenses | **not run** — `cargo-deny` not installed here; `Cargo.toml`/`Cargo.lock` untouched |
| 9 | Format | `cargo fmt --all -- --check` clean |
| 10 | PCM packaging validation | `OK: packaging/metadata.json passed PCM validation` (zip-assembly steps not run) |

Hosted CI on this head is the authority for the three legs I cannot run.

One thing to flag rather than act on: `main` moved to `8ed319a` (#527) while I was making these edits, and that commit touches `docs/API_MIGRATIONS.md` and `tool-directory.md`. I have deliberately **not** rebased, per both your instructions, so this head is still on `f58b222`. If GitHub now reports a conflict against `main`, say the word and I will reconstruct once more onto `8ed319a`.

neusse:
The requested real-fixture assertions, hand-authored-test removal, and migration coverage are present on `c5295071ab90f39894e222439f9831aa28aa2be6`, and all ten hosted checks are green. The remaining merge-train requirement is the one you already anticipated: this head is still based on `f58b222`, before #485 and #527.

Please reconstruct once onto current `main` (`8ed319a3677118e6e056712363cede1adc1245f4` at this review), keep only this PR's unique net-identity commit and all current migration/directory entries, then rerun the ten checks and reply with the new exact head. No additional implementation redesign is requested. `Closes #510` remains the correct terminal reference.

mixelpixx:
Base correction, so this is one reconstruction rather than two: `main` moved past `8ed319a` a few minutes after @neusse's review above — #522 merged as **`30e0abdbba5688fea931990b05bbc63cec6fbac1`**, which is current `main` now.

Please reconstruct onto `30e0abd` instead of `8ed319a`. Your head `c529507` merges cleanly with #522 (I checked before merging: no code conflict, and the only file both touch is `docs/API_MIGRATIONS.md`, where #522 added an entry at the top — keep `main`'s entries above yours as you did last time). Nothing else in the request changes: the same one unique net-identity commit, all current migration and directory entries preserved, ten checks on the exact head, `Closes #510`.

My own two PRs are rebasing onto the same commit tonight, so this should be the last base move before yours lands.

pauliuszaleckas:
Reconstructed onto `30e0abd` as you asked, @mixelpixx — thanks for the base correction, it saved a round. Exact head **`2865c00fad91660cbd9f675093bf75b92c0328f1`**, one unique commit, `Closes #510`.

**The commit is byte-identical to the one you both reviewed.** `git range-diff c529507~1..c529507 2865c00~1..2865c00` reports `=`, not `!` — no implementation change came in with the move, which is the point of doing it as a reconstruction rather than a merge.

**Conflict.** One, `docs/API_MIGRATIONS.md`, exactly as you predicted: #522 added an entry at the top. Resolved your way — `main`'s entries above mine. The file now carries all eight in order:

```
atomic validation for schematic edits            (#527)
mirrored schematic placement                     (#485)
DRC checks schematic parity on every run         (#522)
add_mounting_hole uses KiCad's shipped names     (#512)
DRC item ownership                               (#481)
create_symbol draws a symbol body                (#502)
estimate_cost / validate_for_manufacturing …     (#511)
schematic nets resolve by identity               (this PR)
find_single_pin_nets counts pins, not labels     (#402)
```

`git diff 30e0abd -- docs/API_MIGRATIONS.md` contains only additions — nothing from `main` was dropped. `tool-directory.md` keeps both #527's line and my `find_shorted_nets` rewording; no other file conflicted.

**The ten checks, on `2865c00`:**

| # | Check | Result |
|---|---|---|
| 1 | Check & Test — ubuntu | `cargo check --workspace --locked` clean; `--lib --tests` **1796 passed / 0 failed / 29 ignored**; `--doc` **8 passed / 0 failed** |
| 2–3 | Check & Test — macOS, Windows | **not run** — no such host here |
| 4 | Plugin (Python) | `python -m unittest discover -s plugin/tests` — OK, 10 tests |
| 5 | Nix flake | **not run** — `nix` not installed here; this PR adds no workspace member, binary or test-time env var |
| 6 | Schematic viewer | check clean, `cargo test --locked --manifest-path crates/schematic-viewer/Cargo.toml` **20 passed** (CI runs this on Windows; mine ran on Linux) |
| 7 | Clippy | `--workspace --locked --all-targets -- -D warnings` clean |
| 8 | Dependency licenses | **not run** — `cargo-deny` not installed here; `Cargo.toml`/`Cargo.lock` untouched |
| 9 | Format | `cargo fmt --all -- --check` clean |
| 10 | PCM packaging validation | `OK: packaging/metadata.json passed PCM validation` (zip-assembly steps not run) |

The workspace count moved 1769 → 1796 because the base did; no test of mine was added or removed since `c529507`. Hosted CI on this head owns the three legs I cannot run locally.

Nothing else changed: the real-fixture assertions, the removed hand-authored tests, the negative controls and the migration coverage are all as reviewed on `c529507`.

mixelpixx (approved):
Exact-head review on `2865c00fad91660cbd9f675093bf75b92c0328f1`: approved, and merging.

**Verified on this head, not carried over from the earlier one:**

- Based on current `main` (`30e0abd`), one unique commit, ten hosted checks green, zero unresolved threads.
- Local gate: fmt / clippy `-D warnings` / `--lib --tests` (1781 passed, 0 failed) / `--doc` all exit 0; `fix-doc-counts --check` unchanged.
- **Neuter:** `merge_named_nets()` not called → five tests fail across `sch_connectivity`, `sch_analysis` and `design_review`; restored, green. The merge is load-bearing in all three consumers, not just the one that motivated it.
- @neusse's two assertions are there and passing: `TP2.1 → AAA` (two locals, lexically first) and `TP4.1 → MIX` (local over hierarchical), inside the real-fixture `a_multi_name_net_answers_with_kicads_name`.
- My own ask is met: the migration entry now names `find_shorted_nets` and the `points_on_net` consumers (`get_net_components`, `get_net_connections`, `count_net_connections`), so a caller learns about both behaviour changes from the file rather than by discovering them.

**On the reconstruction.** `git range-diff` reporting `=` is exactly the evidence that makes a base move reviewable without re-reviewing the implementation, and resolving the `API_MIGRATIONS.md` conflict by keeping every entry in order — with `git diff 30e0abd` showing additions only — is the right proof that nothing was dropped. Three rounds of review on this one, and the model it lands is the right one: identity is the union-find root, the winning name is presentation, and KiCad's own ERC output is the oracle for the precedence table.

@neusse — I am dismissing your `CHANGES_REQUESTED` on `c529507` to unblock the ruleset, not overriding it. Your own comment on that head reads: *"The requested real-fixture assertions, hand-authored-test removal, and migration coverage are present … The remaining merge-train requirement is … reconstruct once onto current `main`"* — which is precisely what `2865c00` is, byte-identical. Say the word if you would rather have re-reviewed it yourself and I will treat that as the rule next time.

`Closes #510`.

mixelpixx:
Following up on the one gap in my review: I merged this on your netlist and ERC numbers without being able to re-run them, because the kicad-cli on my machine was 10.0.0-rc2 and it refused to load any `(version 20260306)` schematic — including this fixture.

KiCad 10.0.5 is now installed here, and **both oracles reproduce exactly**:

`kicad-cli sch export netlist` on the committed fixture:

| net | pins |
|---|---|
| `+3V3` | 5 |
| `RETURN` | 4 |
| `SYS` | 4 |
| `/SDA` | 2 |
| `/SCL` | 2 |
| `/AAA` | 2 |
| `/MIX` | 1 |
| `Net-(TP8-Pad1)` | 2 |

`kicad-cli sch erc --severity-all` prints all five precedence lines, and they are the table `driver_priority` encodes:

```
Both AAA and ZZZ are attached to the same items; AAA will be used in the netlist
Both RETURN and GND are attached to the same items; RETURN will be used in the netlist
Both +3V3 and VCC are attached to the same items; +3V3 will be used in the netlist
Both MIX and MIX_H are attached to the same items; MIX will be used in the netlist
Both SYS and PULLUP are attached to the same items; SYS will be used in the netlist
```

So the external evidence behind #510 is independently confirmed on a second machine and a second KiCad build. Nothing to change — recording it because I merged on your word for it, and that should not stay unverified in the record.

Issue #549, opened 2026-09-13
Establish and enforce a shared reliability contract across Konnect tools


neusse:
# Establish and enforce a shared reliability contract across Konnect tools

Coordinate a finite first phase of reliability work: invalid inputs must not become defaults, unavailable evidence must not become a clean result, and partial or uncertain mutations must be visible to callers.

This issue covers Konnect's server, contributor documentation, bundled Claude guidance, and tests only. No runtime changes are implemented by opening this issue or its planning drafts. No AI provider is prescribed.

## Acceptance and PR ownership

- [x] A — Policy: add docs/RELIABILITY_CONTRACT.md; link CONTRIBUTING.md, PR template, DEV.md and bundled Claude guidance. Require a compact behavioral table only for affected behavior: inputs/defaults, invalid/unsupported inputs and structured errors, target/source/prerequisites, actual changes and preserved objects, failure before/after mutation, and safe recovery. Implemented by merged #550.
- [x] B — Validation: compile/cache input schemas; enforce supported schema semantics at dispatch, including nested arrays/unions and declared closed objects; use checked optional and integer readers at internal call boundaries. Catalogue-driven tests cover wrong types, fractional integers, declared bounds, closed-object unknown keys, and schema compilation. Preserve documented valid calls and publish compatibility changes. Coordinate #546/#547/#543; a bound absent from a schema still needs a domain check. Implemented by merged #551.
- [x] C — Test helpers: consolidate the five mock-port race helpers identified by #544 and use reusable fixtures/mocks for the new conformance tests. Verify listener lifecycle, concurrent runs, and cleanup. Implemented by merged #552.
- [x] D — Shared outcomes pilot: adopt consistent complete/partial/failed/uncertain outcomes and retry information for single/batch component placement and hierarchy-audit aggregation; observer accounting must distinguish partial from complete. Retain legacy response fields. Audit DRC, configuration and live/file readers into a named follow-up inventory rather than claiming those migrations complete. Implemented by merged #553.
- [x] E — CI: enforce the new validation/outcome guarantees with targeted automated checks; retain an explicit legacy inventory with paths, rationale, follow-up, and removal criteria. New regressions must not be added to the baseline automatically. Do not require whole-codebase migration for unrelated fixes. Implemented by merged #554.

## Evidence policy

[...]

PR #552, opened 2026-09-13
test(ipc): eliminate ephemeral-port races in konnect-core mocks

### LINKED ISSUE #544: Ephemeral-port TOCTOU fixed by 639bdf9 is still live in five konnect-core mock helpers
`639bdf9` ("fix(tests): eliminate ephemeral-port TOCTOU races in mock servers", 2026-07-28) fixed this pattern in `konnect-ipc`. The same pattern is still live in five `konnect-core` mock helpers, and it failed CI on #523 today.

**It fired in this repository's own CI.** Run [34767572364](https://github.com/mixelpixx/Konnect/actions/runs/34767572364) attempt 1, `Check & Test (ubuntu-latest)`, 2026-09-13T16:07:55Z, on #523:

```
---- tools::pcb_board::board_info_source_tests::a_live_board_is_reported_instead_of_the_last_save stdout ----
panicked at crates/konnect-core/src/tools/pcb_board.rs:2626:29:
mock listen: AddressInUse
test result: FAILED. 1225 passed; 1 failed; 12 ignored
```

macOS and Windows were cancelled by fail-fast, so one race reddens all three platforms. We have seen the same assertion twice more locally, in two other tests, always at `pcb_board.rs:2626`.

**The mechanism is the one `639bdf9` describes.** The helper probes a port by binding a `TcpListener`, reads the port, drops the listener — freeing the port — and only then has nng listen on it:

```rust
let port = {
    let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
    listener.local_addr().unwrap().port()
};                                     // listener dropped; port free again
let url = format!("tcp://127.0.0.1:{port}");
...
socket.listen(&url).expect("mock listen");   // pcb_board.rs:2626
```

**Five surviving sites**, all `#[cfg(test)]`, serving **49 tests** between them (32 of them through the first alone):

| bind → listen | helper |
|---|---|
| `pcb_board.rs:2618 → 2626` | `board_mock::spawn_kicad_reporting_documents` |
| `pcb_board.rs:2923 → 2928` | `board_session_safety_tests::spawn_one_document_response` |
| `pcb_board.rs:3890 → 3898` | `mounting_hole_tests::spawn_rejecting_kicad` |
| `pcb_components.rs:5664 → 5672` | `tests::spawn_rejecting_kicad` |
| `pcb_components.rs:5799 → 5807` | `tests::spawn_kicad_holding` |

**Rate, and what we could not do.** Sweeping the last 200 CI runs (2026-09-03 onward, every attempt of every rerun): 18 failed jobs, 1 carrying `AddressInUse` — roughly 1 in 200, and a lower bound, since only failed jobs with retained logs were scanned. **We could not force it locally**: 50 suite runs and ~15,200 binds, sequential and concurrent, with port-churn antagonists, with a delay injected into the window, and with 12,000 ephemeral ports held to force range wrap — zero natural collisions. What is deterministic is that the window exists: after the probe drops, another `TcpListener` binds the probed port and nng's `listen` then returns `AddressInUse`.

**The remedy already exists in-tree.** `639bdf9` moved the `konnect-ipc` mocks to `inproc://` with a process-unique name, and made the URL normalizer pass `inproc://` through for exactly that purpose (`crates/konnect-ipc/src/client.rs:623-632`, with a comment saying so). Six `konnect-core` helpers already use it. Where TCP is genuinely wanted, `nng` 1.0.1 also exposes `BoundPort` on a `Listener`, so a wildcard `tcp://127.0.0.1:0` can be bound and read back without a window — note that needs `Listener::new` rather than `Socket::listen`, which returns `()`.

**One observation beyond the five sites.** `git blame` shows one predates `639bdf9`, and the other four were written after it, by four different authors, between 2026-08-08 and 2026-08-29 — including the refactor that made one racy helper serve 32 tests. The convention is not recorded in `docs/TESTING_AND_RELEASE.md` or anywhere else, so it gets re-copied from the nearest neighbour.

neusse:
Closes #544.
Part of #549 (workstream C).

## What changed

- adds one test-only `MockIpcServer` fixture that binds a process-unique `inproc://` endpoint before returning;
- makes the fixture guard own the NNG endpoint and worker through `Drop`, so tests cannot leave detached mock servers behind;
- migrates all five racy `konnect-core` helpers identified in #544 and retains the guard at every affected call site;
- records the fixture convention in `docs/TESTING_AND_RELEASE.md` and updates the #549 implementation ledger.

Production IPC behavior and transport selection are unchanged.

## Regression evidence

The fixture tests cover:

- ready-before-return round trips;
- 64 concurrently live, distinct endpoints;
- endpoint release after ordinary cleanup;
- cleanup and endpoint release after a responder panic;
- 32 repeated create/respond/drop/rebind lifecycles.

A source audit confirms the old `TcpListener::bind("127.0.0.1:0")` probe/drop pattern no longer exists in either migrated module. Other ephemeral TCP tests retain the listener that owns the port and are outside #544.

## Validation

- focused board, component, routing, placement, and fixture suites

Live KiCad was not run because this changes only test infrastructure; no production IPC path changed.


mixelpixx:
Post-merge review note (I gated this head while you were merging it; it stands as evidence, not a request). Exact head `814b6b0`: fmt, clippy `-D warnings`, `--lib --tests` 1818 passed / 0 failed, `--doc`, counts unchanged; the five fixture tests pass in ~1 s; every call site keeps its guard alive (`_server` is a named binding, not `let _ =`), and the two explicit `drop(server)` calls are the deliberate KiCad-went-away tests. One neuter did **not** get caught: removing `self.control.close()` from `Drop` leaves all five tests green at 320 s instead of 1 s, because the worker's 10 s `RecvTimeout` releases the endpoint eventually. Filed as #556 with the measurement; a bounded-drop assertion closes it. Nothing here changes the merge.

PR #555, opened 2026-09-13
fix(sexp): honor the process umask when creating files

### LINKED ISSUE #538: Created project files land at mode 0600, ignoring umask — unreadable to a second user or CI account
## Summary

Every file Konnect **creates** lands at mode `0600`, ignoring the process umask. A
project created through Konnect is therefore unreadable to a second user, to a CI job
running as another account, or to a container user with a different uid.

Files Konnect **rewrites** are fine — the rewrite path copies the destination's existing
mode. This is specific to the create-only path.

## Environment

- Konnect 0.11.1, built from `main` @ 39e0061
- KiCad 10.0.5, Fedora Linux, `umask 022` (so `0644` is expected)

## Repro

```
create_project(name="permcheck", path=<dir>)
create_schematic(path=<dir>/solo.kicad_sch)
register_symbol_library(library_path=<dir>/my.kicad_sym, nickname="mylib", project=...)
```

Resulting modes:

```
0600  permcheck.kicad_pro
0600  permcheck.kicad_pcb
0600  permcheck.kicad_sch
0600  sym-lib-table
0600  solo.kicad_sch
```

## Negative controls

Two writers in the same directory, same process, same umask, come out correct — so this
is not the umask and not the filesystem:

```
0644  my.kicad_sym          <- create_symbol (Konnect, different write path)
0644  svg/solo.svg          <- kicad-cli sch export svg
```

## Cause

`create_scratch_file` (used by the *replace* path) opens with `OpenOptions`, so it gets
`0666 & ~umask` = `0644`, and `write_atomic_unlocked_with` then copies the destination's
mode over it. Both are right.

The *create* path, `write_new_atomic_unlocked` in `crates/konnect-sexp/src/writer.rs`,
builds its scratch file with `tempfile::Builder`, which creates at `0600` by design, and
then never adjusts the mode — there is no destination to copy from:

```rust
let mut temporary = tempfile::Builder::new()
    .prefix(".konnect-")
    .tempfile_in(parent)?;
...
temporary.persist_noclobber(path)
```

`0600` survives the rename and becomes the created file's permanent mode.

## Affected call sites

Everything that reaches `write_new_atomic` / `write_new_atomic_unlocked`:

- `create_schematic` (`sch_components.rs`)
- `create_project`, via `commit_file_transaction` → `transaction.rs:701`
- `register_symbol_library` (`library.rs:2227`)
- `schematic_builder.rs:383`
- `export_specctra_dsn` and its manifest (`pcb_export.rs:677`, `:679`)
- the Freerouting SES output (`freerouting_mcp.rs:214`, `:237`)

The last two matter beyond readability: exported artifacts get handed to other tools and
other people.

## Not a bug

`persist_journal` (`transaction.rs:728`) uses the same path for
`.konnect-transaction-*.json`. Those hold complete before/after file images and are
documented as sensitive project data, so `0600` is the right mode there and should stay.

## Expected

A newly created design file gets the mode `File::create` would have given it —
`0666 & ~umask` — leaving the journal at `0600` deliberately rather than by accident.

wangzhengzhuo05:
## Summary

Every file Konnect **creates** was landing at mode `0600`, ignoring the process umask, so a project created through Konnect was unreadable to a second user, to a CI job running as another account, or to a container user with a different uid. Files Konnect **rewrites** were already correct, which made this a create-only path bug. The create path now receives the ordinary create mode, so the umask decides — `0644` under `022`, matching the replace path and the negative controls in the report.

Issue: #538

## Approach

Root cause is `write_new_atomic_unlocked` in `crates/konnect-sexp/src/writer.rs`. It built its scratch file with `tempfile::Builder`, which creates at `0600` by design, and there was no destination to copy a mode from — `0600` survived the rename and became the created file's permanent mode.

The replace path (`create_scratch_file`) does not have this problem because it opens with `OpenOptions ... .create_new(true)`, so it gets the ordinary `0o666 & !umask` default the kernel applies; it then copies the destination's mode over the scratch file. The create path is the only writer that never touched that default.

Fix: ask `tempfile::Builder` for the ordinary create mode before `tempfile_in(parent)`:

```rust
#[cfg(unix)]
{
    use std::os::unix::fs::PermissionsExt;
    builder.permissions(std::fs::Permissions::from_mode(0o666));
}
```

`tempfile` 3.27 passes that mode to `OpenOptions::mode()`, so the kernel applies the umask — the result is `0o666 & !umask`, identical to what the replace path already produces. Gated to `#[cfg(unix)]`; Windows has no mode to set.

Alternatives considered: reading the umask directly (needs a `libc` dependency, and `umask(2)` is a destructive getter — set then restore — that races with other threads), and a post-`persist` `set_permissions` (leaves a window where a crash strands the file at `0600`). Neither is needed: letting the kernel apply the umask is exactly what the replace path already relies on.

## Branch and dependencies

Base branch: `main` (`upstream/main`, commit `74134c5`)

Depends on: nothing

Series order, if any: none — one independent PR

Unique commits/acceptance criteria owned by this PR: the create-only write path honors the process umask, with a regression test

Next PR to promote after this one, if any: none

## Compatibility and safety

No public API, config, or schema changes. No new dependency.

Behavior change is limited to the mode of newly created files, which is the point of the issue: created files are now as permissive as the user's umask allows, instead of unconditionally owner-only. Atomicity is untouched — the scratch-file-then-rename sequence, `persist_noclobber`, `sync_all`, the KiCad lock recheck, and the parent-directory fsync are all unchanged.

The affected call sites are exactly those already listed in the issue, all of which reach `write_new_atomic` / `write_new_atomic_unlocked`: `create_schematic`, `create_project` (via `commit_file_transaction`), `register_symbol_library`, `schematic_builder`, and `export_specctra_dsn` with its manifest.

### Changed tool behavior

|---|---|
| Accepted inputs and declared defaults | Unchanged. `write_new_atomic` still refuses an existing destination via `persist_noclobber`; the only default that moved is the created file's mode. |
| Invalid/unsupported inputs and structured errors | Unchanged. No new error arm; `SexpError::Io` still wraps a failed persist. |
| Target, data source and prerequisite state | Unchanged. Same destination path, same `.konnect-*` scratch sibling, same KiCad lock gate before and during the write. |
| Observed changes and preserved unrelated objects | Newly created files now carry `0o666 & !umask` (e.g. `0644` under `022`) instead of `0600`. Nothing else is touched; the regression test creates a real file and compares its mode against a plain `create_new` control in the same directory. |
| Failure before/after mutation, including applied work | Unchanged. A refusal or failure still happens before the rename, so the destination is never partially written. |
| Recovery from partial/uncertain results without repeating applied work | Unchanged. Recovery is the existing scratch-file cleanup path; this change adds no new state. |

### Commands run

- [x] `cargo fmt -p konnect-sexp -- --check` — clean
- [x] `cargo clippy -p konnect-sexp --all-targets -- -D warnings` — clean
- [x] `cargo test -p konnect-sexp --lib --tests` — 193 + 16 + 10 passed, 0 failed (1 ignored, pre-existing)

Regression test added: `atomic_create_honors_the_process_umask` (`#[cfg(unix)]`, in the existing `atomic_write_tests` module). It calls `write_new_atomic`, then creates a control file with a plain `create_new` in the same directory and asserts both modes match. Comparing against a control keeps the assertion true under any umask while still failing for `tempfile`'s `0600` default.

Mutation check: setting the requested mode back to `0o600` makes the new test fail (`left: 384, right: 436` — `0600` vs `0664` under the local `002` umask); restoring the fix makes it pass. So the test actually guards this behavior rather than merely passing alongside it.

### Environment-dependent checks not run

`protoc` is not installed in this environment, so the workspace-wide commands from the PR template (`cargo test --workspace --locked --lib --tests`, `--doc`, `cargo clippy --workspace`) could not be run here; they fail on `konnect-ipc`'s protobuf codegen before reaching any test. The crate-scoped equivalents above were run instead. The change is confined to `konnect-sexp` and its only test-visible surface, so CI on ubuntu/windows/macos is the authoritative check for the workspace scope.

Closes #538


neusse:
Please preserve the private transaction-journal mode before this enters the merge queue.

The diagnosis and ordinary-created-file fix are sound, but the current change applies `0o666` to every `write_new_atomic_unlocked` caller. `persist_journal` in `crates/konnect-sexp/src/transaction.rs:728` calls that same helper, so `.konnect-transaction-*.json` files would also become umask-readable. Those journals contain complete before/after design images, and issue #538 explicitly says they must remain `0600`.

Please make one focused correction:

- give the internal create path an explicit ordinary-versus-private creation policy (or an equivalently narrow helper), while keeping public design/artifact creates on the ordinary `0o666 & !umask` path;
- make `persist_journal` select the private `0600` path without teaching the generic writer about journal filenames;
- add a Unix regression that inspects a persisted journal before removal and proves it remains `0600`, alongside the existing ordinary-create/control comparison;
- run the negative controls for both guards, then reconstruct or rebase the corrected unique commit onto current `upstream/main` and rerun the ten required checks.

This PR remains independent and first in its own overlap set. Once those items are complete, reply with the new exact head for final review. The issue is now assigned to you and marked claimed so ownership is visible.

wangzhengzhuo05:
Corrected at `a7bca14`, rebased on current `upstream/main` (`10c392a`) as a single commit.

**1. Explicit creation policy on the internal create path.** `writer.rs` now has a
`NewFileMode { Ordinary, Private }` enum and a `write_new_atomic_unlocked_with_mode`
worker. `write_new_atomic_unlocked` passes `Ordinary` (`0o666`, umask decides) — so
`apply_entry` and every public design/artifact create are on exactly the path you
approved. `write_new_atomic_unlocked_private` passes `Private` (`0o600`). The public
`write_new_atomic` signature and behaviour are untouched.

**2. `persist_journal` selects the private path.** It now calls
`write_new_atomic_unlocked_private(path, &source)`. The writer never learns about journal
filenames — the policy is a parameter, and the only mention of journals is a one-line
comment at the call site in `transaction.rs`. `apply_entry` still uses the ordinary helper.

**3. Unix regression, journal inspected before removal.**
`transaction::tests::transaction_journal_is_created_private` persists a real journal via
`persist_journal`, stats the file while it still exists, then builds an ordinary
`create_new` control in the same directory and asserts `journal_mode == 0o600` and
`journal_mode != ordinary_mode` before calling `remove_journal`. The existing
`writer::atomic_write_tests::atomic_create_honors_the_process_umask` control is unchanged.

**4. Negative controls for both guards.**

```
Ordinary -> 0o600 : FAILED  writer::atomic_write_tests::atomic_create_honors_the_process_umask
Private  -> 0o666 : FAILED  transaction::tests::transaction_journal_is_created_private
```

Each mutation kills exactly its own guard and nothing else, so the two policies are
independently pinned rather than one riding on the other.

`cargo test -p konnect-sexp` — 194 passed, 0 failed (7 doc-tests passed, 1 ignored);
`cargo clippy -p konnect-sexp --all-targets` clean; `cargo fmt -p konnect-sexp --check`
clean. Rerunning the ten required checks now.

pauliuszaleckas (approved):
Tested this as the reporter of #538 — the fix is correct. Built the branch and re-ran the
issue's repro against the built server: all five files that were `0600` are now `0644`,
and the umask is genuinely consulted rather than `0644` hardcoded (`077`→600, `027`→640,
`002`→664).

Splitting `Ordinary` / `Private` instead of blanket-widening the create path is the right
call. Both halves hold in a real transaction, not just the unit test — sampling the
project directory while `create_project` runs catches the journal live at `0600` beside
its `0644` design files.

Full gate passes here, including the workspace-wide commands your description lists as
not runnable (I have `protoc`): workspace tests, doctests, clippy, fmt, and the
schematic-viewer suite `--workspace` skips.

**One nit, non-blocking:** `transaction_journal_is_created_private` fails under
`umask 077`:

```
assertion `left != right` failed: the journal must not inherit the ordinary create mode
  left: 384    right: 384
```

Both are `0o600` — under a `077` umask an ordinary create is *also* `0o600`, so the two
legitimately coincide. The code is fine (I sampled a live transaction under `077`: modes
still correct); only the assertion can't distinguish them. It's the only umask-sensitive
test in the workspace, but it would fail CI on a `077` runner. Guarding it does the job:

```rust
if ordinary_mode != 0o600 {
    assert_ne!(journal_mode, ordinary_mode, "...");
}
```

Two things worth a line somewhere, no code change needed: existing projects keep `0600`
through later edits (the replace path preserves the destination's mode — correct, but
anyone who already hit #538 must `chmod` manually), and `Private` is `#[cfg(unix)]`, so
journal privacy is a Unix-only guarantee while DEV.md describes journals as sensitive
without qualification.

neusse:
One important correction to my review after checking the private-mode implementation against Unix creation semantics: guarding only the inequality is not sufficient.

`tempfile::Builder::permissions(0o600)` supplies the creation mode to `open`; the process umask still masks those bits. A restrictive umask can therefore create the journal as `0400`, `0200` or even `0000`, despite #538's explicit requirement that it remain `0600`. The test's exact-mode claim is consequently both environment-dependent and not guaranteed by the implementation.

Please make the private path enforce `0600` after the temporary file is created and before persistence (while it still has a private-at-most creation mode), then test representative explicit umasks in isolated subprocesses so the test does not mutate the parallel test process's global umask. Assert both exact owner read/write and no group/other access. Neutralizing the private enforcement should make the permissive-umask case fail. The ordinary path should continue to receive the normal `0666 & ~umask` result.

This replaces the narrower “remove or guard the inequality” suggestion; the rest of the earlier checklist is unchanged.

neusse (changes requested):
Please correct the new journal-mode regression test before this enters the merge queue.

The ordinary/private implementation now matches #538, and the reporter's live evidence confirms the exact head. The remaining assertion in `transaction_journal_is_created_private` is false under a valid `umask 077`: an ordinary `0666 & ~umask` create and the required private journal are both legitimately `0600`, so `assert_ne!(journal_mode, ordinary_mode)` fails even though the implementation is correct.

Keep the direct `journal_mode == 0o600` assertion and either remove the inequality or guard it when `ordinary_mode != 0o600`, as the reporter suggested. Please also record in the PR evidence that existing already-created `0600` design files retain their mode on replacement and require a one-time manual `chmod` if the owner wants to widen them; this fix governs newly created files. Rerun the full gate and ten checks on the corrected exact head.

wangzhengzhuo05:
Thanks @neusse — closing this.

Your second correction was the important one and I want to record it here so it does not get lost: `tempfile::Builder::permissions(0o600)` only supplies the *creation* mode to `open`, so the process umask still masks it — a restrictive umask can leave the journal at `0400`, `0200` or `0000`, which is not what #538 requires. The fix is therefore to enforce `0600` after the temporary file is created and before persistence, and to test it in isolated subprocesses with explicit representative umasks (asserting exact owner read/write and no group/other access), with the negative control being "neutralize the enforcement → the permissive-umask case must fail". The ordinary path keeps `0666 & ~umask`.

That checklist is correct and complete; I am simply not able to work it in the near term, and leaving a `CHANGES_REQUESTED` PR parked in your merge queue with an unanswered checklist is worse for you than a closed one. So: closed, with the diagnosis and the ordinary-created-file fix preserved on the fork branch `fix/created-file-umask` for anyone who wants to pick it up.

Please also feel free to unassign me on #538 — the claim should be free for whoever takes the enforcement work.

Thanks again for the two rounds of Unix-semantics review; the second one genuinely changed how I would approach file modes.

Issue #296, opened 2026-08-19
Advanced symbol and footprint generation controls


eriksenfredrik:
## Context

I found these gaps while generating a KiCad component for Murata
`LBAD0XX1SC-005`: a 115-land LGA module with a five-unit symbol.

![Murata Type 1SC-NTN](https://www.murata.com/-/media/webrenewal/products/connectivitymodule/lpwa/overview/lineup/product-img_2.ashx?cvid=20201228090309000000&la=en)

The test component passes KiCad 10.0.5 parse/export and DRC. It has 115 pads,
115 pins, no different-net pad overlaps, and 0.325 mm minimum land clearance.

## Missing controls

Footprints:

- chamfered pads and oval drills;
- per-pad mask, fabrication, and zone settings;
- rotation-aware courtyard and silkscreen clearance; and
- safe pin-1 marker placement.

Symbols:

- standard KiCad metadata;
- field placement across all units; and
- string escaping and nested-input validation.

## Proposed split

1. Production pad controls
2. Footprint graphics and clearance fixes
3. Symbol metadata and multi-unit fixes

I have a local implementation and regression tests. Is this three-PR split
acceptable?

neusse:
The three-PR implementation split looks right to me. I would add one more follow-up PR so the new capability is actually usable by AI clients without relying on guessing or manual KiCad cleanup.

Suggested fourth PR:

4. Guidance (skill) and agent workflow updates for custom library generation

The Rust/tool changes provide the missing capability. The guidance tells the client when and how to use those controls safely, especially for complex parts like a 115-land LGA where a syntactically valid symbol/footprint can still be physically wrong.

That guidance (skill) should require:

- search existing KiCad libraries before custom generation;
- exact datasheet and package-suffix match before creating a symbol or footprint;
- a physical lead -> symbol pin -> footprint pad evidence table for custom parts;
- explicit top-view/bottom-view handling for mirrored, circular, socketed, or array packages;
- query-back verification of the generated symbol and footprint before schematic use;
- parse/export/DRC checks after generation;
- visible review of pin-1 marking, pad shapes, drill geometry, courtyard, fab, and silkscreen;
- fail closed if the required pad, graphics, metadata, or validation controls are not exposed by the MCP tool schema.

[...]

Issue #471, opened 2026-09-07
No tool can write a non-rectangular board outline


mixelpixx:
Found in Chris's end-to-end run on Konnect v0.11.0 / KiCad 10.0 / Windows 11 / Claude Desktop (QuadNode PCIe accelerator card). Verified against current main unless noted.

add_board_outline and set_board_size only emit rectangles. A PCIe CEM card outline is ~20 line segments plus 3 arcs (finger tab, key notch, relief and fillet radii), so any real add-in card or connector cut-out has to be written into Edge.Cuts by hand. #413 (footprint-owned Edge.Cuts) is adjacent. Now that delete_graphics exists (#335), an add_board_outline_polyline accepting segments and arcs is the natural pair.

neusse:
## Recommendation: separate outline writing from outline specification

I think this capability is useful, but `add_board_outline_polyline` by itself would solve only the final drawing step. The harder and more safety-critical question is where the geometry came from.

A request such as “make a PCIe card outline” does not provide enough information for a manufacturing result. The agent would still need the applicable standard revision and card type, an explicit coordinate datum, dimensions, radii, connector and bracket relationships, cutouts, and tolerances. We should not encourage an AI client to reconstruct that geometry from memory and then hand a plausible-looking result to fabrication.

Konnect already has much of the low-level machinery needed: `OutlinePrimitive` represents lines and arcs, IPC builders can create both shapes, the file fallback can serialize them, and `delete_graphics` can remove an old outline. The missing work is therefore mainly a safe public contract and validation rather than a new geometry engine.

### Proposed division of responsibility

Use three paths according to the source requirement:

1. Rectangular or rounded-rectangular boards continue using the existing parametric tools.
2. Simple custom boards can use an explicit ordered list of line and arc primitives.
3. Standards-driven or enclosure-matched boards should normally use authoritative DXF/SVG geometry or a vetted `.kicad_pcb` template. The AI should coordinate and validate that input, not invent it.

KiCad 10 already supports importing DXF and SVG graphics onto `Edge.Cuts`. Its board-outline contract also gives us the right validation model: contours must be closed, endpoints must coincide, and contours must not self-intersect or overlap; enclosed closed contours represent internal cutouts. See the [KiCad 10 PCB Editor documentation](https://docs.kicad.org/10.0/en/pcbnew/pcbnew.pdf).

### Suggested Konnect API

Rather than an append-only polyline command, I recommend a guarded `replace_board_outline` operation with:

[...]

Issue #351, opened 2026-08-28
feat(pcb): add lossless promotion of footprint Edge.Cuts to board graphics


scottxiaoyao:
## Problem

Imported KiCad/EasyEDA boards may store the physical board outline and cutouts as `Edge.Cuts` graphics owned by a mechanical footprint.

Konnect can inspect these graphics with `list_board_footprint_graphics`, but it cannot losslessly convert them into top-level native board graphics.

The existing `add_board_outline` tool only supports rectangular or rounded-rectangular outlines. `import_svg_logo` is not a safe substitute because it creates filled polygons and does not preserve native lines/arcs or outline recognition.

## Real-world case

Environment:

- Konnect 0.10.0
- KiCad 10.0.5
- macOS 26.6.2 arm64

The board contains a mechanical footprint named `MECH1` at `(100, 100)`, rotation `0°`.

`MECH1` owns 27 `Edge.Cuts` primitives:

- Main external contour: 11 segments + 8 arcs
- Internal rounded cutout: 4 segments + 4 arcs
- Overall extents: X = 30–170 mm, Y = 50–150 mm
- Internal cutout: 24.250 × 7.333 mm
- The footprint has no electrical pads

The geometry forms two closed contours and is valid in KiCad.

## Current behavior

Because the `Edge.Cuts` items are footprint-owned:

- `score_placement` reports `outline_missing: true`
- MECH1's bounding box is treated as a footprint courtyard
- 133 false component-overlap findings are produced
- The raw placement score is 30 although there are no real courtyard overlaps

There is currently no safe MCP sequence that can:

1. create arbitrary top-level `Edge.Cuts` segments and arcs;
2. preserve the exact transformed coordinates;
3. validate contour closure;
4. delete the source mechanical footprint only after successful conversion.

## Requested capability

Please add either:

1. generic native board-graphic creation tools supporting segments and arcs; or
2. a higher-level lossless conversion tool for footprint-owned `Edge.Cuts`.

A possible high-level operation could be conceptually similar to:

```text
replace_footprint_edge_cuts_with_board_graphics(
    board,
    reference,
    dry_run=true,
    expected_plan_revision=...
)
```

The final public tool name can follow the maintainers' preferred naming.

## Required safety behavior

The operation should:

- default to dry-run;
- report the source primitive count, contour count and global extents;
- apply the footprint translation, rotation and side/mirroring transform;
- preserve exact segment endpoints and arc start/mid/end geometry;
[...]

PR #561, opened 2026-09-14
feat(library): support oval drills and pad readback

BlackMesaTech:
## Summary

Connectors with plated slots cannot be authored through the numeric-only `drill` argument. This adds round/oval drill creation and editing, plus optional per-pad readback to verify the saved geometry. A `1.0 × 1.3 mm` slot can be requested as `{"drill":{"shape":"oval","width":1.0,"height":1.3}}`.

Part of #296: this PR owns oval drills and the readback needed to verify them. The other generation controls remain outside its scope, so it does not close the parent issue.

## Approach

- Extend the existing `create_footprint` and `edit_footprint_pad` tools. A positive numeric drill remains circular; the new object describes an oval in pad-local millimetres. Copper-pad shape stays independent, and pad rotation rotates the drill.
- Replace the complete direct-child drill block structurally. This also fixes scalar edits of a drill with a nested offset: stopping at the first closing parenthesis could otherwise leave malformed source. Existing offset nodes and unrelated footprint content are preserved.
- Add `get_footprint_info(include_pads=true)` for saved-file pad positions, sizes, rotations, layers and drill dimensions/offsets. Repeated pad numbers remain separate physical pads. Default summaries remain unchanged.
- Update the tool directory and bundled library skill. Tool names and counts do not change.

## Branch and dependencies

Independent branch from current upstream `main`, `10c392aa638c79771abc64127b130e5be48c9b4d`. No unmerged prerequisites or subsequent PR is required. The branch contains one unique commit, `be2f08613a2e6bb97095149927fa0e4d8275f172`.

## Compatibility and safety

Existing positive scalar drill calls and omitted-drill behavior remain supported. Nonpositive dimensions, malformed objects and drills on SMD pads are refused. The new object accepts only `shape`, `width` and `height`; offsets are preserved, not set or cleared by this API.

Library operations remain file-based. Edit batches preflight every selected pad before the existing atomic, if-unchanged write. An incompatible later match leaves the original file byte-identical. This changes neither the PCB IPC mutation path nor the file format; generated slots use KiCad's native oval-drill representation.

## Validation

|---|---|
| Accepted inputs and declared defaults | Positive scalar circle or closed oval object; omission preserves an edited drill. Round, vertical/horizontal oval, plated/unplated and rotated-pad coverage. `include_pads` defaults to false. |
| Invalid/unsupported inputs and structured errors | Invalid dimensions/types/shape/extra fields and SMD drills return `invalid_argument`; direct-handler and served MCP tests verify rejection. Malformed or ambiguous saved drill evidence is an error. |
| Target, data source and prerequisite state | Explicit library footprint path; optional pad results identify `source=file`. No live PCB or IPC state is required or claimed. |
| Observed changes and preserved unrelated objects | Tests query saved geometry and compare unrelated fixture bytes; nested offsets, mechanical properties, UUIDs, duplicate pad numbers and other pads survive. KiCad independently loaded the generated 16-pad pattern. |
| Failure before/after mutation, including applied work | New validation failures occur before writing. A later invalid match prevents the entire edit batch. Existing atomic persistence/concurrency error handling remains unchanged. |
| Recovery from partial/uncertain results without repeating applied work | Validation refusal leaves the file unchanged. On an uncertain persistence error, inspect the explicit target with `include_pads=true` before retrying; no live editor state is implied. |

- [x] `cargo test --workspace --locked --lib --tests`: 1,843 passed, 30 existing environment-dependent tests ignored.
- [x] `cargo test --workspace --locked --doc`: 8 passed, 2 ignored.
- [x] `cargo build --locked -p konnect`
- [x] Eight focused regression tests, including a served JSON-RPC create/edit/read path.
- [x] Seven negative controls: disabling positive-dimension validation, through-hole type checks, the closed oval-object check, duplicate-drill write/read checks, offset-structure handling or finite readback each caused a regression test failure. All controls were restored before the full gate.
- [x] macOS / KiCad 10.0.6: the committed executable generated a 16-pad connector pattern through MCP; KiCad's native loader reported 16 plated oval drills with positions, copper dimensions and slot dimensions matching the source within 0.4 nm of rounding. KiCad SVG export succeeded.

KiCad-authored fixtures cover the existing socket and a USB-C mechanical pad with a nested drill offset. Live PCB GUI/IPC, Windows and Linux observations were not run locally; this PR changes saved-library operations. Hosted CI remains the platform gate.

## Review checklist

- [x] Focused diff; no generated files, downloaded catalogues, personal settings or unrelated source changes.
- [x] Based on latest upstream `main`, with only the unique commit listed above.
- [x] Public names, schema, compatibility behavior and documentation agree.
- [x] New behavior and relevant failure paths have regression coverage.
- [x] File edits preserve unrelated content and preflight the complete selected batch.
- [ ] Hosted CI passed on the current head and maintainer review is complete.

Please retain this PR for review and later merging. Auto-merge is not enabled.


neusse:
This is the right focused scope for oval-drill support, and I do not want to expand #561 before it merges.

Two related capabilities remain useful:

1. A focused follow-up under #296 should let callers explicitly create, edit, and clear KiCad's native drill offset. This PR already does the important compatibility work of parsing, preserving, and reading an existing offset.
2. Reusable arbitrary mechanical cutouts should continue through #351 and #471 as footprint-owned or board-level `Edge.Cuts` geometry, rather than introducing drill shapes that KiCad does not support.

Neither item should block this PR. I am opening the drill-offset follow-up separately; the `Edge.Cuts` work is already tracked.

neusse (changes requested):
Please replace the hand-wrapped `OFFSET_PAD` fragment with a complete committed KiCad-authored footprint fixture before this enters the merge queue.

The implementation otherwise matches #296's focused oval-drill/readback scope, and #565 correctly remains a separate follow-up. The remaining problem is fixture provenance: `library_drill_tests.rs` copies one stock pad into a hand-authored minimal `(footprint ...)` wrapper, while both `CONTRIBUTING.md` and `docs/RELIABILITY_CONTRACT.md` require real KiCad output for footprint parsing and mutation tests. A fragment can omit surrounding syntax or structure that triggers the parser defect.

Please:

- commit the complete stock KiCad `.kicad_mod` containing the nested offset drill;
- select the target `SH` pad from that unchanged fixture for preservation/edit/readback coverage;
- retain the complete-file no-unrelated-change assertion;
- use that independent fixture to corroborate readback rather than relying primarily on a footprint generated by the implementation under test;
- rerun the full gate and all ten checks on the resulting exact head and update the PR evidence.

No expansion into offset setting/clearing or `Edge.Cuts` is requested here.

PR #562, opened 2026-09-14
fix(pcb): read pads from legacy footprint references

BlackMesaTech:
## Summary

Fix saved-board pad reads for footprints whose reference uses legacy `fp_text reference` syntax. Such footprints currently return “not found” even though KiCad loads them correctly. The same lookup gap also lets an empty live pad response pass when the saved legacy footprint actually has pads.

## Approach

Reuse the existing `footprint_reference` helper in both the file reader and saved-pad consistency check. The helper preserves modern `property "Reference"` precedence and recognizes the legacy spelling. Regression tests use KiCad-authored geometry with one reference converted to the legacy syntax, retain a modern-reference neighbor, and verify pad numbers, coordinates, nets, layers, read source and unchanged file bytes through both direct and served MCP paths.

## Branch and dependencies

- Base: current `main`, `10c392aa638c79771abc64127b130e5be48c9b4d`.
- Independent change; no dependency on the oval-drill PR #561 and none of its commits are included.
- This PR owns only legacy-reference recognition in pad readback and the existing empty-live-response consistency check.

## Compatibility and safety

No public API, schema, tool-count or configuration changes. This is read-only. Target binding, live-first selection, refusal on reachable-editor rejection and recovery instructions remain unchanged. No CAD write or file-format migration is added.

## Validation

|---|---|
| Accepted inputs and declared defaults | Existing board/reference arguments; both modern and legacy references tested, including served `tools/call`. |
| Invalid/unsupported inputs and structured errors | Existing missing-part, deleted-live-part and rejected-IPC cases remain covered by the full suite; no schema changes. |
| Target, data source and prerequisite state | Live board still takes priority. Offline fixture reports `source: file`; mocked empty live response is refused when legacy saved pads exist. |
| Observed changes and preserved unrelated objects | Both fixture footprints return their independent expected pad geometry/nets/layers. Source bytes remain unchanged. |
| Failure before/after mutation, including applied work | No mutation. Refusal does not return a fabricated empty pad list or change the file. |
| Recovery from partial/uncertain results | Existing save-and-retry diagnostic is preserved for contradictory live/saved pad counts. |

- [x] `cargo test --workspace --locked --lib --tests`: 1,837 passed; 30 existing environment-dependent tests ignored.
- [x] `cargo test --workspace --locked --doc`: 8 passed; 2 ignored.
- [x] Negative control: before the two lookup changes, both new regression tests failed. The offline case returned `Footprint 'R1' not found`; the empty-live case wrongly succeeded. Both pass with the fix.
- [x] The triggering custom footprint was independently present with its pads in native KiCad readback. Candidate behavior is exercised through served MCP regression tests; a new live-editor session with this candidate binary was not run. Windows/Linux interactive sessions and packaging are not claimed.

## Review checklist

- [x] Focused diff with no generated CAD output, personal design files or unrelated cleanup.
- [x] Based on current upstream `main`; independent unique diff.
- [x] Success and material failure behavior covered.
- [ ] Hosted CI and maintainer review of the final head.

Please leave merging to the maintainer workflow; no auto-merge is requested.


neusse:
Please add and link the originating bug issue before this enters final review.

The patch is focused and all ten checks are green, but it currently has no issue or acceptance/closure accounting. The issue should preserve the reported legacy `fp_text reference` reproducer, identify the affected read paths, and state that this PR is terminal if its direct and served readback tests satisfy the complete scope. Then add `Closes #N` to the PR body (or explain why it is only partial) and reply with the issue link. No code reconstruction is requested at this point.

Issue #449, opened 2026-09-07
The current latest version (v0.11.0) supports only two-layer boards by default.


mayjack0312:
The current Konnect routing configuration does not support this 4-layer PCB. Both `native_bridge_mode="prefer"` and `"require"` return the same error:

- unsupported first routing profile:
- expected exactly F.Cu and B.Cu,
- got [F.Cu, In1.Cu, In2.Cu, B.Cu]

This has been diagnosed using `get_recent_calls` and `server_stats`. It is a limitation of the Konnect export phase.

neusse:
Current-main verification (`4a5dd41`): this remains true, and the limitation is broader than the issue currently states.

Evidence:

- `crates/konnect-core/src/specctra.rs` still requires the copper-layer list to be exactly `F.Cu`, `B.Cu` and emits the reported `unsupported first routing profile` error for a four-layer stack.
- `export_specctra_dsn` builds the Rust baseline through `specctra::export_dsn` before it attempts the optional ActionPlugin bridge. Consequently, both `native_bridge_mode: "prefer"` and `"require"` encounter the two-layer baseline refusal before native export can help.
- `crates/konnect-core/src/specctra_ses.rs` separately requires `manifest.supported_profile.copper_layers == 2`, so merely reordering the native export attempt would not provide a complete four-layer route-and-return workflow.
- No linked or pending PR currently addresses #449.

I recommend treating this as an end-to-end routing-profile extension rather than removing the first guard. Acceptance should cover:

1. Preserve the board's ordered copper-layer stack in the DSN and reverse manifest instead of assuming two layers.
2. Allow the KiCad 10 native bridge to provide a multi-layer DSN without first requiring a two-layer Rust baseline, while retaining snapshot binding and semantic validation.
3. Validate SES routes on inner copper layers and validate legal through/blind/buried via spans against the manifest and board stack.
4. Apply the validated result through the existing atomic IPC transaction and readback boundary.
5. Add a real four-layer KiCad fixture and an end-to-end KiCad -> DSN -> Freerouting -> SES -> KiCad regression, including inner-layer routing and via-span evidence.
6. Continue to fail closed for unsupported layer mappings or geometry rather than silently reducing a multi-layer board to two layers.

So the issue should remain open at `status:design-needed`. This is a meaningful Specctra profile expansion across export, manifest, validation, and import—not a small guard removal.

neusse:
## Architecture review: accepted as a native KiCad/Freerouting workflow

[...]

PR #564, opened 2026-09-14
Add guarded tools for schematic text annotations

BlackMesaTech:
## Summary

Plain schematic notes could be added through MCP but could not be read by UUID, revised or removed. This prevented a protocol-removal workflow from clearing obsolete notes without editing KiCad source outside the tools.

Add `list_schematic_texts`, `edit_schematic_text` and `delete_schematic_text` to `sch_batch`. Edits preserve every other saved byte; deletions remove only the selected plain annotation. Net labels, symbol fields, text boxes and nested library text are excluded.

Closes #563.

## Approach

Read the saved schematic's top-level UUID index and return an opaque document revision with each list result. Mutations require that revision and a plain-text UUID, then use the existing revision-aware atomic schematic command writer. They validate the prospective annotation and committed readback before reporting success. Group-member deletion is refused to avoid dangling UUID references.

The existing KiCad 10 `ampli_ht` fixture provides real saved formatting and styled text for regression checks. Tests cover escaped multiline content, unrelated-byte preservation, duplicate and wrong-type targets, stale revisions, concurrent writes, editor locks, group membership and failures after persistence.

## Branch and dependencies

- Base: current `upstream/main`, `10c392aa638c79771abc64127b130e5be48c9b4d`.
- Dependencies: none; one independent commit, no unmerged prerequisite changes.
- Scope: plain annotation CRUD, public contract documentation and registry-derived tool counts.
- No follow-on PR is required for this issue. Merge remains a maintainer action.

## Compatibility and safety

Three additive tools; existing calls and file formats remain compatible. The list source is explicitly the saved file, including when an editor has newer unsaved state. Mutations refuse editor-owned files and stale or ambiguous targets. Text edits preserve position, style, UUID and unknown fields.

The atomic writer prevents concurrent lost updates. A post-write readback failure or mismatch returns `mutation_outcome_uncertain`; it does not claim rollback. Inspect a fresh list before retrying. Control characters other than newline, carriage return and tab are rejected. Those accepted characters, quotes, backslashes and Unicode round-trip.

## Validation

| Behavior | Contract and evidence for this change |
|---|---|
| Accepted inputs and declared defaults | Explicit saved schematic path; edit/delete require UUID and exact `expected_revision`; edit requires text. No mutation defaults. Served-schema test verifies discovery and dispatch. |
[...]

PR #566, opened 2026-09-14
feat(pcb): expand copper layers through guarded live IPC

BlackMesaTech:
## Summary

Add `expand_copper_layers` so an open PCB can gain copper layers through live KiCad IPC. The existing `add_layer` operation edits saved text and cannot safely perform this operation on an open board.

The new tool accepts an even count from 2 through 32, defaults to dry-run, checks the expected live count, preserves all existing enabled layers, refuses removal, and verifies the result with a fresh read. It does not save the board or select a physical fabrication stackup.

## Branch and dependencies

Based on upstream `main` at `10c392a`. Independent of the open library and schematic PRs; this branch contains only the live copper-layer expansion change and its catalogue updates.

## Compatibility and safety

This adds one tool without changing existing APIs. The requested board must be active and reachable. There is no file fallback. KiCad's layer change is not undoable; removal is deliberately unsupported. A send or readback failure reports that the change may already have applied and directs the caller to inspect live layers before retrying. An equal-count request is a no-op.

## Validation

| Behavior | Contract and evidence |
|---|---|
| Inputs and defaults | Even integer counts 2–32; expected live count required; dry-run defaults to true. |
| Invalid inputs | Tests cover odd counts, wrong boolean type, stale expected count and attempted shrinking; none sends a mutation. |
| Target and source | Live requested-board guard; wrong-board and disconnected tests refuse without changing the file. |
| Observed changes | Stateful mock verifies preservation of enabled layers and a fresh readback. Real KiCad 10.0.6 expanded a populated PCB from four to six copper layers. Native readback retained 1.6 mm thickness, 110 footprints, 401 physical pads, all 16 plated connector slots, and all existing tracks/vias. |
| Partial or uncertain result | Readback mismatch returns an error, not success. Error messages after dispatch identify possible applied work. |
| Recovery | Inspect live state; update the expected count before another call. Matching counts do not resend the mutation. |

Passed locally:

- `cargo test -p konnect-core --locked copper_expansion_tests --lib` (six focused tests after the lint correction)
- Formatting and whitespace checks.

The real-board check validates layer addition and preservation, not a fabrication stackup or manufactured-board performance. Hosted CI and maintainer review remain pending. No merge or auto-merge requested.


neusse:
Please link this new public tool to a focused issue and make the closure accounting explicit before final review.

`expand_copper_layers` is not the full solution to #449: it expands the live board stack, while #449 still requires multilayer DSN export, manifest preservation, SES validation/import and via-span evidence. Either open a dedicated issue for guarded live layer expansion and close that issue here, or mark this PR `Part of #449` and identify the terminal work that will close #449. Keep the current no-save/no-stackup-selection boundary explicit.

The implementation is otherwise presented as an independent one-commit change on current main with all ten checks green. Note that it overlaps #564 in generated count/public-directory files; #564 is the current base-forming tool-addition candidate, so this PR will need one clean reconstruction after #564 lands. Do not rebase it repeatedly in the meantime.

PR #603, opened 2026-09-16
feat(pcb): expose typed pad geometry

### LINKED ISSUE #409: get_component_pads and get_footprint_info never expose pad size, shape, or rotation
## Summary

Neither `get_component_pads` nor `get_footprint_info` (even with `include_graphics: true`) return a pad's physical dimensions. Every pad-related read tool returns position, layer, and net — never size, shape, or rotation. This forces routing decisions to be made blind to actual copper extent, which caused real shorts.

## Environment

- Konnect 0.10.0

## Repro

```
get_component_pads(board="...", reference="U1")
→ {"pad_count":28,"pads":[{"layers":["F.Cu","F.Mask"],"net":"","number":"1","x":148.155,"y":74.4975}, ...]}
```

No `size`, `shape`, or `rotation` field anywhere in the response, for any of the 28 pads.

```
get_footprint_info(footprint_path="Seeed Studio XIAO Series Library:XIAO-ESP32-C6-DIP", include_graphics=true)
→ {"graphic_count":19, "graphics":[...19 entries, all on F.SilkS/F.Fab/F.CrtYd...], "pad_count":28, ...}
```

`include_graphics` returns silkscreen lines/arcs, the fab-layer outline, and the courtyard rectangle — zero entries for any of the 28 real pads, despite `pad_count: 28` being reported correctly alongside it.

The only way found to get actual pad geometry was to shell out and parse the `.kicad_mod` file directly:

```python
# via grep/python on the raw footprint file, outside any Konnect tool:
(pad "1" smd roundrect (at 7.62 -8.455) (size 2.432 1.524) ...)
(pad "1" thru_hole circle (at 7.62 -7.62) (size 1.524 1.524) (drill 0.889))
```

## Expected

`get_component_pads` (or `get_footprint_info`) reports each pad's `size`, `shape`, and effective rotation in board space, so routing/clearance decisions can be made without reading the library file directly.

## Actual

Pad geometry is unavailable through any read tool. Routing had to proceed on estimated pad widths.

## Impact

Four real DRC shorts were introduced on the first routing pass because the actual SMD pad width (2.432 mm) was assumed to be close to the THT pad's 1.524 mm — the only number available from any Konnect tool. All four were only caught by a subsequent `run_drc` call, not by anything Konnect could tell in advance.

## Affected versions

Confirmed on Konnect 0.10.0 / KiCad 10.0.5. Distinct from the now-closed #248 (`pad_count` returning 0 due to a whitespace-matching bug) and #56 (mirrored position for rotated footprints) — this is a missing field, not an incorrect one.

neusse:
## Summary

- extend placed-pad IPC and saved-board responses with UUID, pad type, effective rotation, size, shape, drill data, and per-copper-layer geometry
- add opt-in `include_pads` output to `get_footprint_info`, explicitly marked as `footprint_local`
- use one typed file serializer for placed-board and library reads, preserve live/file response-shape parity, and fail instead of silently omitting unreadable pads
- update the shipped library and PCB guidance to request and use measured pad geometry

## Behavioral contract

| Case | Result |
| --- | --- |
| `get_component_pads` with live KiCad | Board-space position and effective rotation from IPC, plus native pad-stack geometry |
| KiCad unavailable / board not open | Same response keys from the saved board file, with `source: "file"` |
| `get_footprint_info` default | Existing compact response; no pad array |
| `get_footprint_info(include_pads=true)` | Typed pads in footprint-local coordinates |
| Unknown future pad type or shape | `null`, rather than a guessed value |
| Custom pad shape | Explicit `shape: "custom"`; no ordinary outline is invented |
| Unreadable saved pad position/identity | Tool error; the pad is not silently dropped |

## Evidence

- real KiCad-authored board fixture proves a rotated footprint reports KiCad's saved effective 90-degree pad rotation, not a double-rotated value
- real KiCad-authored USB-C footprint fixture covers NPTH pads, duplicate pad numbers, SMD roundrects, THT ovals, round drills, and oval drills
- mock IPC readback covers unsaved/live geometry and response-shape parity with file fallback
- malformed-pad tests prove unreadable pads fail loudly

## Tests

- `cargo test -p konnect-ipc -p konnect-core --lib --tests`
- `cargo test -p konnect --lib`
- `cargo clippy -p konnect-ipc -p konnect-core --all-targets -- -D warnings`

Closes #409


neusse:
Added the two explicit coverage cases identified during the final issue-closure audit:

- a live IPC custom PTH pad with distinct `F.Cu`, `In1.Cu`, and `B.Cu` geometries, an oval drill, and non-cardinal rotation;
- a saved back-side footprint proving board-space pad rotation and `B.Cu` layer reporting.

Negative-control evidence: I temporarily changed the shared live serializer to return `shape: null`. `cargo test -p konnect-core live_custom_front_inner_back_stack --lib` then failed at the new assertion (`null` vs `custom`). After restoring the serializer, the same test passed.

Local gates on `cb13564`:
- `cargo test -p konnect-ipc -p konnect-core --lib --tests`
- `cargo test -p konnect --test asset_references`
- `cargo clippy -p konnect-ipc -p konnect-core --all-targets -- -D warnings`
- `git diff --check`

All passed. Required hosted checks are now running on the new head.

PR #588, opened 2026-09-15
fix(schematic): report pins and junctions at a traced point

pauliuszaleckas:
## Summary

`trace_from_point`'s description promised "what is at that point"; it returned
`wires_here` and `labels_here` only. A component pin and a junction dot at the same
coordinate were dropped, and no `pins_here`/`junctions_here` key was present to be
empty, so the response looked authoritative while omitting the two things that decide
whether a connection is real.

The response gains two always-present arrays. `pins_here` carries `reference`, `pin`,
`pin_name`, `electrical_type`, `x`, `y` — spelled as `find_orphan_items` already spells
them. `junctions_here` carries `x`/`y`. `x`, `y`, `net`, `wires_here` and `labels_here`
are unchanged in name, shape and content.

Both new arrays are derived from `ConnectivityIndex`, the shared owner of what attaches
at a point, and the index is unit-aware, so a multi-unit symbol contributes only the
unit actually placed (#35) and pins stacked on one point are all reported.

`tolerance` is now validated: zero, negative and non-numeric values are refused with a
structured `invalid_argument` instead of being answered.

Closes #539

## Addressing the review on `c314162`

**1. `tolerance` was accepted unvalidated.** It is read with `opt_positive_f64` and
declared `exclusiveMinimum: 0` in the schema, so it is refused at dispatch and in the
handler alike — the spelling `find_orphan_items` already uses. `tolerance: -1` no longer
answers `{"net":"NETA", …}` with four empty arrays; it answers
`invalid_argument` naming the field. `an_unusable_tolerance_is_refused_not_answered_empty`
covers `0`, `-1` and a non-number, and asserts the refusal names no net;
`a_positive_tolerance_is_still_accepted` is the paired positive control, so the refusal
cannot widen into refusing the argument outright.

**2. The tool scanned pins and junctions itself.** `ConnectivityIndex` now exposes
`point_detail(x, y) -> PointDetail`, returning the placed pins and junction dots at a
point rather than the counts and booleans the attachment predicates answer with. The
handler builds one index at the requested tolerance and reads that; the two local scans
are gone, so tolerance and attachment semantics have one owner again.

Two supporting changes made that possible:

- `PointIndex` now stores each point's build ordinal alongside its coordinates and
  exposes `hits_at`, so a lookup can recover the *item* and not only that something is
  there. `count_at`/`contains` are expressed over it and are unchanged in behaviour.
  `pin_points` is built from `placed_pins` in order, which is what makes the ordinal an
  index back into that slice.
[...]

PR #605, opened 2026-09-16
fix(schematic): bound the stale-target refusal

### LINKED ISSUE #592: A stale-target refusal repeats one fact once per symbol, and carries it twice
Found dogfooding on a real board (ch347t-swd-debugger, 27 parts) while rebuilding its
schematic from scratch. Measured on 0.11.1 @ `c0f9180`, KiCad 10.0.6 on Fedora, driving
the freshly built server over stdio.

The refusal itself is correct and I am not asking for it to be relaxed — only for the
answer to be a readable size.

## What happens

Copy a `.kicad_sch` to a different filename stem with no sibling `.kicad_pro`, then call
any mutating schematic tool on it. The placed symbols still record the old project name
while the file now resolves to a new one, so `validate_sheet_instance_state` refuses —
and then says so once per symbol:

| Sheet | `error.reason` | `message` | Whole response |
|---|---|---|---|
| 2 symbols | 292 B | 489 B | 979 B |
| 59 symbols | 8 014 B | 8 212 B | **16 425 B** |

That is ~131 B per placed symbol. The 59-symbol sheet is a 27-part board — the rest are
`#PWR` symbols, which is what makes the number surprising: an ordinary small schematic
already produces a 16 KB error.

The information content is one fact — every symbol records project X, this file resolves
to project Y. Each of the 59 entries repeats it, and the `observed [...]` / `expected
[...]` pairs differ only in that one project name, which the message prefix has already
stated.

On an MCP server the caller paying for this is a model's context window, and several
clients truncate it, so the actionable prefix survives and the rest is cost without
information.

## Repro

```sh
# any project with a few symbols
cp proj/proj.kicad_sch proj/renamed.kicad_sch
# then, over MCP:
#   add_power_symbol {"schematic": ".../renamed.kicad_sch", "power_net": "GND", "x": 10.16, "y": 10.16}
```

Measure `len(response.content[0].text)` and compare `error.reason` against `message`.

## Cause

Two independent multipliers, neither wrong on its own.

1. `validate_sheet_instance_state` (`crates/konnect-core/src/tools/mod.rs`) pushes one
   `"{identity}: observed [...], expected [...]"` string per stale symbol into
   `stale_symbols`, then joins the whole vector — no cap, no dedup, and no check for the
   overwhelmingly common case where every entry carries an identical diagnosis.

2. `SchematicTargetError::StaleTarget::into_tool_result` hands that string to
   `CallToolResult::error_kind` **twice**: once as the `reason` field of
   `ToolErrorKind::StaleTarget`, and again interpolated into the human-readable
   `message`. `error_kind` serialises both into one body, so every structured error
   carries its reason twice. That is free when a reason is one line, which is why it has
   gone unnoticed; it matters here only because the first multiplier makes the reason
   long.

## Suggested fix

Collapse the reason when the diagnosis is uniform: state the mismatch once — `every
placed symbol records project 'X'; this file resolves to project 'Y'` — followed by a
bounded sample and a count, keeping per-symbol detail only where symbols actually
disagree with each other. Capping the join alone would fix the size without fixing the
redundancy.

The doubling in `error_kind` is worth a separate look, since it applies to every
structured error rather than just this one.

pauliuszaleckas:
## Summary

A stale-target refusal built `error.reason` by joining one line per placed symbol.
A sheet goes stale as a whole — copy a `.kicad_sch` to a new filename stem and every
symbol records the old project name — so the answer repeated one fact once per symbol.
On a 46-symbol sheet that is 18 535 B of `reason` inside a 37 301 B response, growing
linearly with the sheet. On an MCP server the caller paying for that is a model's
context window, and several clients truncate it.

The refusal itself was correct and is unchanged. Only the size of the answer changes.

Closes #592

## Approach

Two independent multipliers were reported. Per the triage on #592 this PR fixes the
first only; the second (`error_kind` serialising `reason` into both the structured
field and the human `message`) affects every structured error and is left to its own
change. The fix stands alone because the reported problem was growth with sheet size,
not the constant factor.

`validate_sheet_instance_state` now:

- states the expected instance identity once, in the prefix, instead of once per symbol;
- reports how many of the sheet's placed symbols are stale;
- groups symbols by their diagnosis, naming at most 3 symbols and 5 distinct diagnoses
  before counting the rest.

```text
placed-symbol instance metadata disagrees with project 'complex_hierarchy': 46 of 46
placed symbols are stale; every symbol must record exactly [complex_hierarchy:/<root>/<sheet>];
RV201, C201, R203 and 43 more (46 symbols): observed [foreign-project:/<root>/<sheet>]
```

| Sheet (46 placed symbols) | `error.reason` | whole response |
|---|---|---|
| before | 18 535 B | 37 301 B |
| after | 547 B | 1 373 B |

Worst case — every symbol failing differently — went 4 405 B to 2 861 B and is now
capped rather than linear.

**Instance lists are deliberately never sampled.** An instance list has one entry per
placement of the sheet in the hierarchy, not one per symbol, so it was never the axis
that grew. It is also what a caller must write back to clear the refusal, and it is the
grouping key — sampling it would both hide the repair and merge symbols that are stale
in genuinely different ways. A comment at `format_paths` records this.

### Architectural fit

Extends the existing shared preflight (`validate_sheet_instance_state`) that every
mutating schematic tool already calls; no new module, no new error kind, no workaround.

The same unbounded "one line per item, then join" shape exists in roughly eight other
error builders across `tools/*` (`sch_wiring.rs`, `sch_batch.rs`, `sch_components.rs`,
`pcb_sync.rs`, `placement.rs`, `pcb_export.rs`). A shared bound-and-sample helper near
`mcp/error.rs` would be the deeper fix, but that is a multi-file change with its own
per-site sample sizes and belongs in a separate issue alongside the `error_kind`
doubling. This PR stops at the call site #592 named, so `join_bounded` stays a private
implementation detail of `tools/mod.rs` rather than premature shared infrastructure.

## Branch and dependencies

- Base branch: `main` @ 240e295
- Depends on: nothing
- Series order: none
- Unique commits: one, `fix(schematic): bound the stale-target refusal`
- Next PR to promote: none

## Compatibility and safety

No tool, argument, or response-field change. `error.kind` stays `stale_target`,
`error.target` is unchanged, the same documents are refused, and nothing is written.
Only the `reason` text — and the `message` interpolating it — is shorter.

Callers matching on `error.kind` need no change. A caller parsing individual symbols
out of `reason` should read the sampled identities and counts, or inspect the file;
`reason` is prose for a human or a model, not a parsed field. Documented in
`docs/API_MIGRATIONS.md` and `docs/PROJECT_OWNERSHIP.md`.

`docs/API_MIGRATIONS.md`: the new entry is appended above the existing ones. Every
entry already on `main` is left untouched and in its original order, so this file's
diff against `main` deletes nothing and moves nothing.

### Changed tool behavior

|---|---|
| Accepted inputs and declared defaults | Unchanged — no schema, argument, or default touched. |
| Invalid/unsupported inputs and structured errors | Same inputs refused, same `stale_target` kind and `target`; only `reason`/`message` text is shorter. Asserted in all four new tests. |
| Target, data source and prerequisite state | Unchanged — same saved-file preflight over the same parsed root hierarchy. |
| Observed changes and preserved unrelated objects | None. Every new test asserts the child sheet and the root are byte-identical after the refusal. |
| Failure before/after mutation, including applied work | Refusal is preflight, strictly before mutation; nothing is applied. Covered by the byte-identical assertions above. |
| Recovery from partial/uncertain results without repeating applied work | Not applicable — nothing is written, so there is no partial state. The refusal now names the full expected instance list, which is what a repair has to write back. |

Commands and results, all on this exact head (c80adf5, Fedora, Rust stable):

- [x] `cargo fmt --all -- --check` — pass, no output
- [x] `cargo test --workspace --locked --lib --tests` — exit 0, 31 suites ok, 0 failed
- [x] `cargo test --workspace --locked --doc` — exit 0
- [x] `cargo clippy --workspace --locked --all-targets -- -D warnings` — exit 0, 0 warnings
- [ ] Relevant viewer, plugin, packaging, and real-KiCad checks — **not run locally.**
  This change touches no viewer, plugin, packaging, or IPC path, and adds no fixture
  file. Left to hosted CI.

New coverage, all driving the served `tools/call` dispatch through `McpHandler`, not
the handler directly:

| Test | Case |
|---|---|
| `native_placement_collapses_a_uniform_stale_diagnosis` | Whole sheet stale for one reason: stated once with a sample and a count; response bounded. |
| `native_placement_bounds_the_number_of_distinct_diagnoses` | Every symbol stale differently: distinct diagnoses themselves sampled and counted. |
| `native_placement_keeps_distinct_stale_diagnoses` | Mixed sheet: unlike failures keep separate entries; two identical runs produce byte-identical `reason`. |
| `native_placement_never_samples_the_expected_instance_list` | Sheet reused six times: the expected list is given in full, and two symbols missing a different instance past the fourth sorted path stay separate. |

The sample limits are asserted as literals (`PROMISED_IDENTITY_SAMPLE = 3`,
`PROMISED_DIAGNOSIS_SAMPLE = 5`) in the test module, not read from the implementation's
`STALE_IDENTITY_SAMPLE` / `STALE_DIAGNOSIS_SAMPLE`: they are the promise made to callers,
so a production change must fail the tests rather than move the oracle with it. The shared
`served_place_refused` helper asserts `error.kind`, the unchanged `error.target`, and both
files byte-identical, so every row above proves the unchanged target.

Fixtures are the KiCad-authored `complex_hierarchy` demo (`ampli_ht.kicad_sch`, 46
placed symbols). Corruption happens only in temporary copies, as the existing
`project_ownership` fixture README already sanctions.

### Negative controls

Each guard neutralized in turn, suite run, guard restored:

| Control | Neutralized | Result |
|---|---|---|
| A | Grouping bypassed — every symbol its own entry | `collapses_a_uniform_stale_diagnosis` fails |
| B | `join_bounded` names every item (`named = items.len()`) | `collapses_a_uniform_stale_diagnosis` fails |
| C | Group key forced constant — unlike diagnoses merge | `keeps_distinct_stale_diagnoses`, `bounds_the_number_of_distinct_diagnoses`, `never_samples_the_expected_instance_list` fail |
| D | `.take(STALE_DIAGNOSIS_SAMPLE)` removed | `bounds_the_number_of_distinct_diagnoses` fails |
| E | Instance-path cap reinstated at 4 | `never_samples_the_expected_instance_list` fails, showing both `must record exactly [... and 2 more]` and the two unlike failures merged into `RV201, C201 (2 symbols)` |
| F | Production `STALE_IDENTITY_SAMPLE` 3 → 4 and `STALE_DIAGNOSIS_SAMPLE` 5 → 6, tests untouched | `collapses_a_uniform_stale_diagnosis` and `bounds_the_number_of_distinct_diagnoses` fail — the two tests that spend the limits. The oracle no longer moves with the implementation. |
| G | `StaleTarget.target` set to a sibling path | all four new tests fail on the `error.target` assertion |

Control C is deliberately not `reasons.first()` — an earlier attempt at that bit
nothing, because the two corrupted symbols each carried a single reason. It forces the
grouping key constant instead, which genuinely merges groups.

## Risk and rollback

Low. The change is confined to how one diagnostic string is built; no control flow,
no file I/O, no schema. Rollback is reverting the single commit — no migration, no
persisted state, nothing downstream depends on the old text.

## Review checklist

- [x] The branch includes current `upstream/main` (240e295), has no merge conflicts, and CI ran on this exact head.
- [x] The PR shows only its unique commits and diff; there are no dependencies.
- [x] Every review conversation is resolved (none yet).
- [x] New names follow `docs/NAMING_CONVENTIONS.md`; nothing public was renamed.
- [x] New behavior and failure paths have regression coverage, with recorded negative controls.
- [x] File mutations are atomic and preserve unrelated content — this path writes nothing.
- [x] IPC mutations — not applicable, no IPC path touched.
- [x] No tools added or removed; counts and docs unchanged.


neusse:
Please make these two focused test corrections, then reply with the new full head SHA:

- Assert the externally promised sample limits as literal expectations (`3` identities and `5` diagnoses), rather than reading `STALE_IDENTITY_SAMPLE` / `STALE_DIAGNOSIS_SAMPLE` from the implementation under test. Otherwise changing production and test constants together silently changes the oracle, contrary to the reliability contract.
- In `served_place_refused`, assert the structured `error.target` remains the requested child schematic path. The PR currently says all four new tests prove the unchanged target, but the shared helper only checks `error.kind` and unchanged files.

The implementation otherwise matches #592: it keeps the stale-target guard, bounds and groups the diagnosis deterministically, preserves distinct failures, uses served dispatch tests, leaves files byte-identical, and correctly keeps the global `error_kind` duplication out of scope. The generic bounded-join helper and repeated fixture mutation are not merge blockers for this focused change.

Merge-train placement: #588 is the older overlapping `docs/API_MIGRATIONS.md` change and remains the next base-forming PR. Please make these corrections now, but expect one final refresh onto current `main` after #588 lands so this PR keeps only its unique commit and evidence.

pauliuszaleckas:
Both corrections are in. New full head SHA: `c80adf567fcf2047aa49b17eb91c4b61f97defb0`.

- **Sample limits as literal expectations.** The test module now declares
  `PROMISED_IDENTITY_SAMPLE = 3` and `PROMISED_DIAGNOSIS_SAMPLE = 5` with a comment
  recording that they are the promise made to callers, not a mirror of the code. The
  tests no longer reference `STALE_IDENTITY_SAMPLE` / `STALE_DIAGNOSIS_SAMPLE`; those
  are now used only by the implementation, so a production change fails the tests
  instead of moving the oracle with it.
- **`error.target` in the shared helper.** `served_place_refused` now asserts
  `error.target` equals the requested child schematic path, alongside `error.kind` and
  the two byte-identical files. All four new tests route through it, so the body's claim
  is now backed by an assertion.

Two negative controls for the corrections themselves, run on this head:

| Control | Neutralized | Result |
|---|---|---|
| F | Production `STALE_IDENTITY_SAMPLE` 3 → 4 and `STALE_DIAGNOSIS_SAMPLE` 5 → 6, tests untouched | `collapses_a_uniform_stale_diagnosis` and `bounds_the_number_of_distinct_diagnoses` fail — the two tests that spend the limits. Before the correction they passed. |
| G | `StaleTarget.target` set to a sibling path | all four new tests fail on the `error.target` assertion |

Controls A–E re-run unchanged on this head, same failures as recorded.

Local gate on `c80adf5` (Fedora, Rust stable):

- `cargo fmt --all -- --check` — pass, no output
- `cargo clippy --workspace --locked --all-targets -- -D warnings` — exit 0, 0 warnings
- `cargo test --workspace --locked --lib --tests` — exit 0, 0 failed
- `cargo test --workspace --locked --doc` — exit 0

CI on `c80adf5`: all 10 required checks green (3-OS Check & Test, Clippy, Format,
Schematic viewer, Plugin (Python), Nix flake, PCM packaging validation, Dependency
licenses).

The PR body now carries controls F and G and the note on why the limits are literals.

Understood on the merge train — no rebase now; I'll do the single refresh onto current
`main` once #588 lands, keeping this PR at its one unique commit.

neusse:
#588 is merged. Please make the one final refresh of this PR onto current `upstream/main` at `80fc5e05c391a2988f05dd5bc2e527be81953a99`, preserving only #605's unique commit and all `docs/API_MIGRATIONS.md` entries, then rerun the ten required checks and reply with the full new head SHA.

Your implementation corrections on `c80adf567fcf2047aa49b17eb91c4b61f97defb0` remain accepted. This is now the immediate successor in the schematic/documentation overlap set.

neusse (approved):
Maintainer refresh verified at exact head c1c105f3644c8f3b3f86306e1f475275506545af. The patch ID is identical to the previously reviewed change, it now applies cleanly to current main, focused native-placement tests pass locally, all ten required CI checks are green, and Closes #592 resolves correctly.

Issue #613, opened 2026-09-17
set_mirror leaves field text on the unreflected side — the #612 defect, waiting for the mirror tool in #450


pauliuszaleckas:
Heads-up for #450, which is in progress — not a claim on it. `Symbol::set_mirror` has the same defect #612 just fixed in `Symbol::set_rotation`, and the mirror tool #450 adds will be the first caller to reach it.

`Symbol::set_mirror` (`crates/konnect-schematic-editor/src/schematic/symbol.rs`) is one line of body:

```rust
pub fn set_mirror(&mut self, mirror: Option<&str>) {
    self.mirror = mirror.map(str::to_owned);
}
```

A field's `(at …)` is an absolute sheet coordinate, not an offset from the body, so reflecting the body leaves Reference and Value on the unreflected side. `Symbol::translate` carries fields, and `Symbol::set_rotation` now does too (#612); `set_mirror` is the remaining placement mutator that does not.

## Demonstrated

A `Device:LED` at (101.6, 50.8) with its fields on the library anchors — Reference 2.54mm above the origin, Value 2.54mm below:

```
before          Reference (101.6, 48.26)   Value (101.6, 53.34)
after mirror x  Reference (101.6, 48.26)   Value (101.6, 53.34)   ← unchanged
after rot 90    Reference (99.06,  50.8)   Value (104.14, 50.8)   ← control: rotation carries
```

`(mirror x)` negates screen-Y, so the two fields should swap about the origin: Reference to (101.6, 53.34), Value to (101.6, 48.26). That is exactly what `add_schematic_component` writes today when the same symbol is *placed* with `mirror: "x"`, since #485 pushes the library anchors through `PinTransform` — so the placement path is an oracle for the mutation path, the same way it was for #612.

Reproduce by appending to `symbol.rs`:

```rust
#[cfg(test)]
mod mirror_gap_probe {
    use super::*;
    use crate::types::Property;

    #[test]
    fn probe() {
        let mut sym = Symbol::new("Device:LED", 101.6, 50.8);
        let mut reference = Property::new("Reference", "D1");
        reference.sub_nodes.push(At::with_rotation(101.6, 48.26, 0.0).to_sexp());
        sym.properties.push(reference);
        sym.set_mirror(Some("x"));
        let at = sym.properties[0].sub_nodes.iter().find_map(At::from_sexp).unwrap();
        println!("{:?}", (at.x, at.y)); // (101.6, 48.26); should be (101.6, 53.34)
    }
[...]

PR #614, opened 2026-09-17
fix(schematic): carry field text round a rotated symbol

### LINKED ISSUE #612: rotate_schematic_component leaves a symbol's field text at the old orientation — designators land on the body
`rotate_schematic_component` turns the body and nothing else. A field's `(at …)` is an absolute sheet coordinate, not an offset from the symbol, so Reference and Value stay where the *previous* orientation put them.

`Device:LED` anchors its Reference 2.54mm above the origin and its Value 2.54mm below — clear of a horizontal body, straight through a vertical one. Place one flat, turn it to 90°, and both strings land on the body and the wires into its pins.

## Repro

```
add_schematic_component  lib_id=Device:LED x=330.2 y=74.93 rotation=0 reference=D1
rotate_schematic_component  reference=D1 rotation=90
```

## Measured with `kicad-cli sch export svg` (KiCad 10.0.6)

The invisible `<text>` elements carry each string's anchor; the drawn strokes give the body. For D1 at (330.2, 74.93) turned to 90°, the symbol axis is `x = 330.2`, with stroke vertices at `y = 71.12, 73.66, 76.2, 78.74` — pin, body, body, pin.

| | `D1` anchor | `Green` anchor | distance from the axis |
|---|---|---|---|
| after `rotate_schematic_component` | (330.20, 73.02) | (330.20, 78.11) | **0.00mm** |
| after `reset_schematic_field_positions` | (327.66, 75.56) | (332.74, 75.56) | 2.54mm |

At `font-size` 1.6933, `D1` (`textLength` 2.6827) spans 71.68..74.36 along that axis — the upper pin and the top of the body. `Green` (5.5856) spans 75.32..80.90 — the rest of the body and the whole lower pin.

## Not general to rotation

`add_schematic_component` has carried its anchors through the placement transform since #101, so a symbol *placed* at 90° is correct. On the board this came from, 8 symbols are rotated and only the two LEDs are wrong — the two placed flat and turned afterwards.

`Device:R`'s anchor is off to the side, so the same defect on a resistor reads as a slightly odd offset rather than a collision. `Device:LED` anchors above and below, where a 90° body goes.

## Cause

`Symbol::set_rotation` (`crates/konnect-schematic-editor/src/schematic/symbol.rs`) assigns `at.rotation` and returns. Its sibling `translate` explicitly moves every field with the body, because property coordinates are absolute; the rotation path never got the same treatment. `handle_rotate_schematic_component` is the only caller, so that one tool is the whole exposure.

## Suggested fix

Rotate each field's position about the symbol origin, reversing the sense when the body is mirrored (a placement rotates before it mirrors, and a reflection reverses any rotation conjugated by it). Leave the stored field *angle* alone — KiCad adds the symbol's rotation when drawing, as `field_at` documents.

`reset_schematic_field_positions` already repairs sheets written by earlier versions; it accounts for rotation.

pauliuszaleckas:
Closes #612

`Symbol::set_rotation` assigned `at.rotation` and returned. A field's `(at …)` is an absolute sheet coordinate, not an offset from the body, so Reference and Value stayed where the *previous* orientation had put them — the sibling `translate` carries fields explicitly for exactly this reason, and the rotation path never did.

`Device:LED` anchors its Reference 2.54mm above the origin and its Value 2.54mm below: clear of a horizontal body, straight through a vertical one. Place one flat, turn it to 90°, and both strings land on the body and the wires into its pins.

## Approach

The turn rotates each field's position about the symbol origin. It does **not** re-derive positions from library anchors — that would discard a caller's hand-placed field offset, and `reset_schematic_field_positions` is the tool whose job that is.

Two deviations from the issue's suggested fix, both narrowing it:

1. The issue said "reversing the sense when the body is mirrored". That is wrong for `(mirror xy)`, which `konnect_sexp::schematic` parses as *both* axes (`s == "xy"` sets `mirror_x` and `mirror_y`). Two reflections compose into a proper 180° turn, determinant +1, which does not reverse anything; `(mirror none)` is no reflection either. The predicate is axis parity — `axes.contains('x') != axes.contains('y')` — matching the `contains` convention `handle_reset_schematic_field_positions` already uses. Caught in review; `only_an_odd_number_of_reflections_reverses_a_turn` pins it.
2. Rather than a third copy of eeschema's rotation matrix, `konnect-sexp`'s `transform_pad` is split into a new `rotate_about(dx, dy, origin_x, origin_y, rotation_deg)` that `transform_pad` now delegates to argument-for-argument. That puts the schematic field rotation under the existing `pad_transform_rejects_textbook_rotation` guard, which exists because the textbook matrix agrees at 0°/180° and silently reflects end-for-end at ±90°. `transform_pin` was considered and rejected: it bakes in a Y-up→Y-down flip and applies the mirror as a reflection, neither of which this path may do.

`Symbol::map_field_positions` is extracted alongside; `translate` and the new `rotate_field_text` are both one-liners over it, so the next placement mutator has the traversal ready. (`set_mirror` is the remaining one that does not carry fields — latent today, filed separately as #613, since its only production caller places a symbol whose `properties` are still empty.)

The stored field *angle* is deliberately untouched: KiCad adds the symbol's rotation when drawing, so turning it here would draw the text at twice the angle. `field_at` already documents this and pins it with a rendered test.

## Validation

All run on the exact head, `1a5d06b`, against KiCad 10.0.5 on Fedora.

| Command | Result |
|---|---|
| `cargo test --workspace` | 38 suites, 0 failed |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo fmt --all --check` | clean |

**Not run:** the 3-OS matrix (Linux only here), the Nix flake, the plugin (Python), PCM packaging and dependency-license checks — CI covers those.

## Oracle

Expectations come from `kicad-cli sch export svg`, not from the crate under test. Its invisible `<text opacity="0">` elements carry each string's anchor, and the drawn strokes give the body.

Measured on a real board (a CH347T SWD debugger, 27 parts) whose D1 sits at (330.2, 74.93) turned to 90°. The symbol axis is `x = 330.2`, with stroke vertices at `y = 71.12, 73.66, 76.2, 78.74` — pin, body, body, pin:

| | `D1` anchor | `Green` anchor | distance from the axis |
|---|---|---|---|
| before | (330.20, 73.02) | (330.20, 78.11) | **0.00mm — on it** |
| after | (327.66, 75.56) | (332.74, 75.56) | 2.54mm |

At `font-size` 1.6933, `D1` (`textLength` 2.6827) spans 71.68..74.36 along that axis — the upper pin and the top of the body. `Green` (5.5856) spans 75.32..80.90 — the rest of the body and the whole lower pin.

The in-tree tests use a second, independent oracle: **placing** the symbol at the target angle outright, a path `rotated_symbol_carries_its_fields_around_with_it` already pins against eeschema's own numbers. Placing-rotated and rotating-after-placing must agree field-for-field. Each test also states its literal, so the two paths agreeing on a wrong point still fails.

## Negative controls

Each guard neutered in turn, full workspace suite run, then restored. Measured on `1a5d06b`:

| Control | Tests failed |
|---|---|
| Drop the field carry (the bug as it shipped) | 3 — `rotating_a_placed_symbol_carries_its_field_text_around`, `rotating_a_mirrored_symbol_turns_its_fields_the_other_way`, `turning_a_symbol_carries_a_hand_placed_field_offset_round` |
| Ignore the reflection (`turn = delta_deg`) | 1 — `rotating_a_mirrored_symbol_turns_its_fields_the_other_way` |
| Rotate the stored field angle too | 3 — the three above |
| Axis parity → presence check (`!axes.is_empty()`) | 1 — `only_an_odd_number_of_reflections_reverses_a_turn` |

An earlier revision of this table recorded control 3 as failing a fourth test,
`turning_a_symbol_back_restores_its_field_positions`. It does not:
that test walks 90 → 180 → 270 → 0 and compares the file byte-for-byte, so an additive
angle mutation cancels over the full turn (a normalized `rem_euclid(360.0)` variant was
re-run to confirm it is not an artefact of how the guard was neutered). The property is
pinned instead by `field_at`'s rendered test, which is what proves the stored angle must
not turn.

Control 4 is the one worth noting: reverting to the presence check leaves the entire `konnect-core` suite green and fails only the `konnect-schematic-editor` unit test. No tool-level test can reach it — `add_schematic_component` accepts only `"x"` or `"y"`, so it cannot construct `(mirror xy)`. That is why the regression test lives in the data-model crate.

## Risk and rollback

Behavioural change confined to `rotate_schematic_component`, the only production caller of `set_rotation`. `transform_pad` is byte-identical in behaviour (pure extraction; its own tests and the textbook guard still pass unchanged). Placement is untouched — it assigns `at.rotation` directly on a freshly built symbol and derives fields from anchors afterwards.

A sheet a previous version left mispositioned is not rewritten by this change; `reset_schematic_field_positions` repairs it, and already accounted for rotation.

Rollback is `git revert` of the single commit; no file format, argument or response key changed, so nothing on disk needs migrating either way.

## API_MIGRATIONS

The entry is appended at the **end** of the file rather than the top, so the diff against main deletes nothing and moves nothing — every entry already on main keeps its position and order. Flagging it because the file's convention is newest-first; happy to move it to the top if you would rather take the conflict.


neusse:
Please rebase this branch onto current `upstream/main` (`9d33cf0c239257f41a178cd26a2513d6e568d1c0`) and rerun the required checks.

The implementation is currently clean against GitHub's merge simulation, but its merge base is `c5f5a2254e3df37d151e1d2466b13863ca047581` and all ten CI results predate the v0.12.0 release/stamp commits now on `main`. Current governance requires the reviewed head to incorporate current main; a clean merge status does not make the old exact-head evidence current.

This is the next schematic-rotation PR in the queue. Keep its one unique commit and focused diff. After it is refreshed, we can complete review before #626/#627 changes the same rotation/reconciliation area.

pauliuszaleckas:
Rebased onto `9d33cf0c239257f41a178cd26a2513d6e568d1c0`. Head is **1a5d06b72253acc31e86928b57074bb80de5fe3c** — still one commit, force-pushed once. Merge base is now that SHA, so the v0.12.0 release/stamp changes are incorporated.

**No conflict.** Main advanced only by `ee9e3cc` (prepare v0.12.0) and `7cd1a42` (stamp v0.12.0 PCM), which touch `Cargo.toml`/`Cargo.lock`, the viewer manifests and `packaging/metadata.json` — none of this branch's files. The change is byte-identical to `1f4a995`: `git diff 1f4a995 1a5d06b` over this commit's own files is empty. Every `docs/API_MIGRATIONS.md` entry is preserved — main carries 29, this head 30, diff adds 29 lines and deletes none, with this entry still appended at the end as flagged.

**Negative controls, re-derived on this head** — each guard neutered, the full `--workspace --lib --tests` suite run, then restored:

| Control | Tests failed |
|---|---|
| Drop the field carry (the bug as it shipped) | 3 — `rotating_a_placed_symbol_carries_its_field_text_around`, `rotating_a_mirrored_symbol_turns_its_fields_the_other_way`, `turning_a_symbol_carries_a_hand_placed_field_offset_round` |
| Ignore the reflection (`turn = delta_deg`) | 1 — `rotating_a_mirrored_symbol_turns_its_fields_the_other_way` |
| Rotate the stored field angle too | 3 — the three above |
| Axis parity → presence check (`!axes.is_empty()`) | 1 — `only_an_odd_number_of_reflections_reverses_a_turn` |

**One correction to the PR body, and it is mine, not a change on this head.** The description records control 3 as failing 4 tests, "the three above plus `turning_a_symbol_back_restores_its_field_positions`". Re-deriving it here it fails 3: that roundtrip test walks 90 → 180 → 270 → 0 and compares the file byte-for-byte, so an *additive* angle mutation cancels over the full turn and the test does not bite. I re-ran a normalized variant (`(angle + turn).rem_euclid(360.0)`) to be sure it was not an artefact of how I neutered it — also 3. The control still does its job; the fourth failure was overstated and the body is corrected. The property itself stays pinned by `field_at`'s rendered test, which is what proves the stored angle must not turn.

**Oracle unchanged.** `kicad-cli sch export svg` on the CH347T board's D1 at (330.2, 74.93) turned to 90°: `D1` and `Green` anchors move off the symbol axis from (330.20, 73.02)/(330.20, 78.11) — 0.00 mm from it — to (327.66, 75.56)/(332.74, 75.56), 2.54 mm clear. The in-tree tests keep their second oracle: placing at the target angle outright, which must agree field-for-field with rotating after placing, each with its literal stated.

Local gate on this head: 1910 lib/tests passed / 0 failed, doctests clean, `clippy --workspace --locked --all-targets -D warnings` clean, `fmt --check` clean, plus the 20 `schematic-viewer` tests `--workspace` skips.

**CI: all ten required checks green on this head** — run 35312033745.

`Closes #612` still resolves. I'll leave the branch alone from here.

neusse:
Please replace the hand-authored schematic strings in the new mutation regressions with a KiCad-authored fixture, then rerun the focused negative controls and all ten checks on the new head.

The refreshed implementation and #612 behavior otherwise review clean, but `sheet_with_library`, `LED_DEFINITION`, `OFF_AXIS_DEFINITION`, and the manually assembled moved-field sheet exercise mutation against approximations written by the test itself. `docs/RELIABILITY_CONTRACT.md` requires KiCad-authored fixtures for design parsing and mutation. Use a real KiCad 10.0.6-resaved sheet containing the needed LED/off-axis/mirrored/manual-field cases; controlled string surgery may introduce only the specific test delta. Preserve the independent SVG/literal oracle and the existing negative controls.

This remains the next rotation PR after #588. #613 stays separate as intended.

pauliuszaleckas:
Reworked onto a KiCad-authored fixture and reconstructed as one commit on current `main` (`c03a13d`). Head is **ecdb703ef29593c7f947012c72c45af8cd5483e6**.

`rotate_fields_kicad10.kicad_sch` replaces `sheet_with_library`, `LED_DEFINITION`, `OFF_AXIS_DEFINITION` and the hand-assembled moved-field sheet. It is built through Konnect against the stock `Device` and `Regulator_Linear` libraries, then force-resaved by KiCad 10.0.6; the resave changed the file's shape and left every field position untouched. The sheet is built in twins — each symbol appears unrotated and again placed at 90° outright 25.4mm below — so a turn's expectation comes from KiCad's own file, not from a second run of our placement path. `kicad-cli sch export svg` corroborates each one, and because a symbol and its twin share mirror, field angle and justification, the comparison carries no constant:

| Reference field | after the turn | twin | difference |
|---|---|---|---|
| `D1` | (99.0600, 51.4349) | `D2` (99.0600, 76.8349) | (0, 25.4000) |
| `U1` (`mirror x`) | (133.9850, 47.7362) | `U2` (133.9850, 73.1362) | (0, 25.4000) |
| `U3` (plain) | (172.0850, 55.1338) | `U4` (172.0850, 80.5338) | (0, 25.4000) |

Six tests, up from four. New: an unmirrored control (`U3`/`U4`), which fails if the `(mirror x)` token is ignored, and a case for the boundary #642 created — `R1` stands across the `NETJ` wire with a no-connect on pin 2, and one turn carries the fields, adds the junction and moves the marker. `kicad-cli sch export netlist` is the oracle there: `/NETJ` is `R2.1` as committed and `R2.1, R1.1` after the turn.

Negative controls, re-derived on this head — each guard neutered, `--workspace --lib --tests` run, then restored:

| Control | Tests failed |
|---|---|
| Drop the field carry (the bug as it shipped) | 5 |
| Ignore the reflection (`turn = delta_deg`) | 1 — `turning_a_mirrored_symbol_turns_its_fields_the_other_way` |
| Rotate the stored field angle too | 5 |
| Axis parity → presence check (`!axes.is_empty()`) | 1 — `only_an_odd_number_of_reflections_reverses_a_turn` |

Two things I would rather flag than have you find:

The round-trip test no longer compares against the fixture bytes. The first write reserializes KiCad's formatting into Konnect's, so its baseline is the sheet after a no-op turn. It still proves no drift through 90/180/270/0, but it is a weaker claim than before.

`only_an_odd_number_of_reflections_reverses_a_turn` stays in `symbol.rs`. It builds a `Symbol` through `Symbol::new`/`Property::new` with no parsed schematic at all, so it is a geometry unit test rather than mutation against an approximation, and it matches the pre-existing `move_to_carries_property_text_along` beside it. Happy to move it onto the fixture if you read the contract as covering it, at the cost of the control that currently catches the parity bug.

Local gate on this head: `cargo fmt --all -- --check` clean, `cargo clippy --workspace --locked --all-targets -- -D warnings` clean, `cargo test --workspace --locked --lib --tests` 2013 passed / 0 failed, doctests clean. `crates/schematic-viewer` is untouched and was not run. Every `docs/API_MIGRATIONS.md` entry is preserved — the diff against main adds 28 lines and deletes none.

Issue #626, opened 2026-09-17
A placement change orphans a no-connect flag and can then wire the pin it protected into a net


pauliuszaleckas:
A no-connect flag is a sheet item at a coordinate, so a placement change leaves it where it
was while the pin it was attached to moves away. The reconciliation pass then sees no marker
at the pin's new position, and — if that position is mid-span on a wire — adds a junction
dot, wiring in a pin the user had explicitly declared unconnected.

## Demonstrated

KiCad 10.0.6 on Fedora, Konnect 0.12.0 built from `main` @ 9d33cf0 with #615's branch on top.
Both paths behave identically, so this is the shared contract rather than one tool's bug.

`Device:R` at `(158.75, 160.02)` has pin 1 at `(158.75, 156.21)`, clear of a wire running
`(144.78, 160.02) -> (156.21, 160.02)` labelled NETB.

**Move** (`main` behavior, unchanged since #120):

```
add_no_connect            (158.75, 156.21)      # on R2 pin 1
move_schematic_component  R2 -> (154.94, 163.83)  # pin 1 lands mid-span on the NETB wire
```

```json
{"moved": "R2", "junctions_added_count": 1, "junctions_pruned_count": 0}
```

**Turn** (same result once #615 lands):

```
add_no_connect             (158.75, 156.21)
rotate_schematic_component R2 -> 90             # pin 1 lands at (154.94, 160.02), mid-span
```

```json
{"rotated": "R2", "junctions_added_count": 1, "junctions_pruned_count": 0}
```

In both cases the marker is still in the file at `(158.75, 156.21)`, orphaned, and the pin is
now on NETB.

The `has_no_connect` guard in `reconcile_junctions_at`
(`crates/konnect-core/src/tools/sch_wiring.rs`) is doing what it says — its doc comment calls
a no-connect "the user saying this pin stays unconnected, which a move must not override" —
but it only looks at the *arrival* point, and the marker is at the *vacated* one. So the
guard misses exactly the case it exists for.

## Not silent, but not obvious either

`kicad-cli sch erc --severity-all` reports 1 `no_connect_dangling` for the orphaned marker,
so a caller who runs ERC afterwards has a signal. What ERC does not say is that the pin the
marker used to protect has joined a net.

## The question to settle

Two defensible answers, and they differ for other tools too:

1. **Carry the marker with the pin.** A no-connect belongs to the pin, so a placement change
   relocates it. `handle_delete_schematic_component` already has the machinery for finding
[...]

Issue #625, opened 2026-09-17
replace_component relocates pin endpoints and reconciles no junctions


pauliuszaleckas:
`handle_replace_component` (`crates/konnect-core/src/tools/sch_components.rs:4269`) swaps a
placed symbol's `lib_id` and writes through `write_atomic_if_unchanged`. A different symbol
has different pin offsets, so the replacement relocates pin endpoints exactly as a move or a
turn does — but no junction reconciliation runs and no junction counts are reported.

This is the #615 failure mode with a third trigger: there, a turn moved the pins; here, the
part under them changed shape.

## Demonstrated

KiCad 10.0.6 on Fedora, Konnect 0.12.0 built from `main` @ 9d33cf0.

`Device:R` stands vertically with its pins at `y ∓ 3.81`. `Device:LED` lies horizontally
with its pins at `x ∓ 3.81`. So replacing one with the other at the same origin moves both
pins:

```
add_wire                  (114.3, 97.79) -> (139.7, 97.79)   # labelled NETA
add_schematic_component   Device:R as R1 at (160.02, 101.6)
move_schematic_component  R1 -> (127, 101.6)                 # pin 1 lands mid-span; dot added
replace_component         R1 -> Device:LED
```

`get_schematic_pin_locations` before and after, which is the tool's own answer:

| | pin 1 | pin 2 |
|---|---|---|
| `Device:R` | (127, 97.79) — **on the wire** | (127, 105.41) |
| `Device:LED` | (123.19, 101.6) | (130.81, 101.6) |

The response says nothing about it:

```json
{"reference": "R1", "old_lib_id": "Device:R", "new_lib_id": "Device:LED", "units_replaced": 1, "unit": null}
```

and the file still carries `(junction (at 127 97.79))`, now with only the wire passing
through it. `kicad-cli sch export netlist` confirms the part has left the net entirely:

```
unconnected-(R1-K-Pad1)    R1.1
unconnected-(R1-A-Pad2)    R1.2
```

`/NETA` is absent from the netlist — the label names a wire nothing is on.

The converse direction is the serious one and follows from the same arithmetic: replace a
part whose pins are clear of a wire with one whose pin lands mid-span on it, and the
connection is missing from the netlist while the sheet draws as connected.

## Suggested fix

The same treatment the move and (after #615) the turn get: snapshot pin endpoints before the
write, reconcile at the symmetric difference afterwards, report `junctions_added_count` /
[...]

Issue #624, opened 2026-09-17
Three copies of the placement-change junction reconciliation harness, in two files


pauliuszaleckas:
Not a defect — a maintenance note found while fixing #615, worth recording before a fourth copy appears.

"Snapshot pin endpoints, mutate, diff the two sets, reconcile junctions at the difference" now exists three times:

1. `crates/konnect-core/src/tools/sch_components.rs` — `pins_before_placement_change` + `reconcile_junctions_after_placement`, called by `move_schematic_component` and (after #615) `rotate_schematic_component`. Both halves re-read the file from disk; the reconciler reads and parses it once for `after_pins` and then again for the content it edits.
2. `crates/konnect-core/src/tools/sch_batch.rs:955` — `bulk_move_schematic_components` reimplements `pins_of`, the `has_wires` short-circuit and `differs` inline, then calls the shared `reconcile_junctions_at`. This copy works on the strings it already holds, so it costs no extra reads at all.
3. The rules themselves live in `sch_wiring::reconcile_junctions_at`, which all of the above share — that part is fine, and its doc comment is explicit that it must not become "a fifth private definition of what is attached here" (#323).

So the shared thing is the junction *rules*; what is duplicated is the pin-snapshot-and-diff harness around them. The costs:

- The `"(wire"` short-circuit, the coincidence tolerance and the symmetric-difference assembly are each written twice, in two files, and can drift silently — the compiler cannot see that they are meant to agree.
- The two implementations already differ in an interesting way: `sch_components.rs` re-reads the file up to three times per call, while `sch_batch.rs` diffs the before and after strings it already has in memory. The in-memory one is the better shape.
- Any handler that acquires the same need — #623 (`move_region`) and #622 (`batch_place_components`) are both open — has to choose which copy to imitate.

## Suggested direction

One helper, taking the before and after *content* rather than a path:

```rust
fn reconcile_junctions_for_placement(before: &str, after: String) -> (String, usize, usize)
```

`bulk_move` already has both strings, and the `sch_components.rs` handlers can hand it the source they load and the source they wrote, dropping the extra reads rather than paying them twice. The `"(wire"` short-circuit and the tolerance then live in exactly one place, and a new caller gets the whole contract by calling one function.

Low priority: no user-visible symptom, and the copies agree today. Worth doing when one of #622 / #623 is picked up, since either would otherwise add a fourth site.

pauliuszaleckas:
[...]

Issue #623, opened 2026-09-17
move_region relocates pins and reconciles no junctions, in either direction


pauliuszaleckas:
`handle_move_region` (`crates/konnect-core/src/tools/sch_components.rs:3427`) translates every symbol whose origin falls inside a rectangle, calls `sch.overwrite()`, and returns `moved_count` / `moved` / `placements`. It never reconciles junction dots and reports no junction counts, although a region move relocates pin endpoints exactly as `move_schematic_component` does — and that handler has taken the same pins through `reconcile_junctions_after_move` since #120.

#120 named `move_schematic_component`, `move_labels_by_offset` and `bulk_move_schematic_components`. `move_region` was not on that list and was never covered; `bulk_move_schematic_components` does reconcile today, so the region move is now the odd one out among the multi-symbol movers.

## Demonstrated

KiCad 10.0.6 on Fedora, Konnect 0.12.0 built from `main` @ 9d33cf0 with the #615 branch on top; that branch does not touch this handler (its `sch_components.rs` hunks are confined to the move and rotate handlers, the shared snapshot helper, and tests).

`Device:R` standing at 0° puts its pins at `y ∓ 3.81`.

**A dot the region move strands.** `R1` is moved onto the wire by `move_schematic_component`, which adds the dot itself (`junctions_added_count: 1`), and then a region move carries it away:

```
add_wire                  (114.3, 97.79) -> (139.7, 97.79)     # labelled NETA
add_schematic_component   Device:R as R1 at (160.02, 101.6)
move_schematic_component  R1 -> (127, 101.6)                   # pin 1 lands mid-span; dot added
move_region               box (120, 95)-(134, 108), dy +12.7   # R1 -> (127, 114.3)
```

```json
{"moved": ["R1"], "moved_count": 1, "moved_unit_count": 1, "placements": [{"reference": "R1", "unit": 1, "x": 127.0, "y": 114.3}]}
```

No junction keys, and the file still carries `(junction (at 127 97.79))` with nothing but the wire passing through it. KiCad's netlist confirms R1 is gone from the net: `unconnected-(R1-Pad1)`, `unconnected-(R1-Pad2)`.

**A dot the region move never writes.** The serious direction — the same sheet, with the symbol below the wire and the region move bringing it up onto it:

```
add_wire                  (114.3, 97.79) -> (139.7, 97.79)     # labelled NETA
add_schematic_component   Device:R as R2 at (127, 114.3)
[...]

PR #627, opened 2026-09-17
fix(schematic): reconcile junctions when a symbol turns

### LINKED ISSUE #615: rotate_schematic_component never reconciles junctions, while move_schematic_component does
`handle_move_schematic_component` calls `reconcile_junctions_after_move` and reports `junctions_added_count` / `junctions_pruned_count`. `handle_rotate_schematic_component` calls neither and reports neither, although a turn relocates a symbol's pin endpoints exactly as a move does.

The move handler's own comment says why it reconciles — "wires do not change here — pins do". That is equally true of a rotation.

## Demonstrated

KiCad 10.0.6, Konnect 0.11.1 built from `main` @ c5f5a22. `Device:R` is vertical at 0° with pins at ±3.81mm in Y, horizontal at 90° with pins at ±3.81mm in X.

```
create_schematic          junction.kicad_sch
add_wire                  (114.3, 97.79) -> (139.7, 97.79)
add_schematic_component   Device:R as R1 at (160.02, 101.6)     # off to the side
move_schematic_component  R1 -> (127, 101.6)                    # pin 1 lands mid-span
rotate_schematic_component R1 -> 90
```

The move does the right thing:

```json
{"moved": "R1", "junctions_added_count": 1, "junctions_pruned_count": 0}
```

The rotate response carries no junction keys at all, and the dot is still in the file afterwards:

```
(junction
    (at 127 97.79)
```

`get_schematic_pin_locations` on the turned symbol confirms nothing of R1 is there any more:

| | pin 1 | pin 2 |
|---|---|---|
| before the turn (0°) | (127, 97.79) — **on the wire** | (127, 105.41) |
| after the turn (90°) | (123.19, 101.6) | (130.81, 101.6) |

So the junction at (127, 97.79) now has only the wire passing through it — two collinear segments and nothing else. It is unjustified and should have been pruned, exactly as the move path prunes one the pins vacate.

## Not caught downstream

`run_erc --severity all` on that sheet returns the expected `pin_not_connected` / `wire_dangling` violations for the toy geometry, and nothing at all about the stranded dot. So it survives into the saved file unreported.

## The converse case too

Pruning is only half of it. A turn can just as easily put a pin *onto* a wire mid-span — rotate a symbol next to a bus and a pin lands on it — which needs a junction added, or the connection is silently absent from the netlist. That direction is the one that changes connectivity rather than leaving cosmetic litter, so it is the more serious half.

## Suggested fix

`handle_rotate_schematic_component` already loads the schematic and knows the affected symbols. It needs the same `before_pins` snapshot / `reconcile_junctions_after_move` call the move handler makes, and the same two counts in its response.

The helper is named for moves but is position-based, so it should apply unchanged; a rename to something placement-neutral would fit if the mirror tool in #450 ends up needing it too.

Related: #612 / #614 fixed the field-text half of "a turn is a placement change with consequences"; junction state is the remaining gap in making a turn equivalent to placing the symbol at that angle.

pauliuszaleckas:
`handle_move_schematic_component` snapshots its pin endpoints, reconciles junction
dots at the points the pins left and arrived at, and reports the two counts.
`handle_rotate_schematic_component` did neither, although a turn relocates pin endpoints
exactly as a move does — the move handler's own comment says why it reconciles: "wires do
not change here — pins do".

The turn now takes the same snapshot, calls the same reconciliation after the write, and
reports the counts it returns. Both directions are covered: the dot a pin turns off is
pruned, and a pin that turns onto a wire mid-span gains the dot it needs.

Closes #615

## The two directions are not equally serious

KiCad is the oracle for that, not this crate. `kicad-cli sch export netlist` on the
fixture, before and after each turn, with "no fix" meaning the same turn with the
reconciliation bypassed:

| File | `/NETA` | `/NETB` |
|---|---|---|
| the fixture | R0.1, R1.1 | R4.1 |
| `R1` → 90°, pin off the wire | R0.1 | R4.1 |
| `R1` → 90°, no fix | R0.1 | R4.1 |
| `R2` → 90°, pin onto the wire | R0.1, R1.1 | **R2.1, R4.1** |
| `R2` → 90°, no fix | R0.1, R1.1 | R4.1 |

The last pair is the load-bearing row: with no dot written, KiCad reports the pin — mid-span
on the wire, at the coordinate KiCad itself gives for it — as `unconnected-(R2-Pad1)`. That
is the half that changes the design rather than the picture. `kicad-cli sch erc
--severity-all` agrees: 10 violations with the fix, 12 without (the pin stays
`pin_not_connected` and NETB stays `isolated_pin_label`).

The `R1` rows are the converse and agree with each other: a stranded dot costs no
connectivity, and ERC reports 14 violations either way — which is why it needs a fixture to
be caught at all.

## Fixture

`crates/konnect-core/tests/fixtures/rotate_junctions_kicad10.kicad_sch` is KiCad 10.0.6's
own serialization. Built through Konnect against the stock `Device` library —
`create_schematic`, `add_wire`, `add_schematic_component`, `move_schematic_component`,
`add_schematic_net_label` — then `kicad-cli sch upgrade --force`, and that resave is what
is committed: `(generator "eeschema")`, `(version 20260306)`, KiCad's own library records,
field positions and UUIDs.

The dot at `(127, 97.79)` is not hand-placed. `R1` was added off to the side and *moved*
onto the wire, and the move path wrote the dot itself, reporting
`junctions_added_count: 1` — it is the reconciliation the turn used to skip, left in the
file by the tool that does it. The dot at `(190.5, 88.9)` is `add_wire`'s, inserted where a
branch lands mid-span.

`rotate_junctions_kicad10.README.md` carries the provenance, the case table, and both
oracle tables.

## Validation

Every command below was run on this branch's head, `11cd960964d62e5076e9c92baf5efba17d4f9f61`.

| Command | Result |
|---|---|
| `cargo clippy --workspace --locked -- -D warnings` | clean |
| `cargo test --workspace --locked --lib --tests` | 31 binaries, 0 failures |
| `cargo test --workspace --locked --doc` | 0 failures |
| `kicad-cli sch export netlist` (KiCad 10.0.6) | the table above |
| `kicad-cli sch erc --severity-all` (KiCad 10.0.6) | the counts above |

Not run: `crates/schematic-viewer` — untouched by this change (CI runs it regardless), and
the `#[ignore]`d live-KiCad tests, which need a running editor.

## Negative controls

Each guard neutered in turn, the whole suite run with `--no-fail-fast`, then restored:

| Guard neutered | Tests failing |
|---|---|
| the turn's `reconcile_junctions_after_placement` call → `(0, 0)` | 3 — both direction tests and the served-dispatch test |
| `pins_before_placement_change` → always empty | 1 — the prune test (an empty snapshot makes every pin a candidate, so the add case survives) |
| `!idx.has_no_connect(px, py)` in `reconcile_junctions_at` | 3 — the new no-connect test plus 2 pre-existing |

Worth naming: the second control shows nothing in the existing suite covers the *move*
path's prune direction through that snapshot either.

## Tests

- `rotate_junction_reconciliation_tests` (`sch_components.rs`) — prune, add, the no-connect
  guard, a pin leaving a wire *end* (neither added nor pruned, counts still reported), and a
  refused turn that writes nothing: `stale_target`, file byte-identical to the fixture.
- `rotate_junction_dispatch_tests` (`mcp/handler.rs`) — the two counts across the served
  `tools/call` boundary, alongside the existing `*_dispatch_tests` modules, with the dot
  verified in the file afterwards.

## Deviation from the issue's suggested approach

The issue suggested the turn take "the same `before_pins` snapshot / `reconcile_junctions_after_move`
call the move handler makes", and floated a placement-neutral rename. Both are done, plus one
step further: rather than copy the snapshot block into a second handler, it is extracted as
`pins_before_placement_change` and called from both. That also collapses a double read — the
old block read the file to test for `(wire`, dropped the string, and then `pin_endpoints_of`
read it again.

`reconcile_junctions_after_move` → `reconcile_junctions_after_placement` is a private
function in one file; no public name changes.

## API_MIGRATIONS

One entry appended, describing the two added response keys and what a sheet written by an
older version needs. Placed after the entries already on `main`, in their original order, so
the file's diff deletes nothing and moves nothing.

## Risk and rollback

The blast radius is `rotate_schematic_component` plus the shared snapshot helper that
`move_schematic_component` also calls. The reconciliation rules themselves are untouched —
this adds a second caller to code the move path has run since #120. A sheet with no wire
takes the same early return as before.

Rollback is the single commit; sheets already written are unaffected, since the change only
governs what a *future* turn writes.

## Not in scope

A review pass over this branch turned up four neighbouring gaps. None is fixed here; all are
filed, and two are worth naming explicitly because they touch the contract this PR extends:

- **#626 — a placement change orphans a no-connect flag.** The `has_no_connect` guard in
  `reconcile_junctions_at` only looks at the pin's *arrival* point, while the marker stays at
  the *vacated* one, so a pin the caller declared unconnected can be wired into the net it
  lands on. Measured on both paths: `move_schematic_component` does exactly the same on
  `main` today, so this is the shared contract since #120, not something the turn introduces —
  and parity with the move is what this issue asked for. ERC does report the orphaned marker
  (`no_connect_dangling`), but says nothing about the pin joining a net. Deciding whether the
  marker should travel with the pin, or merely suppress the add, changes both handlers and
  belongs in its own change.
- **#625 — `replace_component` relocates pin endpoints and reconciles nothing.** The same
  failure mode with a third trigger: swapping `Device:R` for `Device:LED` moves both pins,
  strands the dot, and drops the part off the net with no junction keys in the response. It is
  the remaining caller that wants the helper this PR generalizes.
- #622 (`batch_place_components` never makes the junction pass its sibling's comment promises)
  and #623 (`move_region` reconciles nothing in either direction) are the same omission in two
  more handlers; #624 tracks the three copies of the snapshot-and-diff harness.

One known gap this PR does not close: `reconcile_junctions_after_placement` runs *after*
`sch.overwrite()`, and its read and write can fail, in which case the turn is already on disk
but the caller gets a bare `handler_error` rather than `mutation_outcome_uncertain`. That is
pre-existing on the move path; this adds a second site rather than a new class of problem, and
converting both would change the move path's error contract.

#613's mirror case waits on the mirror tool in #450.


neusse:
Please hold this PR behind #626.

The junction work is well evidenced, but the current shared reconciliation contract leaves a no-connect marker at the vacated coordinate and then adds a junction where the protected pin arrives. On the move path that already wires an explicitly unconnected pin into a net; this PR would add the same unsafe behavior to rotation. That is a material safety property, not validation debt.

The least-resistance order is:

1. settle and implement the pin-owned no-connect contract in #626;
2. update/rebase this PR so rotation uses that corrected shared contract;
3. rerun the exact-head checks and the real-KiCad netlist/ERC evidence;
4. complete final review of this PR;
5. consolidate the reconciliation harness in #624 before adding #622, #623, and #625 callers.

Keep this PR focused on #615, but it must consume the corrected shared contract rather than knowingly extending the unsafe one.

pauliuszaleckas:
CI note, so the red run in this PR's history is not mistaken for a real break.

The first run's `Check & Test` matrix failed on `ubuntu-latest` in `tools::cli::erc_parse_tests::erc_report_never_touches_the_project_directory`:

```
Failed to spawn kicad-cli: /tmp/.tmpFQqCNB/erc-success
Caused by:
    Text file busy (os error 26)
```

That is a pre-existing race in test support, not this change. `test_support::write_script` (`crates/konnect-core/src/tools/cli.rs:480`) writes a fake `kicad-cli` and the test execs it immediately; a child forked by another test thread while the write handle was open inherits a duplicate of it, and `ETXTBSY` stands until that child execs. Nothing in this diff spawns a process or touches `cli.rs` — it only shifts test scheduling. Filed separately as #630, with the mechanism and why a rename-into-place does not fix it.

The **same commit**, re-triggered with no change (close/reopen, since re-running a job needs admin), is green on all ten required checks — including the ubuntu job that had failed:

| Check | |
|---|---|
| Check & Test (ubuntu-latest) | pass 2m43s |
| Check & Test (macos-latest) | pass 3m28s |
| Check & Test (windows-latest) | pass 5m48s |
| Clippy · Format | pass |
| Schematic viewer | pass 1m45s |
| Plugin (Python) · PCM packaging validation · Dependency licenses | pass |
| Nix flake | pass 6m37s |

Head is `11cd960964d62e5076e9c92baf5efba17d4f9f61`, the same SHA the validation section of the description quotes; `macos-latest` and `windows-latest` in the failing run were cancelled by the fail-fast matrix rather than failing a test.

A review pass over this branch also turned up four neighbouring gaps in the same reconciliation contract, none of them fixed here: #625 (`replace_component` relocates pin endpoints and reconciles nothing) and #626 (a placement change orphans a no-connect flag, which `move_schematic_component` does identically on `main` today) are the two that touch this contract directly; #622, #623 and #624 are the same omission in two more handlers plus the duplicated harness. They are named in the description's *Not in scope* section.

pauliuszaleckas:
Understood — holding this behind #626, and I agree with the reasoning: extending the reconciliation contract to the turn while that contract can wire an explicitly unconnected pin into a net would spread a safety defect rather than contain it. The branch stays where it is; no rebase and no further evidence runs until the corrected contract exists.

Reading back your direction on #626 so the dependency is unambiguous before I build against it: a no-connect is intent attached to a pin identity (symbol instance/unit plus pin number, not coordinate coincidence); the marker moves in the *same* atomic mutation as the symbol, reconciliation then runs against the resulting content, and a carried marker at the arrival point suppresses junction creation; a mutation that cannot map the protected pin one-to-one refuses before writing rather than orphaning the marker or wiring the pin; and the evidence is save/reopen netlist + ERC, no-write-on-refusal, served dispatch, and a negative control.

Once that lands I'll rebase this PR onto it, drop rotation's own arrival-point-only handling in favour of the shared contract, and rerun the exact-head checks plus the real-KiCad netlist/ERC evidence before asking for final review. The consolidation in #624 and the #622/#623/#625 callers stay after this, in the order you set.

neusse:
Reconstructed as #642 now that the required #626/#641 safety contract is on main. The implementation commit, authorship, real-KiCad fixture, and evidence are preserved; the conflict resolution combines rotation reconciliation with atomic no-connect carry and adds one integration assertion for that boundary. Closing this pre-dependency copy as superseded so #615 has one current closing path. Thank you, @pauliuszaleckas.

Issue #597, opened 2026-09-16
bug(placement): refine_placement_force_directed scrambles flagged components without converging or improving the score


Reid-n0rc:
## Summary

`refine_placement_force_directed`, scoped via `references` to a specific set of already-flagged components (courtyard overlaps + off-board parts, 247 of 538 footprints) with the rest of the board passed as `locked`, produced a dry-run plan that displaces components by tens of millimeters — including several IC-adjacent decoupling/interface capacitors moved 20-plus mm away from the connector or IC they decouple — without converging (`"converged": false` after the full 300-iteration ceiling) and **without improving the board's own score** (`score_before` and `score_after_plan` both `0`/`hard_fail`).

## Environment

- Konnect 0.11.1 (commit `1db14b9`)
- KiCad 10.0.6, macOS (aarch64)
- Same board as the linked issues: 8-layer, 538 footprints, 96.7 × 87 mm outline

## Repro

```
refine_placement_force_directed(
  board="...",
  dry_run=true,
  references=[<247 refs already flagged by score_placement's courtyard_overlap
               and outside_outline deductions>],
  locked=[<80 refs belonging to the board's power-supply and DDR/DRAM
           subsystems, explicitly excluded from this refinement>]
)
→ {"converged": false, "iterations_run": 300, "score_after_plan": 0,
   "verdict_after_plan": "hard_fail", "planned_moves": [ ... 247 entries ... ]}
```

Selected planned moves (all verbatim from the response):

| Reference | From | To | Distance |
|---|---|---|---|
| `NT2101` | (27.621, 41.861) | (52.0, 115.5) | 77.6 mm |
| `NT2102` | (106.499, 79.991) | (52.0, 104.0) | 59.6 mm |
| `C201` | (18.404, 63.257) | (72.5, 86.0) | 58.7 mm |
| `D701` | (110.527, 72.005) | (67.0, 108.5) | 56.8 mm |
| `U2201` (AK5707 audio ADC decoupling context) | (46.837, 82.031) | (63.5, 101.5) | ~24 mm, moved away from its own connector `J2101` |
| `TP701`–`TP705` (test points) | clustered at (84–100, 89–95) | scattered to (49–79, 78–93) | 15–35 mm each, no longer clustered |

[...]

Issue #596, opened 2026-09-16
bug(placement): place_decoupling_caps matches candidates by ANY shared net (including GND), sweeping unrelated caps into a nonsensical off-board row


Reid-n0rc:
## Summary

`place_decoupling_caps` selects candidate capacitors by "shares at least one named net with the IC" (per its own tool description: *"Caps are identified by NET PAIRING — a candidate shares at least one named net with the IC — never by reference guessing"*). On a real board, GND is shared by nearly every capacitor, so calling this tool on an IC that has a GND pin sweeps in essentially every decoupling cap on the entire board as a "candidate," regardless of which chip they actually decouple.

The planner then lays out all matched candidates as a single evenly-spaced row beside the IC. With ~190 candidates matched (because of the GND overlap) and a fixed spacing step, the row runs far past both board edges.

## Environment

- Konnect 0.11.1 (commit `1db14b9`)
- KiCad 10.0.6, macOS (aarch64)
- Board: 8-layer, 538 footprints, Edge.Cuts outline x 13.8–110.5 mm, y 41.9–128.9 mm (96.7 × 87 mm)

## Repro

```
place_decoupling_caps(board="...", ic_reference="U201", dry_run=true)
```

`U201` is an SoC (AM62A) with pins on `GND`, `VDD_CORE`, `+1V1`, `LPDDR4_+1V8`, and others. The response's `planned_moves` contains **192 entries** — essentially every capacitor on the board sharing a net with `U201`, dominated by `GND` — laid out along `y ≈ 104.3–105.5` at `x` values stepping from **-90.959 to +205.897**, a span of ~297 mm on a board that is 96.7 mm wide.

`score_after_plan` and `score_before` are both `0`, `verdict_after_plan` is `hard_fail` — the plan does not even improve the board's own scoring metric; it just relocates ~190 components, most of which decouple unrelated ICs elsewhere on the board (audio codec, PMIC, WiFi/SD connectors, Ethernet PHY, etc.), off the board entirely.

![Planned targets vs. actual board outline](https://raw.githubusercontent.com/Reid-n0rc/Konnect/evidence/u901-l1401-and-decoupling-bugs/docs/issue-evidence/2026-09-16-placement-bugs/decoupling_bug.png)

Every current position (green) sits inside the board outline; nearly every planned target (red) sits in a straight line running off both edges.

[...]

Issue #594, opened 2026-09-16
placement: outline containment checks use bbox, not true polygon shape, for non-rectangular Edge.Cuts outlines


Reid-n0rc:
## Problem
Every placement consumer of `board_outline_bbox()` treats the board outline as its rectangular bounding box, not the true polygon/arc shape:

- `crates/konnect-core/src/tools/placement.rs:185` (`handle_score_placement`)
- `crates/konnect-core/src/tools/placement.rs:789` (`handle_auto_place`)
- `crates/konnect-core/src/tools/placement.rs:1094` (`handle_force_directed`)

For any non-rectangular outline — concave, L-shaped, notched, or with rounded corners — a component can sit inside the outline's bbox but outside its true shape (e.g., in a concave notch), and be scored/placed as if it fit.

This is a pre-existing approximation, not introduced by #593 (which only teaches `board_outline_bbox` to recognize `gr_poly`). It already applied to any non-rectangular board, including multi-segment `gr_line` outlines — `gr_poly` just makes such boards easier to author.

## Suggested direction
Not investigated yet. Likely candidates: point-in-polygon / polygon-vs-bbox containment against the outline's actual vertex/arc list (the same data `board_outline_bbox` already reads), falling back to today's bbox behavior for genuinely rectangular boards.

## Origin
Flagged during review of #593 by @neusse: "Keep concave true-shape containment out of this PR, but open/link one focused follow-up so 'bbox recognized' cannot be mistaken for 'concave placement is geometrically correct.'"

Per GOVERNANCE.md's claiming process: I don't have write access to self-label/assign this one — posting for visibility, open to whoever picks it up.

Reid-n0rc:
Linked from mixelpixx/Konnect#599, which fixes #593.

neusse:
Maintainer triage: this is valid and is not a duplicate of #593. #593 made polygon outlines discoverable by the existing bounding-box calculation; this issue identifies the remaining semantic gap between “inside the outline's bounding box” and “inside the usable board shape.”

I have linked it into the coordinate and geometry correctness tracker, #575. A focused implementation should preserve a rectangular fast path, derive true closed outline geometry (including arcs), distinguish outer boundaries from cutouts, and test concave/notched and rounded real-KiCad fixtures. The placement consumers should share one containment result rather than each inventing its own approximation. P1/design-needed is appropriate: it can approve physically invalid placement, but it is not evidence of an immediate write or data-loss path by itself.

[...]

PR #628, opened 2026-09-18
fix(placement): score_placement prefers the live IPC board over a stale saved file

### LINKED ISSUE #595: bug(placement): score_placement silently omits a component after a live move, hiding a real courtyard overlap it introduced
## Summary

`score_placement` can report zero deductions for a component that has just been moved onto another real, populated component — including on a board where KiCad IPC is live and reachable, and where the mutation itself (`move_component`) confirmed `"source":"ipc"`. There is no error, no partial-result marker, and no disclosed source/provenance: the response is indistinguishable from "this component has no problems."

Per [`docs/RELIABILITY_CONTRACT.md`](https://github.com/mixelpixx/Konnect/blob/main/docs/RELIABILITY_CONTRACT.md), *"A check that could not run is `BLOCKED`, never a silent pass"* and *"echoing the request is not evidence that the operation applied it."* This is exactly that failure mode, and it sits inside the surface already named in #574's staged plan (item 4: *"migrate or explicitly classify `score_placement` ... Where IPC cannot supply the required data, require/disclose the saved snapshot rather than presenting it as current live state"*). Filing this as a concrete, reproduced instance rather than folding it into the umbrella, per #574's own note that closing a sub-item is not the same as completing it.

## Environment

- Konnect 0.11.1 (commit `1db14b9`)
- KiCad 10.0.6, macOS (aarch64)
- KiCad running with the board open live over IPC for the entire repro (confirmed via `open_project`)

## Repro

Board has `U901` (an Ethernet PHY, QFN) and `L1401` (a Power Supply inductor), both real, populated, non-DNP components on the same copper side (`B.Cu`).

```
move_component(board="...", reference="U901", x=52, y=104)
→ {"moved":"U901","source":"ipc","x":52.0,"y":104.0}
```

Confirmed live via `get_component_pads("U901")` — pads now centered on (52, 104), matching the move.

Courtyard geometry, read directly via `list_board_footprint_graphics`:

- `L1401`: origin (51.034, 106.767), square courtyard ±2.26 mm → spans **x 48.77–53.29, y 104.51–109.03**
- `U901` after the move: origin (52.0, 104.0), courtyard ≈ ±1.8 mm → spans **x 50.2–53.8, y 102.2–105.8**

These overlap by **3.09 × 1.29 mm ≈ 4.0 mm²** — a larger overlap than nearly every courtyard-overlap entry `score_placement` did report elsewhere on the same board (which ranged 0.001–0.7 mm²).

U901/L1401 real overlap after the move

Then:

```
score_placement(board="...")
→ { "footprints_scored": 538, "footprints_skipped": 0, "score": 0, "verdict": "hard_fail", ... }
```

Searching the complete returned JSON for the substring `"U901"` — anywhere, in any field — returns **zero matches**. Not in `hard_failures`, not in any `deductions[].references`, not in `connector_edges`. The tool reports having scored all 538 footprints and skipped none, yet a component with a real, substantial courtyard collision is absent from every list a defect could appear in.

A second, independent symptom of the same likely cause: a different component (`C409`) was moved 4.2 mm away from its prior position (`(57.679, 94.282)` → `(60.679, 97.282)`, also via live `move_component`). `score_placement`'s courtyard-overlap entries for `C409` **after** that move name the exact same four neighbors (`FB405`, `C416`, `C491`, `C494`) as its overlap list **before** the move — at two positions 4+ mm apart, on a board this densely packed, byte-identical overlap partners are not plausible unless the check is reading a source that didn't reflect the move.

## Expected

Per the reliability contract, one of:
- The response reflects the live board (the actual current defect against `L1401` is reported), or
- If live state genuinely isn't available for this check, the response discloses that explicitly (a `saved`/stale-source field, a `BLOCKED`/partial marker, or similar) rather than silently presenting an incomplete answer as a clean one.

## Actual

Silent omission with no disclosed source and no error — a caller has no way to distinguish "this component was checked and is fine" from "this component was not actually evaluated."

## Impact

This is the reliability-contract failure mode the project treats most seriously, because it inverts the tool's purpose: `score_placement` exists to catch exactly this class of mistake (a move that collides with another part), and in this case it produced a false-clean signal for the single most consequential move made in the session — an agent relocating a component that landed on top of a different, unrelated, real part. The defect was only caught by a human visually inspecting the live board in KiCad after the fact; the tool built to catch it did not.

## Affected versions

Confirmed on Konnect 0.11.1 / KiCad 10.0.6.

## Related

- #574 (tracking issue: authoritative board-state source contract) — item 4 already names `score_placement` as needing migration/classification; this is a concrete repro for that gap.
- #542 (the two currently-named instances, `get_layer_list`/`get_netclasses`) — same defect class, different tool.

Reid-n0rc:
## Summary

Closes #595

`score_placement` always read the saved `.kicad_pcb` file, even when KiCad held that exact board open live with unsaved moves or edits. A hard courtyard overlap introduced by a live `move_component` could be entirely absent from the response, with no error and no way to tell the answer was stale.

## Approach

`score_placement` now prefers the exact board's live IPC snapshot — via `SaveDocumentToString`, the same read-only mechanism `run_drc`'s `sync_live_board` option already uses — when KiCad has that board open, falling back to the saved file when it doesn't (unreachable IPC, or a `Target` error proving the board isn't open there). Scoring policy and every existing response field are unchanged; the response adds `source`: `"ipc"` or `"saved_file"` (matching `run_drc`'s own source-disclosure convention), so a caller can tell which board state was judged. A genuine IPC target error (board open but something else went wrong) still surfaces as a structured error rather than silently falling back to the file.

**Revision note:** an earlier version of this PR used `"file"` for the fallback value and covered the live-vs-saved distinction with a synthetic (near-empty-file) test only. Re-reading the full accepted issue thread (not just the final short "accepted" comment) surfaced that the maintainer's own design-direction comment explicitly specifies `source: "saved_file"` and lists specific required evidence — a real live-move collision repro, served `tools/call` coverage, and a negative control. All three are now included; see Validation below.

### Architectural fit

Extends the existing `with_board_ipc_classified` / live-preferred-with-fallback pattern already used by `get_component_pads`, `list_board_footprint_graphics`, and `run_drc`'s live-sync path — no new mechanism introduced.

## Branch and dependencies

Base branch: `main` (rebased onto latest `upstream/main` before submission)
Depends on: none
Series order: first of the placement-module queue (#595 → #596 → #597 → #594), per maintainer direction — the later three overlap this module and are held until this lands.
Unique commits/acceptance criteria owned by this PR: live-vs-saved board source selection for `score_placement` only. No scoring-policy change.
Next PR to promote after this one: #596 (`place_decoupling_caps` exact-reference rework), rebuilt on current `main` after this merges.

## Compatibility and safety

Additive response field only (`source`). No schema/argument changes, no new refusal paths for existing callers. `score_placement` remains read-only — no board mutation, no save.

### Changed tool behavior

|---|---|
| Accepted inputs and declared defaults | Unchanged — still just `board` |
| Invalid/unsupported inputs and structured errors | Unchanged file-not-found handling; a genuine IPC `Target` error now surfaces via `ipc_target_error_result` instead of silently falling back to the file |
| Target, data source and prerequisite state | New: exact-board live IPC preferred, saved file is the disclosed fallback; response names which one (`source`) |
| Observed changes and preserved unrelated objects | Read-only tool; no mutation |
| Failure before/after mutation, including applied work | N/A — no mutation |
| Recovery from partial/uncertain results without repeating applied work | N/A — no mutation |

**Required evidence per the accepted issue thread**, each independently verified:

- **Live-move collision repro**: `a_live_move_that_creates_a_collision_is_reported_not_the_clean_saved_file` — the saved file is a clean, real KiCad-authored fixture (would score 70, no hard failures); the live IPC snapshot carries the exact courtyard-overlap-inducing move used elsewhere in this test file (`(at 30 15 90)` → `(at 20 15 90)`, C1/C2 collision). Asserts `source: "ipc"`, `verdict: "hard_fail"`, `score: 50`.
- **Saved-file fallback**: existing fixture test now asserts `source: "saved_file"` when no live KiCad is present.
- **Served `tools/call` coverage**: `served_score_placement_reports_the_live_collision_over_the_clean_saved_file` drives the same scenario through the real MCP dispatch layer (`McpHandler::handle_message`), not a direct handler call — mirrors `drc.rs`'s own `served_*` test convention.
- **Negative control**: performed directly, not just written as a test. Temporarily reverted `handle_score_placement` to its pre-fix, saved-file-only behavior and reran the suite. All three tests above failed, and critically, the live-move-collision test reproduced **today's exact reported bug** — `{"verdict":"pass","score":70,"hard_failures":[],"source":"saved_file"}` on a board with a real, live courtyard collision. Restored the fix immediately after; diff is byte-identical to before the experiment (confirmed via `git diff --stat`).

```
$ cargo fmt --all -- --check
(clean)

$ cargo test -p konnect-core --lib tools::placement
test result: ok. 33 passed; 0 failed; 0 ignored; 0 measured; 1292 filtered out

$ cargo clippy --workspace --locked --all-targets -- -D warnings
(clean, exit 0)

(all green, exit 0)

(all green, exit 0)
```

- [ ] Real-KiCad live check — not run in this environment; the live-IPC path reuses `SaveDocumentToString`/`with_board_ipc_classified`, already exercised live by `run_drc`'s `sync_live_board` option. A direct live confirmation of this PR's own path is still worth doing against a running KiCad 10.0.6 before merge if the reviewer wants that evidence.

## Review checklist

- [x] The branch includes current `upstream/main`, has no merge conflicts, rebased onto v0.12.0.
- [x] The PR shows only its unique commits and diff.
- [x] New behavior and failure paths have regression coverage — see the required-evidence list above.
- [x] File mutations are atomic and preserve unrelated content — N/A, no mutation.
- [x] IPC reads verify the requested board via `with_board_ipc_classified`'s `ensure_board_is_active`.
- [ ] Tools added/removed — N/A, no tool added/removed.


neusse:
Please change the issue line to `Closes #595` so GitHub records the terminal closure path, then reply with the unchanged full head SHA (or the new SHA if code changes).

The exact head `47d0e460f8ef97b8dc5f4894ad7ec543d7a1e508` otherwise matches the accepted #595 contract: exact-board live IPC preference, `source: "ipc" | "saved_file"`, unchanged scoring policy, real stale-vs-live collision fixture, served-dispatch coverage, and a performed negative control. It is the base-forming first placement PR; #596 → #597 → #594 remain behind it. The only current blocker is closure accounting: `closingIssuesReferences` is empty because `Issue: #595` does not close the issue.

Reid-n0rc:
Done — description now says `Closes #595`.

Head SHA is unchanged: `47d0e460f8ef97b8dc5f4894ad7ec543d7a1e508`. No code changes since your last review; only the description's issue-reference line was edited, so this comment reflects that edit rather than a new commit.

neusse:
Superseded by #648, which reconstructs this PR's single focused commit on current `main` and preserves Reid Crowe as the commit author. The old base was 14 upstream commits behind; comparing it to current `main` included nearly 4,000 lines of unrelated deletions. Continue review and merge evidence on #648. This PR's implementation and validation remain credited there.

PR #629, opened 2026-09-18
feat(pcb): flip_component drives KiCad 10.0.6's native FlipItems over IPC

### LINKED ISSUE #604: Use KiCad 10.0.6 native FlipItems for live footprint flips
## Summary

`flip_component` refuses to flip any footprint whose `(model …)` block has a non-zero `offset.y`, `rotate.x`, or `rotate.y` — which is exactly the case KiCad's own **F** flip handles correctly by negating those three values as part of the flip. Konnect's tool doesn't implement that transform at all; rather than apply it, it refuses the whole operation. The result: a footprint that flips fine with **F** in the KiCad GUI cannot be flipped through Konnect.

## Environment

- Konnect 0.11.1 (commit `7a0924f`)
- KiCad 10.0.6, macOS (arm64)
- Footprint: `Malmoset_Caps:C_0201_Perfect_T0.33mm`, `(model "../3DMODEL/C_0201_T0.33mm.step" (offset (xyz 0 -0.165 0.13)) (scale (xyz 1 1 1)) (rotate (xyz 0 0 0)))`

## Repro

Board has several placed instances of the footprint above, all on `F.Cu`, needing to move to `B.Cu`.

```
flip_component(board="<board>.kicad_pcb", reference="C435", layer="B.Cu")
```

With the board closed in KiCad (required — see below), this returns:

```
Refusing to edit the board: the 3D model's offset.y is -0.165, and a flip would
have to move it; flipping this footprint would leave the model where it was.
The board file was not modified.
```

Same result for every other instance of this footprint on the board (tried 19 in total). Pressing **F** on the same footprint inside the KiCad 10.0.6 GUI flips it correctly, with the 3D model staying in the right place.

The refusal is deliberate and documented at the source (`crates/konnect-core/src/tools/pcb_components.rs:1585-1615`, `refuse_model_a_flip_would_move`):

> "KiCad's own flip transforms a model's Y offset and its X/Y rotation; this tool leaves `(model …)` untouched, which is silently wrong for any model where those are non-zero. Rather than guess the transform — I have not been able to measure KiCad's flip directly, because KiCad 10.0.5 exposes no `FlipItems` to drive it and its demo boards contain no back-side footprint with a non-zero offset to compare against — this refuses..."

The comment also states the intended transform precisely (`moved: &[usize] = if tag == "offset" { &[2] } else { &[1, 2] }` — i.e. negate `offset.y`, `rotate.x`, `rotate.y`; leave `offset.x`, `offset.z`, `rotate.z` alone), it's just not implemented.

## Expected

Flipping a footprint through Konnect produces the same result as pressing **F** on it in KiCad: 2D geometry mirrors, layers swap, *and* the 3D model's `offset.y`/`rotate.x`/`rotate.y` are negated so the model stays correctly positioned. No refusal needed for this case once the transform is implemented.

## Actual

`flip_component` refuses unconditionally whenever any of those three values is non-zero, regardless of magnitude. There's no override and no partial-apply path (e.g. flip the 2D geometry and warn about the model) — the whole operation is blocked.

## Impact

This isn't a data-safety bug — the refusal is safe, and the comment's own audit (14,818 stock-library footprints, 3 with non-zero `offset.y`, 0 with non-zero `rotate.x`/`y`) shows it's rare in KiCad's own libraries. But it isn't rare in practice: it hit a real board on the first footprint type tried, because many custom/third-party libraries set a 3D model offset to align the STEP model with the footprint origin (as this one does, by -0.165mm). Every instance of an affected footprint — 19 on this board alone, all the same common 0201 cap package — is unflippable via Konnect, with the only workaround being to flip manually in the KiCad GUI and hand back to Konnect for everything else. Since flip already requires a closed board (no live IPC path — a separate, known limitation), this compounds into a genuinely disruptive workflow break for any board using a non-default footprint library with aligned 3D models.

## Affected versions

Confirmed on Konnect 0.11.1 (commit `7a0924f`) / KiCad 10.0.6, macOS.

## Related

- The transform this needs (negate `offset.y`, `rotate.x`, `rotate.y`) is already worked out and stated in the refusing function's own doc comment — this is a "write the code," not a "figure out the math" gap.
- Distinct from flip's live-IPC limitation (KiCad 10.0.5/10.0.6 expose no `FlipItems`, so flip only works on a closed board file) — that's a separate, already-documented constraint, not this issue.

Reid-n0rc:
## Review checklist  - [x] The diff is focused and contains no generated output, personal data, or unrelated cleanup. (Doc/test-assertion updates beyond the original core diff were needed because `flip_component`'s reclassification has real, correct downstream effects — the Claude-hook matcher it drives, and docs describing its old closed-board-only behavior — and are included here rather than left stale.) - [x] The branch includes current `upstream/main` (rebased onto v0.12.0), no merge conflicts. - [x] The branch was based on latest `upstream/main`, not a release tag. - [x] New behavior and failure paths have regression coverage: 11 `flip_footprint_test.rs` unit tests (direction/KIID, exact-match success, the `IRS_OK`+empty-results trap, wrong-KIID/non-OK-status hard failures with rollback, `AS_UNHANDLED`, readback-mismatch, F→B→F round trip) plus 3 new end-to-end `pcb_components.rs` tests through `attempt_ipc_write`. All pre-existing closed-board flip tests pass unmodified except one (`a_reachable_kicad_that_rejects_the_flip_never_touches_the_file`, renamed and updated: a reachable-but-rejecting KiCad now fails closed via the IPC path instead of falling back to the file, matching this repo's existing pattern for that case). - [x] IPC mutations verify the requested board; atomic/partial behavior and safe recovery are explicit under the reliability contract (see table above). - [ ] Tools added/removed — N/A, no tool added/removed; `flip_component`'s registered `BoardAccess` classification changed (covered by the updated `install.rs` test, not a `tool_count`/directory-count change).  🤖 Generated with [Claude Code](https://claude.com/claude-code)


neusse:
Please provide the two required evidence items before this can enter the merge queue:

- Run the accepted real KiCad 10.0.6 live round trip on a KiCad-authored footprint whose 3D model has a nonzero Y offset or X/Y rotation: F.Cu → B.Cu → F.Cu. Record the exact board/footprint, observed layer and model transforms after each flip, preservation of unrelated content, and undo/recovery behavior.
- Perform the accepted negative control: temporarily break the per-item-result/readback guard, run the focused regression test, record the failing test/result, then restore the guard and confirm the diff is clean.

The code shape on `0773ec7623f4aced3a889e4769f91252c372d84a` otherwise follows #604: coherent official 10.0.6 proto snapshot, exact-board targeting, transaction rollback, no false success from envelope status, independent readback, structured unsupported/uncertain outcomes, and unchanged guarded file fallback. `Closes #604` resolves correctly. The `run:e2e-kicad` label is applied, but the generic workflow is not a substitute for this feature-specific live evidence.

neusse:
The exact-head code review found three additional blockers; please address this complete checklist together with the live evidence already requested:

- In `crates/konnect-ipc/src/client.rs`, make `footprint_3d_models` return fallible evidence. It currently drops malformed model messages with `filter_map(...decode(...).ok())` and turns missing vectors into zero using `unwrap_or(0.0)`. Post-mutation verification cannot treat malformed/missing model state as an empty or zero-valued observation; propagate it and classify the flip outcome as uncertain.
- Align all public guidance with the implemented fail-closed rule. `README.md`, bundled `kicad-pcb/SKILL.md`, and the ambiguous `tool-directory.md` wording currently imply an older reachable KiCad falls back to file editing, while the implementation and migration note correctly return `unsupported_capability`.
- Make the public 3D-model evidence units explicit. The same `{x,y,z}` shape currently represents scale, degrees, and millimetres; use typed/unit-suffixed fields or document the exact units at the response boundary.
- Run the required real KiCad 10.0.6 F.Cu→B.Cu→F.Cu round trip with a nonzero model transform and record observed preservation/undo behavior.
- Perform and record the required negative control by breaking the per-item/readback guard, proving the focused regression fails, then restoring it.

Keep the coherent 10.0.6 proto snapshot and #604 closure reference unchanged. This stays waiting on the author; the generic E2E run alone cannot replace the feature-specific evidence.

neusse:
Maintainer refresh pushed at exact head 3b4b9613a403212cb2cf111e00d961fb1b91c953 after #605 landed. This resolves the complete review checklist: malformed or incomplete 3D-model readback is now fallible and yields an uncertain outcome; response fields name offset millimetres, rotation degrees, and unitless scale; README, tool directory, and bundled PCB guidance all state the same fail-closed older-KiCad behavior; focused regression coverage increased to 11 tests; the performed negative control failed exactly when the readback guard was neutralized and passed after restoration; and the real KiCad 10.0.6 F.Cu→B.Cu→F.Cu test passed on a disposable KiCad-authored footprint with a nonzero 2.0 mm model Y offset. Full workspace unit/integration tests, doctests, clippy, format, and the post-rebase focused suites pass locally. CI is now the remaining gate.

neusse (approved):
Exact-head maintainer review complete at 3b4b9613a403212cb2cf111e00d961fb1b91c953. All review findings are resolved, the performed negative control proves the readback guard is load-bearing, the feature-specific live KiCad 10.0.6 F.Cu→B.Cu→F.Cu round trip passed with nonzero 3D-model Y offset, all ten required checks pass, and the opt-in real-KiCad workflow passes. Closes #604 correctly.

Issue #622, opened 2026-09-17
batch_place_components never writes the junction its single-component sibling does, so a batch-placed pin on a wire is unconnected


pauliuszaleckas:
`handle_add_schematic_component` runs a junction pass after it writes, and its comment explains why the batch path does not need one inline (`crates/konnect-core/src/tools/sch_components.rs:689`):

```rust
// A pin landing mid-segment on an existing wire needs a junction dot, or
// KiCad's netlister treats it as unconnected. Runs after the write because
// it re-reads the saved file; `place_one_component` stays pure so the batch
// path can do one junction pass for the whole batch instead of one per part.
let junctions = match crate::tools::add_pin_midwire_junctions(&sch_path, ref_str) {
```

`handle_batch_place_components` (`crates/konnect-core/src/tools/sch_batch.rs:484`) never makes that pass. It calls `place_one_component` per part, `sch.overwrite()`, then reads back — no `add_pin_midwire_junctions`, no `reconcile_junctions_at`, and no junction key in its response. The batched pass the comment promises was never written.

## Demonstrated

KiCad 10.0.6 on Fedora, Konnect 0.12.0 built from `main` @ 9d33cf0 with the #615 branch on top; that branch does not touch this path (`crates/konnect-core/src/tools/sch_batch.rs` has a zero-line diff against main).

Two sheets, identical geometry — a wire carrying `NETA`, and one `Device:R` at `(127, 101.6)`, whose pin 1 therefore lands at `(127, 97.79)`, mid-span on that wire. The only difference is which tool places it:

```
add_wire                  (114.3, 97.79) -> (139.7, 97.79)
add_schematic_net_label   NETA at (133.35, 97.79)
# sheet 1
batch_place_components    [{lib_id: Device:R, x: 127, y: 101.6, reference: R1}]
# sheet 2
add_schematic_component   Device:R as R1 at (127, 101.6)
```

| | junction in the file | response |
|---|---|---|
| `add_schematic_component` | 1, at `(127, 97.79)` | `"junctions_added": [{"x": 127.0, "y": 97.79}]` |
| `batch_place_components` | **0** | no junction key; `outcome.status: "complete"` |

`kicad-cli sch export netlist` on the two sheets:

| Sheet | `/NETA` | unconnected |
|---|---|---|
| `add_schematic_component` | R1.1 | R1.2 |
| `batch_place_components` | **net absent** | **R1.1**, R1.2 |

[...]

PR #632, opened 2026-09-18
fix(schematic): carry a no-connect flag with the pin it protects

### LINKED ISSUE #626: A placement change orphans a no-connect flag and can then wire the pin it protected into a net

pauliuszaleckas:
Closes #626

## The bug

A no-connect flag is a sheet item at a coordinate, so a placement change left it where it
was while the pin it protected moved away. `reconcile_junctions_at`'s `has_no_connect`
guard then looked at the *arrival* point, found nothing there, and — where the pin had
landed mid-span on a wire — wrote a junction dot. KiCad's netlist then puts a pin the
caller had explicitly declared unconnected onto that net.

ERC reports the stranded marker as `no_connect_dangling`. Nothing reports the new
connection.

## The contract, as the maintainer settled it

A no-connect is intent attached to a specific pin, not decoration owned by a coordinate.
`plan_no_connect_carry` (`sch_wiring.rs`) is the single shared implementation:

- **Identity, not coincidence.** A marker is followed by the symbol instance UUID, the
  unit it places, the pin number, and the pin's library-local geometry. The last is there
  because one unit may carry two pins with the same number — KiCad stacks hidden power
  pins that way — and local geometry is invariant under a placement change while a symbol
  swap moves it.
- **One atomic mutation.** The marker's new position is applied to the same in-memory
  model (or the same `SexpEdit` batch) as the symbol, so one write commits both.
  Junction reconciliation then runs against the resulting content and sees the carried
  marker at the arrival point, so it adds no dot.
- **Refuse before writing** when the mapping is not one-to-one:
  - `ambiguous_target` — the pins under one marker land on different points (two pins
    stacked with only one moving), or the carry would stack two markers on one point.
    `target` names the marker and position; `candidates` name the competing pins.
  - `stale_target` — the sheet cannot answer at all: a `lib_symbols` lookup that failed,
    a placed symbol with no UUID, a marker with no UUID.
  A refusal leaves the file byte-identical.
- **Report from observed content.** `no_connects_moved` is read back out of the written
  file. A marker the readback cannot find at its planned point returns
  `mutation_outcome_uncertain` rather than a success with a footnote.

## Scope

Wired into the three tools that relocate pin endpoints **and** reconcile junctions today,
so all three could write the unsafe dot: `move_schematic_component`,
`bulk_move_schematic_components`, and `rotate_schematic_component`.

`rotate_schematic_component` reconciles no junctions on `main` yet — that is #615/#627 —
so today the turn's symptom is only the orphaned marker. Carrying it here is what lets
#627 land safely: with this contract in place, its new reconciliation cannot make the
connection #626 demonstrates. #627 will need a small rebase on the rotate handler.

`move_region`, `replace_component` and `batch_place_components` reconcile no junctions at
all yet (#623, #625, #622) and are deliberately unchanged. They inherit
`plan_no_connect_carry` when they gain reconciliation; the planner already refuses on the
symbol-definition case the maintainer called out (a replacement that removes or renumbers
the protected pin).

## Deviation from the issue's two options

The issue offered "carry the marker" (option 1) or "suppress the add and let it dangle"
(option 2). The maintainer's comment settles on option 1 plus a refusal path, which is
what this implements — option 2 is not taken.

## Evidence

New fixture `crates/konnect-core/tests/fixtures/no_connect_carry_kicad10.kicad_sch`, built
through Konnect tools against KiCad's stock `Device` library and then force-resaved by
KiCad 10.0.6 (`kicad-cli sch upgrade --force`), so what is committed is Eeschema's own
serialization. Provenance, the case table and the oracle tables are in
`no_connect_carry_kicad10.README.md`.

The oracle is `kicad-cli sch export netlist`, run on the file each tool actually wrote:

| File | `/NETB` | `R2` pin 1 `pintype` |
|---|---|---|
| the fixture | R4.1 | `passive+no_connect` |
| `R2` moved | R4.1 | `passive+no_connect` |
| `R2` moved, **no fix** | **R2.1, R4.1** | `passive` |
| `R2` shifted in bulk | R4.1 | `passive+no_connect` |
| `R2` shifted in bulk, **no fix** | **R2.1, R4.1** | `passive` |
| `R2` -> 90° | R4.1 | `passive+no_connect` |
| `R2` -> 90°, **no fix** | R4.1 | `passive` |

The `no fix` rows are the same fixture driven by a `konnect` binary built from `main`
@ 9d33cf0. KiCad puts the protected pin on `/NETB` only when the dot is there, and writes
`passive+no_connect` only while a marker sits on that pin's endpoint — two answers a wrong
implementation cannot produce together.

ERC (`kicad-cli sch erc --severity-all`) corroborates: 8 violations on the fixture and 8
after each fixed placement change, against 2 `no_connect_dangling` (and, on the turn, 10
total) without the fix. Both written files were reopened and re-serialized by KiCad
(`sch upgrade --force`) with the marker at `(154.94 160.02)`.

Refusals were exercised against the real binary over stdio as well as in tests: moving or
bulk-shifting `R3` or `R5` — whose pins stack under one marker — returns
`ambiguous_target` and leaves the file byte-identical.

## Negative controls

Each guard neutered in turn, full `cargo test -p konnect-core --lib` run, then restored.
All counts are from a deterministic run.

| Guard neutered | Tests failing |
|---|---|
| the typed-path carry never runs (`carry_no_connects` returns empty) | 6 |
| the bulk-path carry never runs | 2 |
| the one-arrival-point check accepts a split arrival | 3 |
| the "pin is gone from the result" check falls through | 1 |
| the two-markers-on-one-point check is bypassed | 1 |
| `resolved_placed_pins_by_reference` stops refusing unresolved geometry | 2 |
| a marker with no UUID is skipped instead of refusing | 1 |
| an empty UUID is accepted instead of counting as absent | 1 |
| the readback reports the plan instead of the file | 1 |
| `reconcile_junctions_at`'s arrival-point `has_no_connect` guard (pre-existing) | 5 |

The last row is the pre-existing guard this fix finally feeds; it also takes down
`reconcile_does_not_add_a_dot_over_a_no_connect` and
`attached_no_connect_survives_when_a_remaining_pin_shares_the_point`, which were already
there. The unresolved-geometry row fails two tests because the delete plan now shares that
helper rather than keeping its own copy of the check.

## Known limitations

Two cases this contract deliberately does not close, both measured against KiCad
10.0.6 rather than reasoned about:

- **Pins that swap coordinates.** A `Device:R` turned 180° exchanges its two pin
  positions, so the junction pass — whose candidates are the symmetric difference of
  pin *coordinates* — gets nothing to judge and any dot between the two pins stands.
  A marker carried onto that dot ends up on a pin that is still on the net. The pin is
  on that net **with or without this change** (`/NETB -> R2.1, R4.1` either way), so
  the carry does not cause it; what changes is that ERC reports it as
  `no_connect_connected` (4 violations) instead of leaving a stranded marker and a
  silent connection (2). Closing it means feeding the junction pass from the identity
  derivation, which changes which dots that pass judges — a second reviewable outcome.
  Documented at the divergence in `plan_no_connect_carry`'s doc comment.
- **The unresolved-geometry refusal is sheet-wide.** One placed symbol whose
  `lib_symbols` entry is missing blocks every placement change on a sheet that has any
  no-connect, including changes nowhere near a marker, because Konnect then cannot tell
  which pins sit under a marker at all. `delete_schematic_component` already refuses the
  same way for the same reason, so this follows the established contract rather than
  inventing one — but it is a behaviour change on such sheets and
  `docs/API_MIGRATIONS.md` now names the cause and the real remedy (repair the symbol,
  not the marker). Say the word if you would rather it degraded than refused.

## Tests

- `sch_wiring::no_connect_carry_tests` — 4 planner tests for the refusal branches that a
  placement change cannot reach on its own (pin gone, unresolved geometry, UUID-less
  marker) plus the no-op case.
- `sch_components::no_connect_carry_tests` — 6 handler tests: move, turn, a change that
  carries nothing, both stacked-pin refusals, the marker-collision refusal, and the
  readback's `mutation_outcome_uncertain`.
- `sch_batch::bulk_move_no_connect_tests` — 2 handler tests: the bulk carry and the bulk
  refusal.
- `mcp::handler::no_connect_carry_dispatch_tests` — 2 served `tools/call` tests, one for
  the new response fields and one proving the structured refusal and untouched file
  survive the dispatch boundary.

One of the handler tests covers a case the two paths used to disagree on: the typed
model reads a missing marker UUID as `""` and writes it back as `(uuid "")`, so the
typed path accepted a marker the S-expression path refused — and two such markers shared
that one key. An empty UUID now counts as absent on both.

## Validation

```text
cargo fmt --all -- --check                      ok
cargo clippy --workspace --locked -- -D warnings ok
cargo test --workspace --locked --lib --tests    ok
cargo test --workspace --locked --doc            ok
```

`crates/schematic-viewer` is untouched, so its separate suite was not run.

## API_MIGRATIONS

Adds one entry at the top of `docs/API_MIGRATIONS.md`. Existing entries are untouched and
in their original order, so the file's diff against `main` deletes nothing and moves
nothing.

## Risk and rollback

The new response fields are additive. The behavior change a caller can see is that
`junctions_added_count` is `0` where it used to be `1` for a protected pin — which is the
fix — and that a placement change can now be refused where it previously succeeded. The
refusals are narrow: they need a no-connect on the sheet *and* a pin under it that the
change cannot follow one-to-one. A sheet with no no-connect markers takes an early return
before any extra parsing, so the common path is unchanged.

Rollback is a revert of this commit; nothing persists outside the schematic files the
tools already write.


neusse:
Reconstructed as #641 on current main because this branch had become conflict-bound. The implementation commit and original authorship are preserved; conflict resolution retained both the current-main additions and this PR's additive tests/documentation. Closing this copy as superseded so there is one merge candidate and one closing path for #626. Thank you, @pauliuszaleckas, for the implementation and unusually strong evidence package.

PR #638, opened 2026-09-18
fix(placement): place_decoupling_caps requires exact capacitor references

### LINKED ISSUE #596: bug(placement): place_decoupling_caps matches candidates by ANY shared net (including GND), sweeping unrelated caps into a nonsensical off-board row

Reid-n0rc:
## Summary

Closes #596

`place_decoupling_caps` discovered candidate capacitors by "shares at least one named net with the IC." On a real board, GND is shared by nearly every capacitor, so calling it on an IC with a ground pin swept in essentially every decoupling cap on the board into a single evenly-spaced row, running far past both board edges — up to ~297 mm on a 96.7 mm board in the reported repro.

## Approach

Per the design accepted in the issue thread (agreed with @neusse, after an initial net-exclusion-heuristic proposal was explicitly rejected as still-unverifiable inference): drop net-based candidate discovery entirely rather than refine it.

- `capacitor_references` is now a required, non-empty argument — the caller names exactly which capacitors to place. An unknown reference or a reference that isn't a decoupling-family capacitor footprint is a structured `invalid_argument` error naming every bad reference at once, not a silent skip. A duplicate reference is refused the same way.
- The row-layout, spacing, side resolution, and courtyard-bbox math are unchanged — that logic is mechanical placement given exact parts, not the part that owned design intent.
- New shared contract: `PlanApplicability` (`crates/konnect-core/src/tools/placement/plan_status.rs`), intended for #597's force-directed redesign too. A plan's `plan_status` is `"applicable"` only when every planned target lands inside the board outline **and** the plan actually improves `score_placement`'s score over `score_before` without introducing a hard failure the board didn't already have. Otherwise it's `"blocked"`, with `blocking_reasons` naming exactly why — and apply refuses before the first mutation.

## Branch and dependencies

Base branch: `main`
**Depends on: #595 (PR #628), not yet merged.** Per the accepted design comment and the maintainer's documented landing order, this is the second PR in the placement-module queue (#595 → #596 → #597 → #594) and should not merge ahead of #595. Opening now as a draft for visibility/review; will rebase and mark ready once #595 lands.
Series order: second of the placement-module queue.

## Compatibility and safety

**Breaking, input-schema change** — `capacitor_references` is a new required argument; any existing caller relying on net-based auto-discovery must now supply exact references. Per GOVERNANCE.md this is a **minor** version bump, left for the maintainer's release process rather than bumped in this PR.

`nets_by_reference` (the net-pairing helper this tool alone used) is deleted outright, not left dormant — there is no remaining code path that infers which capacitor decouples which IC.

## Validation

| Behavior | Contract and evidence |
|---|---|
| Accepted inputs and declared defaults | `capacitor_references` (new, required, non-empty array); missing/empty/duplicate/unknown/non-capacitor references are structured `invalid_argument` errors before any planning |
| Target, data source and prerequisite state | Every requested reference resolved against the freshly-read board; no inference |
| Observed changes and preserved unrelated objects | Row layout math unchanged; only candidate selection and the response's `plan_status`/`blocking_reasons` fields are new |
| Failure before/after mutation | A blocked plan (out-of-bounds target, or no score improvement) refuses at apply time before `update_closed_board_footprints` runs — verified by asserting the board file is byte-identical after a refused apply |
| Recovery without repeating applied work | Re-running a dry run after a successful apply reports `plan_status: "blocked"` (no further improvement available) rather than re-executing a no-op |

New regression tests (`crates/konnect-core/src/tools/placement.rs`):
- `decoupling_refuses_a_missing_capacitor_references_argument`, `decoupling_refuses_an_empty_capacitor_references_list`, `decoupling_refuses_an_unknown_or_non_capacitor_reference`, `decoupling_refuses_a_duplicate_capacitor_reference`
- `decoupling_plan_ignores_requested_reference_order` — caller-supplied order doesn't change the plan
- `decoupling_blocks_a_plan_that_does_not_improve_the_score` — apply once (score 70→100), then a second dry run and a second apply attempt both report/refuse as blocked, board byte-identical after the refused apply
- `decoupling_blocks_a_plan_landing_outside_the_board_outline` — #596's own repro shape, scaled to the fixture: moving the IC near the board edge makes the planned row exceed the outline; blocked, naming the reference and its planned coordinates
- `plan_status.rs`'s own unit tests for `PlanApplicability`

```
$ cargo fmt --all -- --check
(clean)

$ cargo clippy --workspace --locked --all-targets -- -D warnings
(clean, exit 0)

(all green, exit 0)

(all green, exit 0)
```

- [ ] Real-KiCad live check — not run in this environment; apply-mode logic is unchanged (`update_closed_board_footprints`), already exercised live by other placement tools.

## Review checklist

- [x] The branch was based on latest `upstream/main` (pre-#595, since #595 hasn't merged — will rebase once it does).
- [x] The PR shows only its unique commits and diff.
- [x] New behavior and failure paths have regression coverage — see the test list above.
- [x] File mutations are atomic and preserve unrelated content — a blocked apply writes nothing (asserted byte-identical).
- [ ] Tools added/removed — N/A, no tool added/removed; input schema of an existing tool changed (documented above).


neusse:
Superseded by #649. The replacement reconstructs this draft's unique implementation onto current `main`, preserves Reid Crowe's authorship, and adds the accepted exact-reference, served-dispatch, structured-refusal, no-write, and migration-documentation corrections. Closing this stale draft so there is one review head for #596.

Issue #574, opened 2026-09-14
Define one authoritative board-state source contract for PCB reads, analysis, and CLI workflows


neusse:
## Summary

Konnect does not yet have one authoritative contract for choosing between the board held live by KiCad and the last-saved `.kicad_pcb`. Individual tools independently choose IPC, file parsing, or `kicad-cli`; some report provenance, some mix sources under one answer, and some silently fall back to a saved file after a live query fails.

This is the tracking issue for fixing that system-wide contract without turning each symptom PR into an unreviewable migration.

Known work already demonstrates both sides:

- #207 established that interactive PCB reads must prefer the exact live board and disclose their source.
- #240 established process-lifetime memory so a board previously observed live cannot later be mutated through an unsafe file fallback.
- #542 identifies the next two inconsistent readers: `get_layer_list` and the board-net portion of `get_netclasses`.
- #408 owns the distinct live-save-CLI ordering contract for DRC.

Closing #542 or #408 must not be treated as completing this umbrella by itself.

## Contract decision

Create one deep board-read module at the source-selection seam. Tool handlers should ask it for a board observation rather than each reimplementing exact-board targeting, IPC failure classification, saved-file fallback, staleness policy, provenance, and recovery instructions.

For eligible read-only tools, add an optional additive selector such as:

- `board_source: "auto"` (default): use the exact requested live board when it is available; use the saved file only when doing so is demonstrably safe and truthful.
- `board_source: "live"`: require the exact board to be open and reachable; otherwise return a structured error.
- `board_source: "saved"`: intentionally inspect the last-saved file and disclose that unsaved live changes are excluded.

The final public name should follow the repository naming conventions, but the semantics above are the requirement.

### Non-negotiable behavior

- Once the exact board has been positively identified as live, a failed live query must not silently become a successful saved-file answer.
- If a board was observed live and IPC is later lost, `auto` must report an unsafe/stale fallback rather than treating the saved file as current. An explicit read-only `saved` request may inspect that snapshot, but must report its freshness limitation.
- A responsive KiCad instance containing a different board is not the live source for the requested board.
- Every successful answer reports observed provenance. Mixed-domain answers report provenance per domain or field rather than using one misleading top-level source.
- Request values are never echoed as evidence that a source was used or synchronized.
- This selector applies only where both live and saved observations are meaningful. It must not weaken mutation safety or claim that `kicad-cli` can consume unsaved editor state.

## Source domains

Not every PCB fact has the same authority:

[...]

Issue #240, opened 2026-08-17
Closed-board fallback cannot tell "KiCAD was never running" from "KiCAD just died holding this board"


mixelpixx:
## Problem

`place_component`, `move_component` and `rotate_component` fall back to editing the board file directly when the IPC transport is classified `IpcFailure::Unreachable`. The gate is correct as far as it goes — an unreachable transport means no live editor can discard the write.

But it cannot distinguish two very different situations:

1. **KiCAD was never running.** Editing the file is safe and is the whole point of the fallback.
2. **KiCAD was running with this board open and died a moment ago** — crashed, was force-quit, or was killed by us. The transport is equally unreachable, but the user had an editor holding that board, possibly with unsaved changes, and the file on disk is now stale relative to what they lost.

In case 2 the fallback edits a file the user is about to reopen and reconcile by hand, and reports:

```json
{"source": "file",
 "warning": "KiCAD IPC was not reachable, so the board file was edited directly. KiCAD will show this footprint when it next loads the board."}
```

which reads as reassurance.

## Observed

Found while verifying #237, where Konnect itself terminated KiCAD. Driving two `place_component` calls in one session against KiCAD 10.0.5:

```
id=3 ERROR: KiCAD rejected the placement over IPC: NNG recv failed: Timed out.   <- KiCAD dies here
id=4:       {"placed":"J2", ..., "source":"file", "warning":"... KiCAD IPC was not reachable ..."}
pcbnew after: []
```

The second call wrote a footprint into the board file roughly a second after the editor holding it disappeared.

#237's own crash is fixed (`ff518c8`), so Konnect no longer causes case 2 by that route. This issue is about every other route into it: a KiCAD crash from any cause, a force-quit, an OS restart, a killed process.

## Why the first call is *not* part of this

Worth stating so it doesn't get "fixed": the failing call classified an NNG recv timeout as `Rejected`, not `Unreachable`, because a timeout carries no `TransportUnreachable` marker. That fails closed and leaves the file alone, which is right when Konnect cannot tell what happened. The problem is only the *next* call, which sees a cleanly refused connection and concludes nobody was ever there.

## Possible directions

Not obviously easy, which is why this is its own issue rather than a rider on #237:

- **Remember within the process.** If any earlier call in this server's lifetime reached a live KiCAD holding this board, refuse the fallback for that board and say why, rather than editing it. Cheap, no new state on disk, and covers the crash-mid-session case exactly. Does not survive a Konnect restart.
[...]

Issue #577, opened 2026-09-15
Track KiCad IPC endpoint discovery, ownership, and recovery


neusse:
## Ask

Coordinate KiCad IPC endpoint discovery, ownership, waiting, re-probing, and cleanup as one lifecycle program while preserving focused implementation issues and pull requests. The goal is one observable endpoint state model rather than independent startup assumptions in every caller.

This tracker governs reaching and maintaining a trustworthy KiCad IPC session. Board-state source selection after a trustworthy session exists remains owned by #574.

## Open issue accounting

- [ ] #256 — open a specific board and wait for the exact PCB Editor IPC endpoint.
- [ ] #425 — re-probe an unresolved KiCad endpoint after server startup instead of freezing the initial absence for the process lifetime.
- [ ] #531 — distinguish a legitimate Windows named pipe from a squatter or incompatible responder.
- [ ] #532 — diagnose the reported state where the expected KiCad socket never appears despite the editor setting being enabled.
- [ ] #556 — ensure the shared IPC test server releases both its endpoint and its execution promptly when explicit `close()` is omitted.

Closed #103 is process-ownership and orphan-cleanup regression history. Issue #405 is deliberately excluded until its schematic-write hang is diagnosed; it should join this tracker only if evidence identifies endpoint lifecycle as the cause.

## Shared lifecycle contract

- Represent endpoint state explicitly: not configured, discovering, waiting, ready, wrong document, stale, unresponsive, occupied by an incompatible responder, and closed.
- Bind success to the exact observed endpoint and requested document identity; a responsive but wrong editor is not success.
- Permit bounded re-probing after startup and editor changes without requiring a Konnect restart.
- Return structured diagnostics that distinguish absent, permission-denied, collision/spoofing, protocol mismatch, timeout, and stale endpoint states.
- Make process and endpoint ownership explicit. Konnect cleans up only resources it owns and never kills an unrelated live session.
- Launch-and-wait workflows have bounded timeouts, cancellation, deterministic cleanup, and actionable recovery information.
- Windows named pipes and Unix-domain sockets implement the same public states even when platform diagnostics differ.
- Test fixtures release listeners and background tasks promptly on success, error, timeout, cancellation, and drop.

## Evidence and implementation shape

[...]

PR #656, opened 2026-09-20
fix(pcb): read layers and netclasses from the live board

### LINKED ISSUE #542: get_layer_list and get_netclasses read the saved file while their sibling writers use IPC
## Summary

`get_layer_list` and `get_netclasses` read the `.kicad_pcb` file directly while the
writers beside them in the same toolsets go through KiCad's IPC API. On a board open in
KiCad with unsaved changes, they answer about the file on disk, not the board KiCad
holds — the same class of problem [#207] fixed for `get_board_info`,
`get_component_pads` and `get_pad_position`.

## Environment

- Konnect 0.11.1, built from `main` @ 39e0061
- KiCad 10.0.5, Fedora Linux

## Detail

Handlers in `pcb_board.rs` / `pcb_routing.rs` that reach only `read_to_string` +
`parse_sexp`, with no IPC path:

| Handler | Reads file | Notes |
|---|---|---|
| `handle_get_layer_list` | yes | pure reader — returns the saved stackup |
| `handle_get_netclasses` | yes | pure reader — returns saved netclasses |

For contrast, `pcb_board.rs` has 3 handlers and `pcb_routing.rs` has 4 that do use
`with_board_ipc` / `with_ipc` / `KiCadIpcClient`. So within one file, a caller gets live
answers from some tools and saved-file answers from others, with nothing in either
response saying which.

## Scope — this is a staleness bug, not a corruption bug

The file-only **writers** in these toolsets (`add_board_text`, `add_layer`,
`set_board_size`, `add_board_outline`, `delete_graphics`, `add_net`, `modify_trace`) all
go through `write_atomic`, which calls `ensure_kicad_design_document_is_closed`, and
`kicad_editor_lock_path` covers `.kicad_pcb` as well as `.kicad_sch`. They therefore fail
closed against KiCad's `~<name>.kicad_pcb.lck` while the board is open. No write is at
risk here.

Reads legitimately take no lock, which is why the exposure is reads only.

## Expected

Either the two readers get an IPC path with a file fallback, matching the pattern
[#207] established, or their responses carry the `source` marker that the IPC-first
tools already use, so a caller can tell a live answer from a saved one.

## Not reproduced live

The structural claim above is read off the current tree. I did not demonstrate a
divergent answer against a running KiCad with unsaved changes — that part follows from
[#207]'s already-accepted premise rather than from a fresh measurement.

[#207]: https://github.com/mixelpixx/Konnect/pull/207

pauliuszaleckas:
`get_layer_list` and `get_netclasses` parsed the saved `.kicad_pcb`
unconditionally. A board open in KiCad with unsaved changes was answered from
the file on disk, and `get_layer_list` said nothing about it.

Both now go through one shared board-read seam,
`crates/konnect-core/src/tools/board_source.rs`, which owns exact-board
targeting, the staleness policy, the provenance vocabulary and the structured
refusals — rather than two handler-local fallbacks. Reading KiCad's answer
itself lives one level down in `live_board.rs`, shared with the write gates
(see **Shared classification** below).

Closes #542
Part of #574

## The selector

Both tools take an additive `board_source`, defaulting to `auto`, so no
existing caller changes:

| Mode | Behaviour |
|---|---|
| `auto` (default) | The exact live board wins. The saved file is read only when the response can state why no live observation was used. |
| `live` | Structured error unless the exact board is open in a reachable KiCad. |
| `saved` | Deliberately inspects the last saved file and reports that unsaved editor changes are excluded — KiCad is not consulted even when it is holding the board. |

Non-negotiables from #574, implemented in the seam:

- Once KiCad has positively identified the board as live, a failed query is
  returned as a failure. It never becomes saved-file success under `auto` or
  `live`.
- A board this process watched KiCad hold, whose IPC endpoint then disappears,
  is refused under `auto` (`unsafe_file_fallback`), by the same #240 rule that
  protects writes. An explicit `saved` request may still inspect that snapshot
  and reports `explicitly_requested_after_live_observation`.
- A reachable KiCad holding a different board is not this board's live source:
  `live` returns `wrong_document` naming the boards KiCad does hold, `auto`
  reads the file and says `board_not_open_in_kicad`.
- The request value is never echoed back as evidence that a source was used.

## Per-domain provenance

Neither answer is single-source, so neither reports one source string.

`get_layer_list` obtains the enabled set and each layer's displayed name from
`GetBoardEnabledLayers` / `GetBoardLayerName`. A layer's ordinal `id` and its
`signal`/`power`/`mixed`/`user` type have no IPC equivalent, so they are joined
from the saved `(layers …)` table by canonical name and reported as
file-backed — `null` for a layer the editor has enabled that the file does not
carry.

```json
"sources": { "enabled_layers": "ipc", "layer_names": "ipc",
             "copper_layer_count": "ipc",
             "layer_ids": "saved_board", "layer_types": "saved_board" },
"source_evidence": { "board_state": "ipc",
                     "excludes_unsaved_editor_state": false,
                     "reason": null, "detail": "…" }
```

`get_netclasses` stays inherently mixed:

```json
"sources": { "definitions": "project_file", "patterns": "project_file",
             "board_nets": "ipc", "matched_nets": "derived" }
```

Its existing `nets_source` string is kept (public API) and now also covers the
live case.

## Shared classification

The write side already reads KiCad's answer in `attempt_ipc_write` and
`refuse_if_board_open_in_kicad`. Adding a third transcription of that ladder
for reads would have put the rule with the worst failure mode in this codebase
— a board observed live is never treated as safely absent afterwards (#240) —
in four independent copies, two of which the reader would have no structural
reason to look at.

So `tools/live_board.rs` owns the reading, and the three gates map its verdict
onto their own policy. The policies stay different on purpose: on an
unreachable IPC with a sibling lock present, a write **refuses** and a read
**discloses and proceeds**, because a stale read is disclosable and a stale
write destroys work. What is now single is the evidence, including the lock
probe — whose three answers (absent / present / uninspectable) the read path
had flattened to two, reading an inspection failure as an absence that the
write path is documented never to infer.

No response text changed on the write paths; their existing tests pin that.
The negative-control table below shows the payoff: neutering the #240 veto once
now fails six tests across reads *and* writes.

`read_board` also hands its closure the `DocumentSpecifier` KiCad resolved,
rather than leaving each handler to call `find_open_board` again. That removes
a redundant `GetOpenDocuments` per call, and makes it impossible for a #574
consumer to reach for a document-less client method — `get_nets()` rather than
`get_nets_in(document)` — and silently answer about whichever board KiCad
opened first.

## Deviations from the issue

- **A live layer's `id` is file-backed, not live.** KiCad's IPC API identifies a
  layer by its `BoardLayer` enum value (`BL_F_Cu` = 3); the `id` this tool has
  always reported is the file's stackup ordinal (`F.Cu` = 0). Those are
  different numbering schemes, and `add_layer` allocates in the file's. Putting
  the enum value under the existing key would silently repurpose it, so `id` is
  joined from the saved table by canonical name and labelled `saved_board`,
  `null` when the file has no such layer.
- **A sibling KiCad lock does not refuse a read.** When IPC is unreachable and
  `~<name>.kicad_pcb.lck` exists, `auto` still answers from the file but
  reports `kicad_ipc_unreachable_with_editor_lock` instead of
  `kicad_ipc_unreachable`. Reads are non-destructive and the limitation is
  disclosed; the write-side veto in `pcb_board.rs` is untouched.
- **`BoardAccess` classification.** Both tools now declare
  `LivePreferredWithFallback`, which is what they do. That adds them to the
  `pre-pcb-fallback` hook skill's tool list, whose text is worded for edits;
  the executable read-source inventory is #574 stage 1, not this PR.

## Fixture and oracles

`crates/konnect-core/tests/fixtures/board_source_divergence_kicad10.kicad_pcb`
— built from `specctra_two_resistors.kicad_pcb`, then re-serialized by
**KiCad 10.0.6** with `kicad-cli pcb upgrade --force`, so the committed bytes
are pcbnew's own output. Full provenance and case table in its `README.md`.

Its saved state shares no value with the KiCad double the tests run against, in
either direction, so a test cannot pass while reading the wrong adapter:

| Fact | Saved file | Live double |
|---|---|---|
| copper layers | 4 | 2 |
| `In1.Cu` / `In2.Cu` | present | absent |
| `F.Cu` displayed as | `F.Cu` | `LiveRenamedTop` |
| renamed fab layer | `SavedOnlyFabName` | absent |
| nets | `SAVED_ONLY_A`, `SAVED_ONLY_B` | `LIVE_ONLY_P`, `LIVE_ONLY_Q` |

External oracles for the saved half, all produced by `kicad-cli` from the
committed file and none of them reading Konnect code:

| Oracle | Confirms |
|---|---|
| `kicad-cli pcb export gerbers` | 18 enabled layers, one plot each, named by the displayed name — including `…-SavedOnlyFabName.gbr` |
| the same run's `…-job.gbrjob` | `"GeneralSpecs": { "LayerNumber": 4 }` and four `Copper,L1…L4` files |
| `kicad-cli pcb export ipc2581` | the same enabled set, with `F.Cu`/`In1.Cu`/`In2.Cu`/`B.Cu` as `CONDUCTOR` in stackup order |
| `kicad-cli pcb export ipcd356` | the nets `SAVED_ONLY_A` and `SAVED_ONLY_B` |

Neither export encodes a copper layer's `signal`/`power`/`mixed` kind, which is
why that attribute is reported as file-backed even in a live answer. Every
oracle value is restated as a literal `const` in the test module, with a
comment saying it is the promise — none is imported from the implementation or
re-derived from the fixture at test time.

## Tests

14 new contract tests in
`crates/konnect-core/src/tools/board_source_contract_tests.rs`, all driven
through served `tools/call` (`handle_message` with a JSON-RPC `tools/call`
request), so the advertised schema, the dispatch gate and the response body are
covered — not just the handler bodies. Plus 5 unit tests on the selector in
`board_source.rs` and 4 in `live_board.rs`, plus 2 on the write gates' lock
precedence in `pcb_board.rs`.

| Case | Test |
|---|---|
| stale-live-over-file precedence | `auto_answers_the_live_stackup_over_a_divergent_saved_one` |
| mixed-source provenance (layers) | `a_live_answer_reports_its_file_backed_fields_as_file_backed` |
| explicit saved inspection | `saved_inspects_the_file_even_while_kicad_holds_the_board` |
| wrong open board (live + auto) | `a_kicad_holding_another_board_refuses_live_and_explains_auto` |
| live-required refusal | `live_refuses_when_no_kicad_is_reachable` |
| post-identification IPC failure | `a_failure_after_identification_is_not_answered_from_the_file` |
| previously-live-then-lost IPC | `a_board_observed_live_then_lost_refuses_auto_and_discloses_under_saved` |
| mixed-source provenance (netclasses) | `netclasses_report_live_nets_beside_project_file_definitions` |
| saved-mode net derivation | `netclasses_saved_mode_matches_the_saved_nets_and_discloses_it` |
| post-identification failure (netclasses) | `netclasses_refuse_after_a_failed_live_query` |
| selector validation | `an_unknown_board_source_is_refused_by_name` |
| default and no request echo | `omitting_the_selector_reads_the_saved_board_and_says_so` |
| editor lost mid-call | `live_board::…::a_transport_that_dies_after_identification_is_a_loss_not_a_cold_start` |
| one declined layer name | `a_declined_layer_name_does_not_fail_the_live_read` |
| empty enabled-layer reply | `an_empty_enabled_layer_reply_is_not_reported_as_a_live_stackup` |
| lock outranks the observation | `board_session_safety_tests::an_exact_board_lock_still_outranks_the_previously_live_verdict` |

## Negative controls

Each guard neutered in turn, `cargo test -p konnect-core --lib` run, then
restored. Baseline 1428 passing.

| # | Neutered | Failures | Which |
|---|---|---|---|
| A | the #240 veto in the shared classifier | **6** | `a_board_observed_live_then_lost_…` + `board_session_safety_tests::{a_board_closed_since_konnect_saw_it_live_still_refuses_the_file, a_file_only_guard_blocks_a_previously_live_board_after_transport_loss, an_operation_rejected_after_identification_still_protects_the_next_call, live_then_dead_blocks_a_file_fallback_and_preserves_the_board}` + `pcb_components::tests::a_previously_live_board_blocks_place_components_file_fallback` |
| B | an uninspectable lock reads as absent | 2 | `live_board::tests::an_unreadable_lock_is_not_an_absent_one`, `board_session_safety_tests::an_uninspectable_exact_board_lock_fails_closed_with_distinct_evidence` |
| C | `Rejected` returns `Saved` instead of `Refused` | 2 | `a_failure_after_identification_is_not_answered_from_the_file`, `netclasses_refuse_after_a_failed_live_query` |
| D | `live` stops requiring a live board | 2 | `a_kicad_holding_another_board_refuses_live_and_explains_auto`, `live_refuses_when_no_kicad_is_reachable` |
| E | explicit `saved` early return removed | 3 | `saved_inspects_…`, `netclasses_saved_mode_…`, `a_board_observed_live_then_lost_…` |
| F | unknown selector accepted as `auto`, schema `enum` dropped | 3 | `an_unknown_board_source_is_refused_by_name`, `an_unknown_or_mistyped_mode_is_an_argument_error`, `the_advertised_schema_lists_the_modes_the_reader_accepts` |
| G | `get_layer_list` hardcoded to `BoardSource::Saved` | 6 | `a_board_observed_live_then_lost_…`, `a_failure_after_identification_…`, `a_kicad_holding_another_board_…`, `auto_answers_the_live_stackup_…`, `live_refuses_when_no_kicad_…`, `omitting_the_selector_…` |
| H | `get_netclasses` hardcoded to `BoardSource::Saved` | 2 | `netclasses_refuse_after_a_failed_live_query`, `netclasses_report_live_nets_beside_project_file_definitions` |
| I | live answer's file-backed fields labelled `ipc` | 1 | `a_live_answer_reports_its_file_backed_fields_as_file_backed` |
| J | session memory sampled before the IPC call, not after | 1 | `a_transport_that_dies_after_identification_is_a_loss_not_a_cold_start` |
| K | lock no longer outranks the observed-live verdict | 1 | `an_exact_board_lock_still_outranks_the_previously_live_verdict` |
| L | any layer-name failure aborts the whole read | 1 | `a_declined_layer_name_does_not_fail_the_live_read` |
| M | an empty enabled-layer reply is an answer again | 1 | `an_empty_enabled_layer_reply_is_not_reported_as_a_live_stackup` |

Control A is the one worth reading twice: it is the whole argument for sharing
the classification. Before this PR the same edit had to be made in four places
to be complete, and any one of them could be missed silently.

Every control was a deterministic source edit, not an iteration-order change,
and the tree was restored to a byte-identical state (`cargo fmt --check` clean,
`grep` for the control markers empty) before committing.

## Validation

Run on `bf9b8b8`, Fedora Linux, Rust 1.96.0, KiCad 10.0.6:

```text
cargo test --workspace --locked --lib --tests   # ok, 0 failed (1428 in konnect-core lib)
cargo test --workspace --locked --doc           # ok, 0 failed
cargo clippy --workspace --locked --all-targets -- -D warnings  # clean
cargo fmt --all -- --check                      # clean
cd crates/schematic-viewer && cargo test        # ok, 20 passed
```

Not run: the real-KiCad e2e suite (`e2e-kicad.yml`, `#[ignore]`d live tests) —
no KiCad IPC session was available here, so the live half is covered by the
deterministic double rather than by an unsaved-board observation, as the issue
allows.

## Risk and rollback

No mutation path's behaviour changed. `attempt_ipc_write` and
`refuse_if_board_open_in_kicad` were rewritten onto the shared classification
but reach the same verdict from the same evidence and emit the same text —
their existing tests pin that, and controls A and B above show both gates
still fail closed. No file mutation became easier to perform behind KiCad.

The behaviour change a caller can see is that both tools may now refuse where
they previously returned a saved-file answer — when KiCad holds the board and
rejects the query, or when an observed-live editor has vanished. Both refusals
name `board_source: "saved"` as the way to inspect the snapshot deliberately.
Rollback is the single commit; passing `board_source: "saved"` restores the
previous answer exactly.

## docs/API_MIGRATIONS.md

One new entry, inserted above the existing ones. The diff against `main` is
43 added lines and 0 deleted: nothing already on `main` is moved or removed.

`docs/TROUBLESHOOTING.md` and the `kicad-pcb` skill asset each gain a short
paragraph, because `unsafe_file_fallback` can now reach a caller who was not
writing anything and both documents previously said it always meant a refused
mutation.

## Review round

A self-review of the first draft found four defects in it, all fixed above and
each now pinned by a test and a negative control (J–M):

- **The session-memory sample moved to the wrong side of the IPC call.** The
  pre-refactor code read it *after*; the draft read it before. KiCad can
  identify the board and then go away before the next command — and since each
  command dials its own socket and `get_layer_list` issues one per layer, that
  window is ~30 dials wide. The draft classified that as a cold start and
  answered from the file for a board KiCad had just proven it was holding,
  which is the exact thing this PR exists to prevent, and a regression of #240
  on the write gates too.
- **A sibling lock stopped outranking the observed-live verdict.** Folding two
  arms into `LostAfterObservation` dropped the lock check, changing a stable
  machine-readable `reason` from `kicad_lock_present` to
  `board_previously_observed_live`. The variant now carries whether the
  transport is what went away, which is the only case where a lock is the
  better evidence.
- **One declined layer name failed the whole read.** With the board identified,
  a saved-file fallback is forbidden — so a single cosmetic lookup failure
  turned a tool that always answered into one that refused. KiCad-level
  refusals are now best-effort per layer (`display_name` stays unset, the
  canonical name stands in); a transport loss still propagates, because a
  partial read must not be dressed up as a complete live answer.
- **An empty enabled-layer reply was reported as a live stackup of nothing.**

## Known remaining, for #574

`get_board_info`, `get_board_extents` and `get_component_pads` are IPC-first
readers in these same two files that still swallow every `IpcFailure` and
answer from the file. They are not converted here — this PR is the first
focused consumer, not the migration — but they are the nearest holdouts, and
until they move a session can get a refusal from `get_layer_list` and a
cheerful `"source": "file"` from `get_board_info` about the same board. Worth
an early slot in stage 4.


mixelpixx (changes requested):
Reviewed exact head `bf9b8b8db87cad89a4324abb649f30378848520f`: one commit on current `main` (`c5b674e`), all ten required checks green on it, no review threads. This is careful work and the seam is the right shape. One finding blocks, and it only showed up against a real KiCad, which the body says was not available to you, so I ran that half here.

## What I verified

**Local gate on the head (Windows 11):** `cargo fmt --all -- --check`, `cargo clippy --workspace --locked --all-targets -- -D warnings`, `cargo test --workspace --locked --lib --tests` 2026 passed / 0 failed, `--doc`, `cargo xtask fix-doc-counts --check` unchanged, `reliability_contract` green. Exit codes captured directly.

**The write gates, arm by arm against `main`.** `attempt_ipc_write` and `refuse_if_board_open_in_kicad` reach the same verdict from the same evidence and emit the same text: rejected → same refusal / same `None`; proves-not-open → file edit unless observed live; other target errors → the typed refusal; unreachable → lock first, then the observed-live veto, then the file. Session memory is still sampled after the call. `with_board_ipc_classified` now delegates to the bound variant, and `ensure_board_is_active` was already `find_open_board(..).map(|_| ())`, so targeting is unchanged. I agree with "no mutation path's behaviour changed".

**Tests and fixture.** The contract tests drive served `tools/call` through `McpHandler::handle_message`, as #574 asks; the fixture is `kicad-cli pcb upgrade` output with oracles that read no Konnect code.

**One negative control of my own**, independent of your table: under `auto`, make a rejection after identification answer from the saved file. Three served tests fail (`a_failure_after_identification_is_not_answered_from_the_file`, `netclasses_refuse_after_a_failed_live_query`, `an_empty_enabled_layer_reply_is_not_reported_as_a_live_stackup`); it compiled, and the tree was restored byte-identical.

## Live evidence (KiCad 10.0.5, Windows 11, release build of this head over stdio)

A standalone pcbnew holding a copy of the `ecc83` demo, with the saved file changed *behind* the editor afterwards (`F.Cu`'s user name `top_cu` → `SAVED_ONLY_top` on disk only), so the two sources cannot agree by accident:

| State | Call | Result |
|---|---|---|
| pcbnew holds the board | default / `auto` / `live` | `F.Cu` → `top_cu` (the editor's), `sources`: enabled_layers / layer_names / copper_layer_count `ipc`, layer_ids / layer_types `saved_board`; `board_state: ipc` |
| same | `saved` | `F.Cu` → `SAVED_ONLY_top`; `reason: explicitly_requested_after_live_observation` |
| same | `get_netclasses` `auto` / `saved` | `board_nets: ipc` / `saved_board`; definitions and patterns `project_file` both times |
| same | `live`, a different board | `wrong_document`, naming the board pcbnew does hold |
| same | `auto`, a different board | saved file, `reason: board_not_open_in_kicad` |
| pcbnew killed mid-session (lock file left behind) | `auto` and `live` | `unsafe_file_fallback`, `reason: board_previously_observed_live` |
| same | `saved` | saved file, `explicitly_requested_after_live_observation` |

Every claim in the body that I could exercise holds on the real editor.

## Blocking: with only the KiCad project manager running, the default call now refuses

Start KiCad and do not open the PCB editor. That is the state every user is in between launching KiCad and opening a board. On `main` both tools answer from the file there. On this head:

```
get_layer_list {board}                       (default, and board_source: "auto")
→ isError, kind: editor_unavailable
  "KiCad did not complete the layer list for this board: KiCad IPC error: no handler
   available for request of type kiapi.common.commands.GetOpenDocuments (AS_UNHANDLED).
   Konnect did not substitute the saved board file, which may be older than the
   editor's state. …"
get_netclasses {board}                       → the same refusal
get_layer_list {board, board_source: "saved"} → answers, reason: explicitly_requested
```

There is no editor whose state the file could be older than. `find_open_board` fails inside `GetOpenDocuments` itself, `IpcFailure::from_error` classifies any status error as `Rejected`, and `read_board` refuses every `Rejected`. But #574's rule is scoped: "*Once the exact board has been positively identified as live*, a failed live query must not silently become a successful saved-file answer." Here nothing was identified; KiCad said it has no handler for board documents at this endpoint. `LiveBoard::Rejected`'s own doc ("It may hold the board") is true after identification and not before.

Requested change, kept inside this PR's scope:

- Distinguish a rejection **before** identification from one **after**. The status is already modelled: `ApiStatusError::is_unsupported()` (`AS_UNHANDLED` / `AS_UNIMPLEMENTED`), used the same way in `editor_navigation.rs`. When `GetOpenDocuments` is refused with it, `observe` reports its own verdict (say `NoBoardEditor`).
- `read_board`: under `auto`, treat it like `NeverReached` — probe the sibling lock, read the saved file, and disclose with its own machine-readable reason (for example `no_pcb_editor_at_endpoint`, plus the `…_with_editor_lock` / `…_with_uninspectable_lock` forms you already have). Under `live`, refuse as now. If this session observed the board live earlier, it stays `LostAfterObservation`. A standalone pcbnew on its own endpoint is invisible from here, which is equally true of `NeverReached`; the lock probe is the evidence for both.
- A rejection after identification keeps refusing, whatever its status. Control C stays.
- **The write gates map the new verdict exactly as they map `Rejected` today**, so no mutation behaviour changes here. Today that status makes `attempt_ipc_write` refuse while `refuse_if_board_open_in_kicad` proceeds; that inconsistency predates you and belongs to #577. A one-line comment at the `Rejected | NotOpen => None` arm saying it is preserved, not decided, would stop the next reader taking it for policy.
- Served regressions with a double that answers `GetOpenDocuments` with `AS_UNHANDLED`: `auto` answers from the file with the new reason; `live` refuses; observed-live-then-unhandled refuses; the existing write-gate tests unchanged. Plus a negative control for the new arm.
- Migration entry and TROUBLESHOOTING: one sentence for this state.

## Non-blocking

- The migration heading has no release class. Neighbouring entries carry one; this is a **minor** (additive argument, new response fields, a former answer that can now refuse).
- On Windows the `wrong_document` message prints the open board with a `\\?\` prefix. Cosmetic, probably older than this PR; worth a follow-up issue, not a change here.
- "Known remaining" is right, and I would put it first in #574 stage 4: until `get_board_info`, `get_board_extents` and `get_component_pads` move, one session can get a refusal from `get_layer_list` and a cheerful `"source": "file"` from `get_board_info` about the same board.

Closure accounting is correct as written: `Closes #542`, `Part of #574`. Reply with the new head when the change is in and I will re-run the gate and the three live states on it.

pauliuszaleckas:
Thank you — the blocking finding is correct, and it is a state I had no way to
reach with a mock: `GetOpenDocuments` refused with `AS_UNHANDLED` never entered
my test matrix because my double only ever refused *after* answering it. Fixed
on `274cab9`.

## What changed

**Where the classification happens.** `IpcFailure` keeps only the rendered
message, so by the time any caller sees a `Rejected` the typed status is gone —
which is why the two cases were indistinguishable. The binding step now
classifies while the chain is still intact, using the same
`ApiStatusError::is_unsupported()` the editor-state tools use, never message
text:

```rust
// tools/mod.rs, inside with_bound_board_ipc_classified
let document = match client.find_open_board(&requested) {
    Ok(document) => document,
    Err(error)
        if konnect_ipc::ApiStatusError::from_error(&error)
            .is_some_and(|status| status.is_unsupported()) =>
    {
        return Ok(BoardBinding::NoBoardEditor(format!("{error:#}")));
    }
    Err(error) => return Err(error),
};
memory.observe_live(&observation);
```

That scopes it exactly as you asked: only the identification step can produce
it, so a command refused with the same status *after* the board is named stays
`Rejected` and still refuses. Control C is untouched and still fails 2 tests.

**`LiveBoard::NoBoardEditor { message, observed_live }`.** I carried
`observed_live` on the variant rather than folding it into
`LostAfterObservation` inside `observe`, because the gates genuinely disagree
about it and I did not want that disagreement to be invisible. The read path
honours it; the write gates deliberately do not read it.

**Write gates.** Both map the new verdict exactly as they mapped this status
before it had a name — `attempt_ipc_write` refuses, `refuse_if_board_open_in_kicad`
returns `None` — with the comment you asked for at each site saying it is
preserved, not decided, and naming #577. `board_session_safety_tests` (19) and
`open_document_ambiguity_tests` (11) are unchanged and green.

**Read path.** Under `auto`: probe the sibling lock, read the saved file,
disclose `no_pcb_editor_at_endpoint` (plus the `…_with_editor_lock` /
`…_with_uninspectable_lock` forms). Under `live`: refuse. Observed live earlier:
`unsafe_file_fallback`. The unreachable and no-editor cases now share one
`no_live_board` helper — a KiCad this process cannot see is invisible either
way, and the lock is the only evidence for both — with the reason code naming
which it was. The codes resolve through a `match` to `&'static str` so the
vocabulary stays a closed set rather than a string that function happens to
build.

## New tests (served `tools/call`)

| Case | Test |
|---|---|
| no PCB editor, default and `auto`, both tools | `a_kicad_with_no_board_editor_still_answers_by_default` |
| no PCB editor, `live` | `a_kicad_with_no_board_editor_refuses_a_live_request` |
| observed live, then the editor closes under the session | `a_board_seen_live_before_the_editor_closed_still_refuses` |

The third one needed `MockIpcServer::spawn_at`, so the project manager can take
over the *same* endpoint after the PCB editor's guard is dropped. A replacement
handler would have had a fresh `BoardSessionMemory` and the test would have
passed for the wrong reason.

## New negative controls

| # | Neutered | Failures | Which |
|---|---|---|---|
| N | `NoBoardEditor` refuses under `auto` again (the reported regression) | 1 | `a_kicad_with_no_board_editor_still_answers_by_default` |
| O | the observed-live exception is dropped | 1 | `a_board_seen_live_before_the_editor_closed_still_refuses` |

Both compiled, both restored byte-identical.

## Non-blocking, addressed

- Migration heading now carries **(minor release)**, and the entry covers this
  state — including that mutating tools are unchanged by it.
- `TROUBLESHOOTING.md` now says up front that a KiCad with no PCB editor is
  *not* the `unsafe_file_fallback` state, so the refusal section is not read as
  covering it.

## Non-blocking, not done here

- The Windows `\\?\` prefix on the `wrong_document` message — agreed it predates
  this PR. Say the word and I will open an issue with your repro rather than
  widen this branch.
- "Known remaining" first in #574 stage 4 — agreed; I will leave that note on
  #574 rather than here.

## Validation on `274cab9`

Fedora Linux, Rust 1.96.0:

```
cargo test --workspace --locked --lib --tests   # ok, 0 failed (1431 in konnect-core lib)
cargo test --workspace --locked --doc           # ok, 0 failed
cargo clippy --workspace --locked --all-targets -- -D warnings  # clean
cargo fmt --all -- --check                      # clean
cd crates/schematic-viewer && cargo test        # ok, 20 passed
```

Still no live KiCad here, so the three states you ran remain your measurement,
not mine — the new ones are covered by a double that answers `AS_UNHANDLED`.

pauliuszaleckas:
Correction to the head SHA above: the branch was rebased onto `main` @ `cdcafa8`
shortly after I posted, so the head to gate is
**`bc1bdca9b7cda44af9492869b3df5611ffdfd724`**, not `274cab9`.

The rebase carries no code change. `git range-diff 274cab9^..274cab9 cdcafa8..bc1bdca`
shows one difference: #663's `docs/API_MIGRATIONS.md` entry landed on `main` in between,
so mine now follows it instead of opening the file — every existing entry untouched and
in its original order. The commit message is unchanged. All ten required checks are green
on `bc1bdca`.

Everything else in my reply above stands as written against that head.

neusse:
The requested correction is now present on exact head `bc1bdca9b7cda44af9492869b3df5611ffdfd724`, rebuilt on current `main`, with all ten required checks green and no unresolved review threads. The author is no longer the next actor.

@mixelpixx, please re-review this exact head and replace or dismiss the earlier `CHANGES_REQUESTED` review if the `NoBoardEditor` split and served regressions satisfy your live finding. GitHub still reports the PR as blocked solely by that earlier review state.

pauliuszaleckas:
New head: **`13fe0f49cbf8fcc1d9f8ea0392f4191236e0c66e`**, same single commit on the same base (`cdcafa8`), rebuilt to carry two corrections I found reviewing my own diff after the last push. Neither changes where any of the five states lead; both change what the code claims, and one closes a stall I would rather not ship.

Sorry for moving the head while it was queued for re-review — I would rather it be re-reviewed once, on a head that does not need this follow-up.

## 1. `AS_UNHANDLED` and `AS_UNIMPLEMENTED` are not the same answer

The previous head classified the binding failure with `ApiStatusError::is_unsupported()`, which is true for both, and reported either as `no_pcb_editor_at_endpoint`: *"KiCad is running but reports no PCB editor at this endpoint"*.

Only `AS_UNHANDLED` says that. `AS_UNIMPLEMENTED` is a build that does not implement `GetOpenDocuments` at all — it may well have a board open with unsaved changes, and nothing about its editors was established. That is exactly the kind of unearned positive claim the provenance vocabulary exists to avoid, and the editor-state tools already keep the two apart (`editor_navigation.rs` answers `unsupported_capability`, and `IpcEditorObservation` says *"GetOpenDocuments is {code_name} on this endpoint"*).

So the binding step now classifies into `BoardEditorAbsence::{NoPcbEditor, OpenDocumentsUnimplemented}`, and the second gets its own reason code:

```
open_documents_unimplemented_at_endpoint
open_documents_unimplemented_at_endpoint_with_editor_lock
open_documents_unimplemented_at_endpoint_with_uninspectable_lock
```

with the detail *"KiCad is running but does not implement the command that names its open documents, so whether a PCB editor holds this board was never established"*.

Unchanged: both still mean no live board, both still answer from the saved file under `auto`, both still refuse under `live`, both still refuse when this server saw the board live earlier, and mutating tools still map both exactly as they mapped a rejection before either had a name. `excludes_unsaved_editor_state` stays `true` for both — that is a property of reading the file, not a claim about KiCad.

The `LiveBoard`/`BoardBinding` variant is renamed `NoBoardEditor` → `Unserved { absence, … }`, because a variant named for one of the two answers would misdescribe the other.

## 2. The per-layer name lookups were unbounded

`get_enabled_layers_with_names_in` asked `GetBoardLayerName` once per enabled layer. Each is its own socket with a 30 s receive timeout, and a receive timeout is not `TransportUnreachable`, so it took the "declined one name" path and the loop continued. On a full stackup against a KiCad that answers the enabled set and then wedges — a modal dialog, a long zone refill — a default `get_layer_list` would have blocked for up to 32 × 30 s where it used to be an instant file parse.

The names are cosmetic and the canonical name is always available, so they now have a budget (`LAYER_NAME_BUDGET`, 10 s): the loop stops when it is spent and warns with how many layers went unnamed. `get_enabled_layers_named_within` takes the budget explicitly, which is also what makes it testable without a wedged KiCad.

## 3. Markdown nit

The `API_MIGRATIONS.md` heading this PR adds now has the blank line before it that every other section has.

## Evidence on `13fe0f49cbf8fcc1d9f8ea0392f4191236e0c66e`

| Command | Result |
|---|---|
| `cargo test --workspace --locked --lib --tests` | pass — 32 suites, 0 failed |
| `cargo test --workspace --locked --doc` | pass |
| `cargo clippy --workspace --locked --all-targets -- -D warnings` | clean |

New tests, and the controls for them:

| Test | Control | Result |
|---|---|---|
| `an_unimplemented_open_document_command_claims_nothing_about_editors` — a mock KiCad answering `AS_UNIMPLEMENTED`; asserts the new reason, that the detail says what KiCad answered, that it does **not** say "no PCB editor", and that `live` still refuses | classify `AS_UNIMPLEMENTED` back to `NoPcbEditor` | **1 failed** |
| `a_spent_budget_stops_the_lookups_and_keeps_the_enabled_set` — counts the `GetBoardLayerName` round trips, so the stop is proven rather than inferred from missing names | delete the deadline check | **1 failed** |
| `a_budget_with_room_resolves_every_name` — the same mock with room to spare, so the budget cannot pass by never resolving anything | — | — |

`a_kicad_with_no_board_editor_still_answers_by_default`, the maintainer's live finding, is unchanged and still passes on this head.

One thing I am deliberately leaving out of this PR: `handle_get_board_info` still calls `with_board_ipc_classified` and re-derives the document inside its closure, paying a second `GetOpenDocuments` — the pattern `with_bound_board_ipc_classified` exists to retire. That call site predates this branch (it came in with `e3366b9`), so migrating it here would be unrelated cleanup in the diff. Happy to send it as its own PR.

pauliuszaleckas:
GitHub reported this PR as conflicting after the last push — #665 landed on `main` and took the spot my `docs/API_MIGRATIONS.md` entry occupied, which also meant no merge ref and therefore no CI runs at all on the previous head.

Rebuilt as one commit on current `main` (`87e3e8f`): head is now **`00cf058e16883648c4ac18831b409efc50211bcc`**. One force-push, no further rebases from me while this is in review.

The rebase carries no code change. `git range-diff 13fe0f4~1..13fe0f4 87e3e8f..00cf058` shows a single difference — my `API_MIGRATIONS.md` entry now follows #665's `add_power_symbol` entry instead of #663's config entry. Every entry already on `main` is untouched and in its original order; the file's diff against `main` deletes nothing and moves nothing.

Gates re-run on this exact head: `cargo test --workspace --locked --lib --tests` (32 suites, 0 failed), `cargo test --workspace --locked --doc`, `cargo clippy --workspace --locked --all-targets -- -D warnings`, `cargo fmt --all -- --check` — all clean. Everything in my previous comment still describes this head.

neusse (changes requested):
Please make the layer-name timeout behavior match the advertised 10-second total bound before this can land.

`get_enabled_layers_named_within` checks its deadline only before starting each synchronous `get_board_layer_name_in` call. Each call still uses the client's fixed 30-second receive timeout, so a call begun just before the deadline can block for roughly 30 seconds. The current immediate-response/zero-budget tests do not prove the stated wall-clock bound.

Please either:

- enforce the remaining budget as the receive timeout for each layer-name request, with a delayed mock proving the operation stays within the declared bound; or
- narrow the public/code contract to say that Konnect stops *starting* new lookups after ten seconds, explicitly acknowledging the possible one-call overrun and testing that exact behavior.

The rest of the corrected exact head matches #542/#574, including source selection, provenance, fail-closed live reads, write-gate preservation, and the AS_UNHANDLED/AS_UNIMPLEMENTED distinction. After this timeout contract is corrected, push a new head and request exact-head review; the existing stale owner review will also need to be superseded or dismissed.

pauliuszaleckas:
The finding is correct, and the first of your two options is the right one —
the bound should be what it says it is. New head:
**`caa7095fc335de5bcf8de947e968ea5b8649c71a`**, one commit on current `main`
(`116a207`), which also clears the conflict #668 created.

## What was wrong

`get_enabled_layers_named_within` checked the deadline only as a gate on
*starting* another lookup, and every lookup then went out through
`send_command`, whose receive timeout is a fixed 30 s. A lookup begun a
millisecond under the deadline could therefore wait 30 s on its own and the
loop would "stop at the budget" afterwards, having already overrun it by
three times over.

## The fix

The budget is now the receive timeout as well as the gate. `send_command`
gained a `send_command_within` beneath it that takes the caller's own
reply deadline; the naming loop passes what is left of the budget:

```rust
let remaining = deadline.saturating_duration_since(std::time::Instant::now());
if remaining.is_zero() {
    tracing::warn!(named = resolved, unnamed = …, "…reporting the canonical name for the rest");
    break;
}
match self.get_board_layer_name_within(document.clone(), layer.id, remaining) {
```

A receive timeout is a plain `anyhow` error with no `TransportUnreachable`
marker, so a lookup cut short by the budget lands in the same best-effort arm
as a KiCad that declines a name: `display_name` stays unset, the canonical
`name` stands, and the read still succeeds. A transport that dies mid-loop
still propagates, unchanged. `COMMAND_RECV_TIMEOUT` (30 s) and
`COMMAND_SEND_TIMEOUT` (5 s) are now named constants; every other command goes
through `send_command` and is untouched.

The send timeout is deliberately *not* clamped to the remaining budget. The
socket is already dialed by then, so it bounds only a send queue that will not
drain — and shrinking it would turn a nearly-spent budget into a
`TransportError`, i.e. a failed live read, which is exactly the outcome the
best-effort arm exists to avoid.

## Evidence — delayed mock, wall clock

`crates/konnect-ipc/tests/mock_server_test.rs`, module `layer_name_budget`.
The mock now answers `GetBoardLayerName` after a settable delay.

| Test | KiCad takes | Budget | Asserts |
|---|---|---|---|
| `a_lookup_started_under_the_deadline_still_stops_at_it` | 8 s | 400 ms | elapsed < 4 s **and** < KiCad's reply; exactly 1 lookup asked; enabled set intact; no `display_name` |
| `the_budget_is_spent_across_the_lookups_not_renewed_by_each` | 300 ms each | 500 ms | the second layer is unnamed — the first spent most of the budget |
| `a_spent_budget_stops_the_lookups_and_keeps_the_enabled_set` (existing) | — | 0 | 0 lookups asked |
| `a_budget_with_room_resolves_every_name` (existing) | — | 30 s | both named |

All four run in 0.50 s.

## Negative controls, on this head

| # | Neutered | Failures |
|---|---|---|
| M | the per-call receive timeout: `send_command_within` ignores it and uses `COMMAND_RECV_TIMEOUT` | **2** — `a_lookup_started_under_the_deadline_still_stops_at_it` (*"the naming loop waited 8.001173379s, past its 400ms budget"*), `the_budget_is_spent_across_the_lookups_not_renewed_by_each` |
| N | the budget is renewed per lookup: pass `budget` instead of `remaining` | **1** — `the_budget_is_spent_across_the_lookups_not_renewed_by_each` |
| O | the deadline gate removed from the loop | **1** — `a_spent_budget_stops_the_lookups_and_keeps_the_enabled_set` |

Control M reproduces your finding as a measurement: 8.0 s against a 400 ms
bound, and it is the reply that ends the wait, not the deadline. The tree was
restored byte-identical after each.

## Also on this head

`docs/API_MIGRATIONS.md` now states the bound where callers will look: ten
seconds of KiCad's time for the `display_name` lookups, a bound on the waiting
rather than only on when the last lookup may start, with `display_name: null`
on the layers it did not reach. It went into the existing `get_layer_list` /
`get_netclasses` entry rather than opening a second one. The reconstruction put
that entry after #668's — every entry already on `main` is untouched and in its
original order, and the file's diff against `main` deletes nothing and moves
nothing.

Nothing else changed: the range-diff against the previous head is this timeout
work and that paragraph.

## Validation — on `caa7095fc335de5bcf8de947e968ea5b8649c71a`

| Command | Result |
|---|---|
| `cargo test --workspace --locked --lib --tests` | pass — 32 suites, 2073 passed, 0 failed |
| `cargo test --workspace --locked --doc` | pass |
| `cargo clippy --workspace --locked --all-targets -- -D warnings` | clean |
| `cargo xtask fix-doc-counts --check` | unchanged |

Linux only here; the 3-OS matrix is CI's. No live KiCad was available for this
head — the new evidence is wall-clock against a mock, which is what the bound
is about.

**All ten required checks are green on this exact head**, and
`closingIssuesReferences` still resolves to #542.

Ready for exact-head review. The earlier `CHANGES_REQUESTED` from @mixelpixx
still needs superseding or dismissing.

neusse (approved):
Exact-head review complete for `caa7095fc335de5bcf8de947e968ea5b8649c71a` on current `main` `116a207620264de9bc96bf239182cfd0358fcff9`.

The new per-call receive deadline and delayed wall-clock controls resolve the 10-second budget finding. The rest of the implementation still satisfies #542 and its scoped part of #574: exact live-board selection, explicit auto/live/saved behavior, post-identification fail-closed handling, mixed-source provenance, real-KiCad-derived fixture coverage, served-dispatch tests, and unchanged write-gate policy. Both spec and standards reviews found no remaining blocker, all ten required checks pass on this exact head, and `Closes #542` / `Part of #574` is correct.

The only remaining block is GitHub's stale `CHANGES_REQUESTED` review from the older head.

neusse:
@mixelpixx, the blocking editor-state finding from your review and the later timeout-budget finding are both corrected on exact head `caa7095fc335de5bcf8de947e968ea5b8649c71a`; current exact-head spec/standards review and all ten required checks are green. Please re-review or dismiss the stale `CHANGES_REQUESTED` review on `bf9b8b8` so GitHub can expose the current head's actual merge state.

mixelpixx (approved):
Exact-head re-review of **`caa7095`** (one commit on `main` `116a207`). This supersedes my `CHANGES_REQUESTED` on `bf9b8b8`: the blocking finding is fixed, and I measured it on a real KiCad rather than taking the double's word.

## What I checked against my request

Delta read from the last head I reviewed. `main` touched none of these files in between, so the diff is the author's own:

- **Classification at the binding step only.** `BoardEditorAbsence::from_error` reads the typed status off `find_open_board`'s error: `AS_UNHANDLED` → `NoPcbEditor`, `AS_UNIMPLEMENTED` → `OpenDocumentsUnimplemented`, anything else stays a rejection. A refusal after the board is named never reaches it, so control C still holds.
- **Read path.** `observed_live` beats both absences (`unsafe_file_fallback`); `live` refuses; `auto` reads the saved file with a reason from a closed `&'static str` set (`no_pcb_editor_at_endpoint`, `open_documents_unimplemented_at_endpoint`, each with the `…_with_editor_lock` / `…_with_uninspectable_lock` forms). Splitting `AS_UNIMPLEMENTED` out is right: it establishes nothing about editors, and the detail no longer claims it does.
- **Write gates unchanged.** Both map `Unserved` exactly as they mapped this status before it had a name, each with the "preserved, not decided, #577" comment. `board_session_safety_tests` and `open_document_ambiguity_tests` are byte-identical to the head I first reviewed.
- **Layer-name budget.** The remaining budget is the receive timeout of each lookup, not only the gate on starting one, and a lookup cut short falls into the best-effort arm (canonical `name` stands, `display_name` unset). That resolves George's finding.

## Evidence on `caa7095` (Windows 11, KiCad 10.0.5)

**Local gate**, exit codes captured directly: fmt, clippy `-D warnings`, `--lib --tests` **2061 passed / 0 failed** (Windows; the Linux count includes Unix-only tests), `--doc`, `fix-doc-counts --check` unchanged, `reliability_contract` green, release build ok.

**Live, release build of this head over stdio**, on a fresh copy of KiCad's `ecc83` demo. In states C and D the saved file's `F.Cu` user name is changed *behind* the editor (`top_cu` in pcbnew, `SAVED_ONLY_top` on disk), so the two sources cannot agree by accident:

| State | Call | Result |
|---|---|---|
| **A.** project manager only, no lock (the state of my finding) | default and `auto`, both tools | answer from the saved file, `reason: no_pcb_editor_at_endpoint`, `excludes_unsaved_editor_state: true` |
| same | `live` | `editor_unavailable`, naming `AS_UNHANDLED` |
| same | `saved` | saved file, `reason: explicitly_requested` |
| **B.** project manager only, stale sibling lock | default and `auto` | saved file, `reason: no_pcb_editor_at_endpoint_with_editor_lock` |
| **C.** pcbnew holds the board, file diverged | default / `auto` / `live` | `F.Cu` → `top_cu`, the editor's; `board_state: ipc` |
| same | `saved` | `F.Cu` → `SAVED_ONLY_top`; `reason: explicitly_requested_after_live_observation` |
| same | `live`, another board | `wrong_document`, naming the board pcbnew holds |
| same | `auto`, another board | saved file, `reason: board_not_open_in_kicad` |
| **D.** one server session: live, then pcbnew killed | `auto` / `live`, both tools | `unsafe_file_fallback`, `reason: board_previously_observed_live` |
| same session, then the project manager alone takes the endpoint | default / `auto` / `live`, both tools | still `unsafe_file_fallback`, detail "KiCad now reports no PCB editor at this endpoint": the observed-live veto beats the new verdict, as it should |
| same | `saved` | answers, `explicitly_requested_after_live_observation` |

The real project manager answers `AS_UNHANDLED` to `GetOpenDocuments`, which is what the double answers, so the served tests model the state I hit. I have no KiCad build that answers `AS_UNIMPLEMENTED`, so that arm rests on the double.

**Independent negative controls** (mine, not the author's M/N/O), each alone on the exact head and restored byte-identical:

| Neutered | Result |
|---|---|
| the observed-live exception dropped in `read_board` | **caught**: `a_board_seen_live_before_the_editor_closed_still_refuses` |
| the layer-name lookup waits the default 30 s instead of what is left of the budget | **caught**: `a_lookup_started_under_the_deadline_still_stops_at_it`, `the_budget_is_spent_across_the_lookups_not_renewed_by_each` |
| `BoardEditorAbsence::from_error` widened: *any* status refusing `GetOpenDocuments` counts as no PCB editor | **not caught**: the whole `konnect-core` suite passes (1472) |
| `attempt_ipc_write` sends an unserved endpoint to the file instead of refusing | **not caught** (1472 pass) |
| `refuse_if_board_open_in_kicad` refuses an unserved endpoint instead of returning `None` | **not caught** (1472 pass) |

None of the three uncaught ones is a defect on this head: the code does the right thing in each. They are validation debt, below.

## Non-blocking

- **The other side of the classifier is untested.** Every scene that refuses `GetOpenDocuments` itself answers `AS_UNHANDLED` or `AS_UNIMPLEMENTED`; none answers `AS_BUSY` or `AS_NOT_READY`. So "anything else stays a rejection and refuses under `auto`" rests on the `else { None }` alone, and widening it passes every test. That is the branch that matters most here: a pcbnew that is busy, or still starting with the board open (I have seen `AS_NOT_READY` from a freshly launched pcbnew on this machine), may hold unsaved state, and a widened classifier would answer it from the file. One served test with the double answering `AS_BUSY` pins it: `auto` refuses with `editor_unavailable`, no saved-file answer.
- **"Mutating tools treat this state exactly as before" has no test at the new variant.** Both write-gate mappings of `Unserved` can be flipped without a failure. That was already true of `AS_UNHANDLED` before this PR gave it a name, so it is not a regression, but a test pinning today's mapping would make #577's eventual decision show up as a visible test change rather than a silent one.

Neither needs a new head here. I will send the three tests as one small follow-up PR once this lands (and after #677's reconstruction, which touches the same binding helper), unless @pauliuszaleckas would rather carry them.

- **#673 has a second site.** The lock path in `source_evidence.detail` also prints with the Windows `\\?\` prefix (state B: `the sibling lock '\\?\C:\…\~ecc83-pp.kicad_pcb.lck' is present`). Same cause as `wrong_document`'s open-board list; I will add it to #673 rather than widen this PR.

Approving this exact head. Sequencing against #675, which also inserts at the top of `docs/API_MIGRATIONS.md`, is George's call; I have not merged.

mixelpixx:
The three follow-up tests from my review are written and verified on top of `caa7095`: branch `test/656-unserved-coverage`, commit `1c506ec`, test-only. Each of the three neuters that passed silently here is now caught by exactly one test; gate 2064 / 0. I will open the PR once this lands, so nothing here needs to change. @pauliuszaleckas, no need to carry them.

mixelpixx:
Follow-up tests are up as #681 (`1c506ec`, test-only, CI 10/10), now that this has landed.

